PR #65, opened 2022-03-22
Use frontmatter to get local image

hanbyul-here:
tldr:
I can think of two not-that-great options to let editors use local images. 

1. ask users to put local assets in `static` folder
2. predefine local assets in front matter

---

At first, I thought all the images and static assets can go to `public/static` folder 
(like how Chart component is handling local data 

```
<Chart
  dataPath='/public/example.csv' // <- data is in /static/public folder
  dateFormat="%m/%d/%Y" 
  idKey='County' 
  xKey='Test Date' 
  yKey='New Positives' 
  highlightStart = '12/10/2021'
  highlightEnd = '01/20/2022'
  highlightLabel = 'Omicron'
/>
```
)

but then I realized this makes users to put images or static assets to put in two separate places which is not ideal. but I am not sure how we can pass parcel-generated image path in async-loaded mdx contents 🤔 . (what @danielfdsilva  explained about parcel's image asset generation system helped me figure out the problem fast.) so this is what I came up with, defining local assets in frontmatter and using `embedded` props to indicate that it is a local asset to component. I feel like it is not an intuitive way of using local assets, but can't think of any other ways.. wonder any of you have a better idea? @ricardoduplos @danielfdsilva ?

This pr is just a proof of concept, but opening for discussion.

### DIFF
--- a/app/scripts/components/common/images/index.js
+++ b/app/scripts/components/common/images/index.js
@@ -9,6 +9,7 @@ import {
 } from '$components/common/figure';
 
 import { captionDisplayName } from '$components/common/blocks/block-constant';
+import { useThematicAreaDiscovery } from '$utils/thematics';
 
 export const Caption = function ({ children, attrAuthor, attrUrl }) {
   return (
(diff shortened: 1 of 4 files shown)


danielfdsilva:
Closing as current approach seems to be flexible enough

PR #159, opened 2022-06-09
Component overrides

danielfdsilva:
Experimentation with component overrides.

Check http://localhost:9000/sandbox/override


app/scripts/components/root-about/index.js
```diff
@@ -1,63 +1,34 @@
...
-          with datasets and their respective stories. Through visualizations,
-          trend analyses, and comparative analyses of datasets users can explore
-          the applications and implications of those data. VEDA Dashboard will
-          improve the usability of the Earth science data and enable faster
-          science – one of the goals of the NASA open source science initiative.
-          Please submit any feedback or questions using the Feedback button in
-          the top right of the dashboard.
-        </p>
-      </FoldProse>
+      {aboutContent ? (
```
hanbyul-here, line 22: Docs is recommending to use  `<ComponentOverride>` to use over-ridden component but it looks like about page is not using `ComponentOverride`? Any reason?

danielfdsilva: So `<ComponentOverride>` would be used to override specific components in the site (like a footer). It does not show a loading placeholder (as it would be weird and tricky to have loading indicator for all occasions).

The about page is being loaded with the `MdxContent` to behave exactly like any other mdx page.

Also, at the moment, `<ComponentOverride>` does not provide any blocks to the `MdxProvider` like `MdxContent`does. Not sure we need `Blocks` when overriding components.
Would like to hear your thoughts on this.

hanbyul-here: Ah too bad that we can't consolidate the syntax. (well but it will be mainly us who will touch this, so I think it should be just well documented.) 
Hm I am not very sure about whether to offer `Blocks` or not. but I think it will be needed for Veda dashboard considering how people are using `usage` tab now. so maybe a good idea to offer it?

danielfdsilva: I think that the usage page would be a different thing. It seems that `usage` would be a long form page that can be different for each dataset, so the implementation would be different (like outlined in https://github.com/NASA-IMPACT/delta-ui/issues/188#issuecomment-1168576285)

danielfdsilva:
@hanbyul-here Wrote some more in-depth documentation for the override feature.

I created [developer specific documentation](https://github.com/NASA-IMPACT/delta-ui/blob/feature/component-overrides/PAGE_OVERRIDES.md) in `delta-ui` which basically explains how to setup the overrides.

And then [user specific documentation](https://github.com/NASA-IMPACT/delta-config/blob/feature/component-overrides/docs/CONFIGURATION.md#component-overrides) in `delta-config` explaing how to use existing overrides.

Please have a look and let me know if it is clear. Any typos fix or re-wording you feel needed are welcome. :)

app/scripts/components/root-about/index.js
hanbyul-here: Ah right. Apology that I was confused with about page and usage page. Yeah I think we might need it, or not. We probably can add it when we need it in the future.

PAGE_OVERRIDES.md
```diff
@@ -0,0 +1,169 @@
+# Overrides
+
+💡 These notes cover the development part of the content overrides. For usage check [delta-config](https://github.com/NASA-IMPACT/delta-config/).
```
hanbyul-here, line 3: A link to specific part of the doc might be more helpful. https://github.com/NASA-IMPACT/delta-config/blob/feature/component-overrides/docs/CONFIGURATION.md#component-overrides

Also, this is a crucial part of setting up overrides content, I think we can use a stronger words that you need to read this doc first.

danielfdsilva: Agree with the link. Will update once merged so we have the correct url.

These docs are to be used only by the developers to create the overrides. I don't think it should be needed to read the other ones. This "developer docs" file should contains all that is needed to understand what they are and how to create them for the end user. These end users don't need to know how they're created, but only how to use them. What do you feel is missing from these docs?

danielfdsilva:
- I'd say that thematic areas, datasets, etc are not customizable (as in there is no default). You either create them or not, and are the different content types of the app. The about page has a default value (which is not really usable, but alas), and you can change it if you want. That being said, I understand that from the user point of view, this difference is very nuanced, so maybe there could be a better approach.
  - We could expect a particular file to exist somewhere for example, but that would deviate from this generic approach and introduce more dev work.
  - Do you have ideas?
- Besides the about page I can't think of another that would need overriding. At least for the time being, but maybe they'll show up in the future.
- I'd avoid introducing more complexity in the config structure.

PAGE_OVERRIDES.md
hanbyul-here: but developers (or anybody who wants to override) still need to start from defining overrides in delta.config? Let's say somebody decides to override about page.  I think it is unlikely that an editor is going to set up `delta.config`, then ask a developer to write a `ComonentOverride`.  It would be helpful for a developer to be able to see a whole workflow.

hanbyul-here:
Yeah I like how you described how 'contents' are different from 'override component'. This will be a helpful addition to the documentation I think. I think we can leave it with an explanation now, we can always work more in the future :)

PAGE_OVERRIDES.md
hanbyul-here: I see what you mean. As far as I can understand, in creator scenario, a developer still needs to edit `mock/delta.config.js` to test with what I am writing, right? I think this step is missed from this doc.

danielfdsilva: yes. that is correct. I'll look into adding a little info on that

danielfdsilva: @hanbyul-here Added a piece around testing

hanbyul-here (approved):

PR #195, opened 2022-06-29
Add redirect file for routing in Netlify

hanbyul-here:
Redirect all the subrouting to index.html so react-router can handle routing


danielfdsilva (approved):
Pretty handy to have previews for delta ui.
Thanks for setting up

PR #196, opened 2022-06-29
Tweak scrolly map threshold

hanbyul-here:
Make the layer change happen bit earlier so the text and the layer more align together.


danielfdsilva:
This feels better indeed

danielfdsilva (approved):

Issue #199, opened 2022-07-01
Add globe projection support


danielfdsilva:
With mapbox officially [releasing the globe projection](https://www.mapbox.com/blog/globe-view/?utm_campaign=email%7Cblast%7Cglobe-launch-weather&utm_content=Globe+view+announcement+1&utm_medium=email_action&utm_source=customer.io) we can start looking into implementation.

I've done some quick tests, and I'm now looking for a proper implementation.

![image](https://user-images.githubusercontent.com/1090606/176933867-a4a6737e-194b-446b-9a40-826f2b3444d9.png)

![image](https://user-images.githubusercontent.com/1090606/176933889-5645af2e-9045-40cf-a31d-b5436267f814.png)

## Acceptance Criteria

- [ ] A proof of concept that showcases how users might select from available mapbox projections in the map explorer and discovery maps views
- [ ] A proof of concept that showcases how users might select default projections during configuration of datasets and discoveries

aboydnw:
@danielfdsilva @hanbyul-here @ricardoduplos I added some AC here, feel free to add or change them if I misunderstood from the conversation today

PR #201, opened 2022-07-04
Projection options

danielfdsilva:
Following the initial work on #199, I looked into adding projection support to the maps.

## Explore page
In the explore page the user can use the controls on the map to change the projection (currently uses a globe icon, but may need to change it. cc @ricardoduplos )

image

When a conic projection (like Albers or Lambert Conformal Conic) is selected, the user is able to change the default values for the Center and S/N Parallels.

image

Upon selecting a projection, it gets stored in the url so that the current view can be easily shared with other users.

The user configuring a dataset can also decide to customize the projection to use for that dataset. This is one through the `projection` property. The `name` is always required. If using a conic projection they must also specify the `center` and `parallels`.

```yaml
projection:
  name: 'lambertConformalConic'
  center: [0,30]
  parallels: [30,30]
```

**Notes on the behavior**:
- If a layer has a default projection defined in the mdx file, it will be enabled when the user selects the layer from the panel.
- The user is then free to alter the projection via the dropdown.
- If they decide to share a dataset after selecting a different projection, that is what the recipient of the url will see, effectively ignoring the configured projection - This was done to allow the user to share different views.

This can be tested in the [sandbox dataset](http://localhost:9000/covid-19/datasets/sandbox/explore)

## Map blocks

The map blocks can be setup with a starting projection using the `projectionName`, `projectionCenter`, and `projectionParallels`:

```jsx
<Map
  datasetId='sandbox'
  layerId='no2-monthly'
  // ... other props
  projectionName='globe'
/>
```

By default the map blocks do not allow the user to change the projection, but this interaction can be enabled with the `allowProjectionChange` prop.

This can be tested in the [MDX sandbox page](http://localhost:9000/sandbox/mdx-page)

## Scrollytelling

The scrollytelling chapters can define a projection in which to view the map using the `projectionName`, `projectionCenter`, and `projectionParallels` props:

```jsx
<Chapter
  center={[0, 0]}
  zoom={2}
  datasetId='no2'
  layerId='no2-monthly-diff'
  datetime='2021-01-01'
  projectionName='equirectangular'
>
  <Chapter
    center={[-77, 38.90]}
    zoom={10}
    datasetId='no2'
    layerId='no2-monthly-diff'
    datetime='2021-01-01'
  >
```

Once a chapter with a set projection is reached, that projection will be used on subsequent chapters, until one specifies a different projection. In the code above the second chapter will still use the `equirectangular` projection.

This can be tested in the [scrollytelling sandbox page](http://localhost:9000/sandbox/mdx-scrolly)


app/scripts/components/common/mapbox/projection-selector.tsx
```diff
@@ -0,0 +1,299 @@
...
+// .mapboxgl-ctrl button:not(:disabled):hover {
+//   background-color: rgba(0, 0, 0, 0.05);
+// }
+const SelectorButton = styled(Button)`
+  &&& {
+    ${createButtonStyles({ variation: 'primary-fill', fitting: 'skinny' })}
+  }
+`;
+
+export type Projection = Exclude<
```
hanbyul-here, line 67: kind of nitpick It might be beneficial to give more specific name to projection interface? It follows the same pattern with the component which can be confusing.

hanbyul-here:
Wahh this is exciting 👯 I will wait for @ricardoduplos 's review before going through the final review.
This might be a bit out of scope for this PR, but I wonder we have a clear standard for what gets reset vs not reset when layer is changed. Or is this something we should actively seek for feedback from users?

app/scripts/components/common/mapbox/projection-selector.tsx
danielfdsilva: Not sure I understand "follows the same pattern with the component". Can you clarify?

Would you have a naming suggestion?

hanbyul-here: I meant the same naming pattern (PascalCase). I was scamming typescript interface naming recommendation and it seems recommended to think about [why this interface exists](https://google.github.io/styleguide/tsguide.html#naming-style) . so maybe `ProjectionOptions`?

danielfdsilva: Ok, that makes sense!

danielfdsilva:
@hanbyul-here Name updated!
I do agree that we should have a standard, and I guess that different users will see this differently, so some feedback would be good.

hanbyul-here (approved):
One thing I am bit unsure is how much value `allowProjectionChange` add. I feel on both ways that 
- user will only pass the projection when they want to change the projection, so we can just use `projectionName` to detect if user wants to change the projection or not. 
- but then it might be better to be exclusive?
I will leave this up to you.

hanbyul-here (commented):
Screen Shot 2022-07-08 at 11 13 34 AM

Sorry, I just noticed this now. I see mdx map's compare layer on https://deploy-preview-201--veda-ui.netlify.app/sandbox/mdx-page mdx page is not loading? edit: that was because the layer doesn't have compare layer defined 😓 I made a small pr to fix it: https://github.com/NASA-IMPACT/delta-ui/compare/feature/projections...fix/sandbox-mapblock?expand=1

hanbyul-here:
I think this change is worth running with @karitotp .
Karito, if you have time, can you test changing projections with Globe button on the map on this page https://deploy-preview-201--veda-ui.netlify.app/agriculture/datasets/sandbox/explore?position=0%7C0%7C1 ? and let us know if you find any unexpected behavior?

danielfdsilva:
@hanbyul-here Regarding "much value allowProjectionChange add":
Maybe this was not clear, but the `allowProjectionChange` is to allow the user using the app to change the projection (via the switcher button). The configurator user may decide to use a different projection (ex globe), but do not allow the user browsing the app to switch to something else. That's where `allowProjectionChange` comes into play.

karitotp:
Sorry for the delay here, @hanbyul-here, I did a review and I found the following:

- The "Household and Disability Score" layer is loading for default in Lambert Conformal Conic projection, while the other layers load for default in Mercator projection.

Screenshot from 2022-07-12 23-01-41

- The HLS layer is not loading.

- In some projections in the Nighlight Montly layer, one action is producing that the icons is displayed out of the map like below images, I tried to find what is producing this but I did not find the exactly reazon, but what I saw mostly is that it is after do a zoom in on the map and then change to one of the projection of the Nighlight Montly layer, either by zoom in directly on the map or by clicking on one of the icons on the map like in the layer Household and Disability Score, then, in that zoom, changing to the Nighlight Montly layer and select a projection like Albers or one of the other options will display the icons out of the map.

Screenshot from 2022-07-14 09-08-37

danielfdsilva:
I was able to understand why this happens and how to replicate.
In the `mercator` projection multiple world copies are rendered by default so you can scroll sideways and the "world won't end". When you start moving sideways the marker position is updated automatically so that they remain on screen. Mapbox does this by changing the longitude value of the marker so that the marker appears in the "next" world. This results in longitudes higher than 180º.

This is good for `mercator` with multiple world copies, but when switching to a projection without copies (like any other) the markers are rendered outside the map, since mapbox is not updating the marker positions. I think this is a mapbox bug.

A quick video to demonstrate:

https://user-images.githubusercontent.com/1090606/179202031-f5597906-eb57-43df-979c-86b9c243e6b4.mov

### Possible solutions
- The easiest one would be to disabled world copies for mercator, effectively preventing marker jumping
- We could also reset the marker's position manually when the projection is changing, but this is harder to implement.

@hanbyul-here thoughts on the approach?

ricardoduplos (approved):
Made a few tweaks. This is good to go.

danielfdsilva:
@hanbyul-here Disabled the world copies to solve the marker issue since this is the quickest approach.
If it turns out not to be ideal we'll iterate.

PR #206, opened 2022-07-15
audit/ update some packages

hanbyul-here:
dependapot alerts have been bothering me. so just did some house keeping.

+ I noticed that one of critical error is coming from Gulp, which hasn't been updated for a while. Do we have any plan to replace Gulp in the future? @danielfdsilva ?


danielfdsilva:
@hanbyul-here Thank you for going over the dependencies.
Gulp is used as the task runner, almost only running commands. There is currently no plan to replace it. Maybe we can use the `package.json` `override` feature and upgrade the problematic package version (if compatible that is)

danielfdsilva:
While testing this I noticed that the tests are erroring because there's some sort of configuration problem. Given that this PR upgrades Jest, it would be a good idea to also fix the configuration to have everything work well.
Holding off merging until we fix that.

danielfdsilva (approved):
I updated the tests and was able to support jest 28 in the process.
All looks good. Have a final look and go for merge! :)

hanbyul-here:
Just realized that the build is broken: https://deploy-preview-206--veda-ui.netlify.app/ looking at it now.

hanbyul-here:
@danielfdsilva apology that I didn't catch it before, but build is failing on local too. Can you take a look at the problem?

PR #207, opened 2022-07-20
Make nav to reflect scroll direction

hanbyul-here:
- Make navigation react to scroll direction when there are two (global and local) navs on the page. When a user scrolls down, global nav goes away. When a user scrolls up, global nav pops in. (works the same on every screen.)
- Regrouped global nav and local nav nav as it is simpler/can be styled with css when they are grouped under one parent element. 
- The hamburger button on a smaller screen looks like below now. I was experimenting with hamburger menu and I kind of liked it? but of course I can put this back into the previous design
Screen Shot 2022-07-20 at 10 28 50 AM

- I edited aria attributes of navs since I am on it. (followed this page http://web-accessibility.carnegiemuseums.org/code/navigation/ and the example website linked on the page.)


app/scripts/components/common/nav-wrapper.js
```diff
@@ -0,0 +1,44 @@
...
+
+const AnimatedPageHeader = styled(PageHeader)`
+  transition: margin-top 0.32s ease-out;
+  margin-top: 0;
+  &.up {
+    margin-top: 0;
+  }
+  &.down {
+    margin-top: calc(
+      -2.5rem - (1rem * var(--base-space-multiplier, 1) * 0.75 * 2)
```
danielfdsilva, line 23: Can you add a little comment explaining the values, namely the `0.75 * 2`

app/scripts/components/common/nav-wrapper.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+`;
+
+function PageNavWrapper({ localNavProps }) {
+  const renderLocalNav = !!localNavProps;
+  const scrollDir = useScrollDirection();
+
+  return (
+    <NavWrapper>
+      <AnimatedPageHeader className={scrollDir} />
```
danielfdsilva, line 34: I'd like to pick your brain as to why use classes instead of styled-component props. Do you feel this performs better?

danielfdsilva (changes requested):

app/scripts/components/common/nav-wrapper.js
hanbyul-here: I covered here: https://github.com/NASA-IMPACT/delta-ui/pull/207/files#diff-53f55889f079d886355da3a9f5d993dd6052bafba682ac471d3fe410ec41743dR71 (Also I could use variableGlsp instead of raw value so it is a win ✌️ - couldn't get the nested calc before!)

app/scripts/components/common/nav-wrapper.js
hanbyul-here: No your comment does make me wonder if it is more performative to use classname than dynamic props (possibly? classname will just direct styles for existing css compare to props changing styles dynamically?) In this case, I think using props might be better to respond to scroll direction's value so I changed it to use props.

danielfdsilva (approved):

PR #217, opened 2022-07-28
Update ga package to ga4

### LINKED ISSUE #216: Fix GA  (Update GA plugin)
The current GA plugin we are using doesn't seem to support GA4, which is default now : https://github.com/react-ga/react-ga/issues/493 
In result, we have no data. We can either use the workaround mentioned in the issue, or replace the plugin to the one that supports ga4: (https://www.npmjs.com/package/react-ga4). Since Universal tracking (which is the recommended workaround) will be discontinued in a year, it will make sense to update the plugin we are using.

hanbyul-here:
Close #216


aboydnw (approved):
Did a quick smoke test in the netlify preview, everything worked.

PR #221, opened 2022-08-02
Set renderWordCopies:true for some projections

### LINKED ISSUE #220: Enable renderWorldCopies for some projections (ex. Globe, Mercator)
Globe projection prevents you from panning on a low zoom when `renderWorldCopies` is set as `false`. Some other projections look better with `renderWorldCopies: true`

hanbyul-here:
Close #220

### DIFF
--- a/app/scripts/components/common/mapbox/map.tsx
+++ b/app/scripts/components/common/mapbox/map.tsx
@@ -37,6 +37,8 @@ interface SimpleMapProps {
   onProjectionChange?: (projection: ProjectionOptions) => void;
 }
 
+const projectionsWithWorldCopies = ['globe', 'mercator', 'equirectangular'];
+
 export function SimpleMap(props: SimpleMapProps): JSX.Element {
   const {
     mapRef,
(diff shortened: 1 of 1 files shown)


danielfdsilva (changes requested):
This re-introduces the problem referenced at https://github.com/NASA-IMPACT/delta-ui/pull/201#issuecomment-1185390161

I think we may need to reposition the markers when the projection changes.

danielfdsilva:
Opened an issue in the [mapbox repo](https://github.com/mapbox/mapbox-gl-js/issues/12129)

hanbyul-here:
You are referring 'markers outside of map' problem right?  I thought limiting the options to only specific sets of projections works?  
https://deploy-preview-221--veda-ui.netlify.app/covid-19/datasets/sandbox/explore?position=124.7106%7C71.6393%7C0.67&datetime=2019-01-01T00%3A00%3A00.000Z&layer=nightlights-hd-monthly&projection=albers%7C-96%2C37.5%7C29.5%2C45.5

This preview returns a screenshot like below. Am I missing something?
Screen Shot 2022-08-02 at 12 32 25 PM

danielfdsilva:
@hanbyul-here yes. That problem happens when you switch from a projection with multiple world copies to one without. The problem is having world copies, not the other way around.
If while in mercator you move the map to the right until the markers start moving and then switch projections will see the problem.

Mapbox adds to the marker's longitude so it shows on the "next world". On projection change it should be updating the markers positions so that their longitude is within [-180, 180] again, but it is not doing that. We can either wait for a fix, or we have to implement this repositioning ourselves.
A way to achieve this could be removing and adding the markers again.

hanbyul-here:
Oh I see. closing for now then.

PR #225, opened 2022-08-03
Fix build after #206 bug

danielfdsilva:


hanbyul-here (commented):
Can you provide a brief explanation of why the build was broken?

danielfdsilva:
the `babel.config.js` was overriding the `.babelrc` with the wrong preset

hanbyul-here (approved):
aha, thanks!

Issue #219, opened 2022-08-01
Polar projection


aboydnw:
## Description

Based on feedback from EIS Freshwater when reviewing alternative projections.

> The snow people like to work in polar projection.

Note from Hanbyul in Slack:

> Oh I quickly looked up mapbox gl doc and it seems we can offer polar projection by tweaking the parameters of lambertConformalConic projection, (https://docs.mapbox.com/mapbox-gl-js/guides/projections/) maybe we can pre-bake these parameters and offer polar projection

## Acceptance Criteria
1. A user viewing the dataset explorer page can choose to view the dataset in a polar projection

danielfdsilva:
Link for `lambertConformalConic` with parameters to emulate a polar projection: https://www.earthdata.nasa.gov/dashboard/eis/datasets/tws-anomaly/explore?datetime=2002-12-01T00%3A00%3A00.000Z&position=19.5853%7C69.572%7C2.99&projection=lambertConformalConic%7C-40%2C0%7C90%2C90

PR #231, opened 2022-08-11
Add custom polar projection

danielfdsilva:
Addresses #219.

⚠️ Although the issue is relatively simple, this PR ended up being a bit complex and touching quite some files. I'll do my best to explain what is going on.

-----

Mapbox only supports a [limited list of projections](https://docs.mapbox.com/mapbox-gl-js/style-spec/projection/), none of which is a polar projection. On their [projection guide](https://docs.mapbox.com/mapbox-gl-js/guides/projections/#customize-a-conic-projection) they explain that a polar projection can be faked by tweaking the `center` and `parallels` values for the `lamberConformalConic` projection. This is the approach we are taking.

> Sidenote: I found a bug with `lamberConformalConic` that crashes mapbox when the value for parallels is `-90`. I [logged the issue](https://github.com/mapbox/mapbox-gl-js/issues/12153), and our current fix can be to use `-89.99`. Close enough and it doesn't crash.

We were rendering a list of projections in the dropdown storing them in the format used by mapbox:
```
const projectionList = [
  {
    name: string;
    label: string; // Human readable name of projection
  }
] 
```

The `name` property was being used as a unique id and passed directly to mapbox's `setProjection()` method.
The custom polar projections we want to add do not exist, but use the `lamberConformalConic` projection with special values of `center` and `parallels`. Implementing this with the current structure would result in repeated identifiers.
```js
const projectionList = [
  {
    name: 'lamberConformalConic',
    label: 'Polar North',
    center: [-40, 0],
    parallels: [90, 90]
  },
  {
    name: 'lamberConformalConic',
    label: 'Polar South',
    center: [-40, 0],
    parallels: [-89.99, -89.99]
  }
];
```

We can no longer use `name` as an identifier, and the block property `projectionName` also doesn't make sense.
The solution was to refactor this list with an internal unique identifier and the corresponding projection name for mapbox:
```js
const projectionList = [
  {
    id: 'polarNorth',
    mbId: 'lamberConformalConic', // name value to use when setting the projection
    label: 'Polar North',
    center: [-40, 0],
    parallels: [90, 90]
  },
  {
    id: 'polarSouth',
    mbId: 'lamberConformalConic',
    label: 'Polar South',
    center: [-40, 0],
    parallels: [-89.99, -89.99]
  }
];
```

In this way we can uniquely identify the projection by their `id` (including in the url for sharing purposes), and before sending it to mapbox, we get the appropriate projection name.

In the process of updating this, the code to render the different dropdown items was becoming confusing so I refactored the component to improve its clarity, by splitting it into different components/utils files.
Changing the structure and behavior of the projection's selector also meant changing the typescript types which lead to additional changes. There were also some other typescript related errors that were fixed as well (this also accounts for quite some file changes.)

The `Map` block and the scrollytelling `Chapter` block were also updated changing the projection prop from `projectionName => projectionId`.

- [ ] **TODO:** Update the documentation on [delta-config](https://github.com/NASA-IMPACT/delta-config/tree/develop/docs) if this approach is validated.

@hanbyul-here @ricardoduplos Please have an additional look to be sure nothing is breaking.


app/scripts/components/common/mapbox/projection-selector/utils.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+
+      if (
+        !parallels ||
+        !validateLat(parallels[0]) ||
+        !validateLat(parallels[1])
+      ) {
+        const o = projectionsConic.join(', ');
+        projectionErrors.push(
+          `- Invalid projectionParallels. This property is required for ${o} projections. Use [Southern parallel latitude, Northern parallel latitude].`
```
hanbyul-here, line 122: Do `polarSouth` and `polarNorth` projection also require these values from a user?

app/scripts/components/common/mapbox/aoi/mb-aoi-draw.d.ts
```diff
@@ -1,4 +1,5 @@
-import { DefaultTheme } from 'styled-components';
+import { MutableRefObject } from 'react';
```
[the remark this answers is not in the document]
danielfdsilva: Since these were just typescript types and we're not fully using typescript it ignored them and built anyway.

app/scripts/components/common/mapbox/projection-selector/utils.ts
danielfdsilva: Great catch! They're not needed. Will fix

hanbyul-here: Thanks for the fix! Do you think polar projections still need to ask `center` property from users?

hanbyul-here:
I still get an error that map block is malformed when I pass `polarNorth` as projection id to one of chapter in scrollytelling map? (https://github.com/NASA-IMPACT/delta-ui/blob/main/app/scripts/components/sandbox/mdx-scrollytelling/page.mdx) 
I changed
```
  <Chapter
    center={[-77, 38.90]}
    zoom={10}
    datasetId='no2'
    layerId='no2-monthly-diff'
    datetime='2021-01-01'
  >
``` 
to 
```
  <Chapter
    datasetId='no2'
    layerId='no2-monthly-diff'
    projectionId='polarNorth'
    datetime='2021-01-01'
  >
```

danielfdsilva:
@hanbyul-here Added some code to circumvent the overlay. There's a whole [thread about this](https://stackoverflow.com/questions/46589819/disable-error-overlay-in-development-mode) on StackOverflow

hanbyul-here:
Yeah I believe the error is from the chapter missing `center`, `zoom`. (Error being thrown from `validateChapter`: https://github.com/NASA-IMPACT/delta-ui/blob/main/app/scripts/components/common/blocks/scrollytelling/chapter.tsx#L70) 

Screen Shot 2022-08-23 at 11 16 21 AM

hanbyul-here:
After discussion:
center/zoom are always required//  They are independent from projectionCenter/projectionParallels (which shouldn't be required for polar projections.)

hanbyul-here (approved):

PR #232, opened 2022-08-23
Implement incremental date navigation and comparison behavior improvement

### LINKED ISSUE #228: Incremental date navigation and comparison behavior improvement
## Description
Related to https://github.com/NASA-IMPACT/delta-ui/issues/200

image

Note: *ignore the timeslider*

Users should be able to navigate to the next/prev date available without opening the calendar widget.
The comparison now needs to be enabled with a toggle instead of just selecting a date.
- If there's no comparison available, the toggle should not be visible
- When the comparison is activated, select the same date by default
- The comparison date should no longer be clearable from the calendar widget. To stop comparing the toggle should be used

## Acceptance criteria
- [ ] Users can use the arrows to navigate dates
- [ ] Users can enable and disable the comparison using a toggle

danielfdsilva:
Fix #228

image

### DIFF
--- a/app/scripts/components/datasets/s-explore/index.tsx
+++ b/app/scripts/components/datasets/s-explore/index.tsx
@@ -14,6 +14,7 @@ import {
   CollecticonShrinkToLeft
 } from '@devseed-ui/collecticons';
 import { ProjectionOptions } from 'delta/thematics';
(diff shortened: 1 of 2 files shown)


hanbyul-here (approved):
👏 I enjoyed the size of the pr ! :D

PR #235, opened 2022-08-30
Date time slider

danielfdsilva:
Demo at: http://localhost:9000/sandbox/dateslider


hanbyul-here:
I have general concerns about the use of animation of the slider. (This might be more of a question for @ricardoduplos ? but maybe we can discuss it together)

Some people are sensitive to animations, feeling uncomfortable/nausea. (This is why reduced motion is one of accessibility feature in MacOS.)  I wonder if we gave enough thoughts to the animation used in this slider. ex. I noticed that the slider moves pretty fast when it needs to reset. wouldn't it be better to at least selectively disable the animation using a query like ` prefers-reduced-motion `?

app/scripts/components/datasets/s-explore/timeseries/trigger-rect.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const { width, height, data, x, zoomXTranslation, zoomBehavior } =
+    useTimeseriesContext();
+  const elRef = useRef<SVGRectElement>(null);
+  const hoverDataRef = useRef(false);
+
+  // Since we're applying the zoom behavior on an invisible trigger rect, it
+  // swallows all mouse events. To be able to handle a hover state or even a
+  // click on a data point we create a list of coordinates for where the data
+  // is, and then check if a mouse event is within a given radius of them.
+  const dataHotZones = useMemo(() => {
```
hanbyul-here, line 42: I thought there must be some easy css tricks using `pointer-events` but wow it seems surprisingly tricky to propagate events from the element with higher z-index to lower one!

danielfdsilva:
@hanbyul-here You are totally right and that is a very considerate thought. It is easy to disable animations conditionally. Let's see if @ricardoduplos wants to take this a different route

danielfdsilva:
Implemented reduced motion behavior and fixed styles according to @ricardoduplos behavior.

danielfdsilva:
The timeseries slider was also added to the dataset explore page. Feel free to review

ricardoduplos (approved):

app/scripts/components/datasets/s-explore/timeseries/context.tsx
```diff
@@ -0,0 +1,35 @@
+import { ScaleTime, ZoomBehavior } from 'd3';
```
hanbyul-here, line 1: It seems all the components that use this context are children componetns of `timeseries/index.tsx`. Why did you decided to use context here?

hanbyul-here (commented):
It is a nitpick, but `timeseries` might not be the best name for the component. `timeseries` is an established name to refer a type of data, and `timeseries` here is a specific component to navigate the time of the data.  More component-like name, like `DateNavigator` or `TimeSlider` might work better.

app/scripts/components/datasets/s-explore/timeseries/context.tsx
danielfdsilva: When I stared I had a more nested structure which I was able to simplify. I decided to leave the context for scalability purposes. Would you say this complicated things and should be removed?

danielfdsilva:
@hanbyul-here Good call on the name.
Seems that I wasn't sure about it either given that I named the PR one thing, the branch another, and there were also inconsistencies in the files.

I renamed everything to `Dateslider` (slider and handles dates as opposed to time), and moved the files to `components/common`.

app/scripts/components/datasets/s-explore/timeseries/context.tsx
hanbyul-here: I think using props generally shows how the data flows better. but maybe this will get used in somewhere outside of dateslider component in analysis page? I will leave this up to you.

danielfdsilva:
I don't think context would be used outside the dateslider, and you it was indeed complicating things a bit. Removed it in favor of props.

hanbyul-here (approved):
I like how it works with reduced motion!

PR #238, opened 2022-09-01
Legend and loading

danielfdsilva:
Quick PR to fix a couple of problems I noticed while adding datasets:
- The legend had a skew and the last gradient value was being shown more prominently.
- The layer loading indicator took a moment to kick in resulting in a strange empty panel

### DIFF
--- a/app/scripts/components/common/mapbox/layer-legend.tsx
+++ b/app/scripts/components/common/mapbox/layer-legend.tsx
@@ -35,9 +35,15 @@ type LegendSwatchProps = {
 };
 
 const makeGradient = (stops: string[]) => {
-  const d = 100 / stops.length - 1;
-  const steps = stops.map((s, i) => `${s} ${i * d}%`);
-  return `linear-gradient(to right, ${steps.join(', ')})`;
+  const cCount = stops.length;
+  if (cCount === 1) {
+    return stops[0];
+  } else {
+    const [first, ...others] = stops;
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/mapbox/layer-legend.tsx
```diff
@@ -35,9 +35,15 @@ type LegendSwatchProps = {
...
 
 const makeGradient = (stops: string[]) => {
-  const d = 100 / stops.length - 1;
-  const steps = stops.map((s, i) => `${s} ${i * d}%`);
-  return `linear-gradient(to right, ${steps.join(', ')})`;
+  const cCount = stops.length;
+  if (cCount === 1) {
+    return stops[0];
+  } else {
+    const [first, ...others] = stops;
```
hanbyul-here, line 42: I think this approach is fine, but does this do anything different from adding parenthesis to the original code? like 
```
  if (stops.length === 1 ) return stops[0];
  const d =  100 / (stops.length - 1);
  const steps = stops.map((s, i) => `${s} ${i * d}%`);
  return `linear-gradient(to right, ${steps.join(', ')})`;
 ```

danielfdsilva: Oh wow. You're absolutely right. 😅
Talk about overly complicating things. Thank you

hanbyul-here (approved):

PR #239, opened 2022-09-01
Change chart library to Rechart

hanbyul-here:
changed chart library to Rechart, mainly because larger user pool, more active maintenance status, and more direct solutions to the features (ex. brush, sync multiple charts) we need to implement soon.

part of https://github.com/NASA-IMPACT/delta-ui/issues/227

You can see preview: https://deploy-preview-239--veda-ui.netlify.app/sandbox/mdx-chart

- [x] change margin/spacing depending on screen size. 
- [x] use typescript (still need some help)


app/scripts/components/common/rechart-chart/legend.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import styled from 'styled-components';
+import { glsp, themeVal } from '@devseed-ui/theme-provider';
+import { LegendProps, LineChart } from 'recharts/types';
+
+import { ListItem } from './tooltip';
+import { highlightColorThemeValue } from './constant';
+
+// NEED HELP: getting error: 'LineChart' refers to a value, but is being used as a type here. Did you mean 'typeof LineChart'
+interface ReferenceLegendComponentProps extends LineChart {
```
hanbyul-here, line 10: Can I get your help @danielfdsilva  ?
This is the `component` prop for  of  CustomizedComponent(https://recharts.org/en-US/api/Customized). I am generally not sure how to type this.

danielfdsilva: This one was not simple. Should be easier, but even looking through the source code, I did not see other options.
If you do `interface ReferenceLegendComponentProps extends CategoricalChartProps {` it works since `CategoricalChartProps` is the base for the Charts.

You'll need to import it from `import { CategoricalChartProps } from 'recharts/types/chart/generateCategoricalChart';`

app/scripts/components/common/chart/block.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [uniqueKeys, setUniqueKeys] = useState<string[]>([]);
+
+  const newDataPath = dataPath.split('?')[0];
+  const extension = fileExtensionRegex.exec(newDataPath)[1];
+
+  useEffect(() => {
+    const getData = async () => {
+      let data;
+      if (extension === 'csv') data = await csv(dataPath);
+      else data = await json(dataPath);
```
danielfdsilva, line 27: Can we make this a ternary to improve readability?

app/scripts/components/common/chart/constant.js
```diff
@@ -0,0 +1,13 @@
+export const chartHeight = '32rem';
```
danielfdsilva, line 1: Should we consider the chart's height as a ratio of its width? Something to think about

app/scripts/components/common/chart/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+  useEffect(() => {
+    if (isMediumUp) {
+      return;
+    } else {
+      setChartMargin({
+        ...defaultMargin,
+        left: 0
+      });
+    }
+  }, [isMediumUp, setChartMargin]);
```
danielfdsilva, line 93: The `setChartMargin` is stable and not needed here

app/scripts/components/common/chart/legend.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { highlightColorThemeValue } from './constant';
+
+interface ReferenceLegendComponentProps extends CategoricalChartProps {
+  highlightLabel: string;
+}
+
+const LegendWrapper = styled.ul`
+  max-width: ${(props) => props.width};
+  margin: 0 auto;
+  margin-top: ${glsp(0.6)};
```
danielfdsilva, line 18: This results in a strange value. `0.5` would be more appropriate.

app/scripts/components/common/chart/legend.tsx
```diff
@@ -0,0 +1,73 @@
...
+  font-size: 0.75rem;
+  color: ${themeVal('color.base-400')};
+  * {
+    align-self: center;
+  }
+`;
+
+const HighlightLabel = styled.text`
+  font-family: sans-serif;
+  font-size: 12px;
```
danielfdsilva, line 35: Can we change to `rem` like the `LegendItem` above?

app/scripts/components/common/chart/legend.tsx
```diff
@@ -0,0 +1,73 @@
...
+  margin-right: ${glsp(0.25)};
+  font-size: 0.75rem;
+  color: ${themeVal('color.base-400')};
+  * {
+    align-self: center;
+  }
+`;
+
+const HighlightLabel = styled.text`
+  font-family: sans-serif;
```
danielfdsilva, line 34: Is there a specific reason not to use the app's theme font-family?

app/scripts/components/common/chart/legend.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <ListItem color={entry.color} />
+            {entry.value}
+          </LegendItem>
+        ))}
+      </LegendWrapper>
+    );
+  } 
+  return null;
+};
+export default LegendComponent;
```
danielfdsilva, line 73: If a file exports multiple components is usually a good idea not to have a default one, unless there's a very clear hierarchy.

app/scripts/components/common/chart/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+export const convertToTime = ({ timeString, dateFormat}: {timeString: string, dateFormat: string}) => {
+  if (!timeString) return null;
+
+  const parseDate = timeParse(dateFormat);
+  return parseDate(timeString).getTime();
+};
+
+export function getFData({ data, idKey, xKey, yKey, dateFormat }: { data: any[], idKey: string, xKey: string, yKey: string, dateFormat: string }) {
```
danielfdsilva, line 16: Can this function be documented to better understand what it does?

app/scripts/components/common/chart/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+          [xKey]: convertToTime({
+            timeString: curr[xKey],
+            dateFormat
+          })
+        });
+      }
+    }
+
+    return acc;
+  }, {});
```
danielfdsilva, line 44: It is a bit complicated to understand the flow of this function, specially since a `.reduce` function is expected to return something. Can the be rewritten in a simpler way?

danielfdsilva (changes requested):
Not sure if it is the data or something else but the sandbox chart does not look right.

image

There are also some typescript errors:
```
app/scripts/components/common/chart/index.tsx:125:19 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.

125                   timeString: highlightStart,
                      ~~~~~~~~~~

app/scripts/components/common/chart/index.tsx:129:19 - error TS2322: Type 'string | undefined' is not assignable to type 'string'.

129                   timeString: highlightEnd,
                      ~~~~~~~~~~
app/scripts/components/common/chart/legend.tsx:48:32 - error TS2532: Object is possibly 'undefined'.

48     <g transform={`translate(${width - 100}, 0) rotate(0)`}>
                                  ~~~~~
```


-----

## Other

There are a couple of additional points that I'd like us to address.
These are mostly cosmetic but I think they help with the readability of the code and consequently make it easier to understand.

- It is helpful to have prettier and the eslint formatter integrated with VS Code since it takes care of breaking long lines and normalizing bracket spacing (among others).

- Some of these files could benefit from an empty line between instructions, making them easier to peruse.

- If you're creating components in typescript and defining prop types as interfaces, then it is not needed to have the `.propTypes', as it is duplicated information.

- When defining components it is better to export them immediately and it is clearer what is a local variable and what is exported. 
Additionally let's standardize how functional components are defined to either a formal function or an arrow function.
```
export function MyComponent() {

}

// OR

export const MyComponent = () => {

}

// Instead of

export const MyComponent = function () {

}
```

app/scripts/components/common/chart/utils.ts
hanbyul-here: I added comments, I hope it helps. The function have gotten less readable because it was trying to avoid 
 additional iterations. 

If you have a better idea how to format `[{year: 2005, no2: 10}, {year: 2005, so2: 5}] ` to `[{year:2005, no2:10, so2:5]` (key names - year(xKey), no2, so2(uniqueKeys- driven from idKey) - are dynamic, coming from data), I would love to hear!

+ The function would look like below with `find` method. 
```
  const fData = data.reduce((acc, curr) => {
    // check current xKey value is already in acc
    const currentTime =  convertToTime({
      timeString: curr[xKey],
      dateFormat
    });
    const currentXKeyObjectIdx = acc.findIndex(e => e[xKey] == currentTime);
   // if not, push with xkey along with other properties
    if (currentXKeyObjectIdx < 0) {
      acc.push({
        [xKey]: convertToTime({
          timeString: curr[xKey],
          dateFormat
        }),
      [curr[idKey]]: parseFloat(curr[yKey])
    });
  } else {
   // if there is already an object with current xkey value, replace it to an object with an additional property
    const currentXKeyObject = acc[currentXKeyObjectIdx];
    acc[currentXKeyObjectIdx] = {
      ...currentXKeyObject,
      [curr[idKey]]: parseFloat(curr[yKey])
    };
  }
  return acc;
  },[]);
```

hanbyul-here:
Lint integrated into the editor is the only way that I can keep up with lint rules! That being said, is there any way that you can integrate the styles you would like to see in our lint rules? ex. Is [this function component definition rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md) something you want to incorporate?

danielfdsilva:
That rule is interesting. I'll have a look.
We can try to find rules for these things, but for some (like whitespace) I'm not sure they exist.

app/scripts/components/common/chart/utils.ts
danielfdsilva: What I'm seeing from the data and what you said, it is a matter of grouping the data by year (the xKey), and merging the other properties.

I'd implement it as follows:
```js
// Group data by the xKey.
  const resultByXKey = data.reduce((acc, entry) => {
    // ## From:
    // {
    //   "xkey": xKey value,
    //   "yKey": yKey value,
    //   "idKey": idKey value
    // }
    // ## to
    // {
    //   "xkey": xKey value,
    //   "idKey value": yKey value
    // }
    const formattedObj = {
      // [xKey]: entry[xKey], // year not formatted
      [xKey]: convertToTime({
        timeString: entry[xKey],
        dateFormat
      }),
      [entry[idKey]]: parseFloat(entry[yKey])
    };

    const groupValue = formattedObj[xKey];

    if (acc[groupValue]) {
      return {
        ...acc,
        [groupValue]: {
          ...acc[groupValue],
          ...formattedObj
        }
      };
    } else {
      return {
        ...acc,
        [groupValue]: formattedObj
      };
    }
  }, {});
  
  const results = Object.values(resultByXKey);
```

This works well for the 1st chart, but returns different values for the 2nd.
This code returns results for the year 2022 (which are present in the source), but your code doesn't. Not sure what is expected.

hanbyul-here: Ah it is great idea to use `Object.values`. 
I think the reason the data didn't have 2022 entries was the logic I put was eliminating the data point that doesn't have all the idKeys. (To put more simply, not all the counties have data for 2022)
I reflected your comment into the code!

app/scripts/components/common/chart/constant.js
hanbyul-here: Yeah that is a good idea especially we will have some smaller charts for analysis page. I will do follow up about it.

PR #240, opened 2022-09-07
Lint files

danielfdsilva:
As part of #239 I added the eslint rule and formatted the files. Most of the cosmetic options (like parameter spacing) were achieved automatically using VS Code "Format Document" which integrates with eslint.


hanbyul-here (approved):
thank you!

PR #243, opened 2022-09-09
Lazy load MdxContent to improve bundle splitting

danielfdsilva:
image

The `root-about` file was including mapbox in its bundle.
This happens because parcel will bundle the module with the first use. The about page allows for overrides using MdxContent which loads the Map block and so mapbox ended up in this bundle.
I added a lazy load for the MdxComponent thus forcing a split.

image

Interesting catch @hanbyul-here

### DIFF
--- a/app/scripts/components/about/index.js
+++ b/app/scripts/components/about/index.js
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { lazy } from 'react';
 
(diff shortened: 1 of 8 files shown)


hanbyul-here (approved):
Ahh good to understand why this was happening. Thanks for debugging!

PR #244, opened 2022-09-12
Update AOI tool to allow drawing polygons

danielfdsilva:
http://localhost:9000/sandbox/aoi


app/scripts/components/common/mapbox/aoi/mb-aoi-draw.js
```diff
@@ -2,9 +2,7 @@ import { useEffect, useRef } from 'react';
 import MapboxDraw from '@mapbox/mapbox-gl-draw';
 import { css } from 'styled-components';
 
-import * as RestrictedRectangleMode from './restricted-rect-mode';
```
hanbyul-here, line 5: This file doesn't seem to be used any more. Do we want to delete this file?

danielfdsilva: Sorry, forgot it there

hanbyul-here (approved):

PR #246, opened 2022-09-14
Brush, Dynamic data for chart component

hanbyul-here:
- [x] (proof of concept for) Dynamic lines
- [x] Brush

Made chart component to render brush with data preview, also to render dynamic data. I did not polish the control uis for dynamic lines since that would be better done when we start putting things into the scaffolding.


app/scripts/components/sandbox/analysis-chart/index.js
```diff
@@ -1,18 +1,86 @@
...
+  background-color: lightgrey;
+  margin-bottom: 50px;
+  padding: 20px;
+`;
+
+// manually collected data from the endpoint below for testing purpose. date ranges from 202203 to 202205
+// https://staging-raster.delta-backend.com/cog/statistics?url="s3://veda-data-store-staging/geoglam/CropMonitor_${date}.tif"1
+
+const dataForChart1 = {
+  timeSeriesData: timeSeriesData345,
```
hanbyul-here, line 33: @danielfdsilva as we discussed before, feel free to suggest how the data should be structured.

danielfdsilva: Left a suggestion on https://github.com/NASA-IMPACT/delta-ui/pull/258

danielfdsilva (approved):
Looks good. We can incorporate the data changes in another pr

PR #254, opened 2022-09-16
Handle single data point on dateslider

danielfdsilva:
image

### DIFF
--- a/app/scripts/components/common/dateslider/data-points.tsx
+++ b/app/scripts/components/common/dateslider/data-points.tsx
@@ -143,22 +143,23 @@ export function DataPoints(props: DataPointsProps) {
   );
 }
 
-
 interface DataLineProps {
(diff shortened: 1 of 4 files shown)


hanbyul-here:
I was checking sandbox page, and I realized that single date slider is displaying year 2020, when the sample data only goes back to 2021. Probabaly because `Date` gets translated into my timezone.

Screen Shot 2022-09-16 at 2 49 43 PM

I am assuming.. this won't be the problem on the explorer page since the date comes as string forms from stac endpoints?

danielfdsilva:
@hanbyul-here Yes, it has to do with TZ, and for the datasets that is already handled. Nevertheless, I updated the sandbox to handle the date correctly.

danielfdsilva:
@hanbyul-here Merging since only the sandbox was being impacted and is now fixed

PR #256, opened 2022-09-20
React query to v4

danielfdsilva:
React Query V4 was released not long ago. This PR upgrades to v4 following their [migration guide](https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4).


danielfdsilva:
The only network request we're doing so far are for the dataset explorer. Everything seemed to be working but an additional pair of eyes would be good.

PR #258, opened 2022-09-21
Timeseries requester functionality

danielfdsilva:
@hanbyul-here Here's the draft of the requester functionality as we discussed. I changed it to emit an event instead of having a callback and looks cleaner. Let me know your thoughts.

Also, here's the response format. Let me know if you need more info and whether this is helpful and simple/clean enough:

_⚠️ some props omitted for brevity_

```json
{
  "id": "no2-monthly",
  "name": "No2",
  "status": "succeeded",
  "meta": {
    "total": 3,
    "loaded": 3
  },
  "data": {
    "isPeriodic": true,
    "timeDensity": "month",
    "domain": [
      "2020-01-01T00:00:00Z",
      "2022-01-01T00:00:00Z"
    ],
    "timeseries": [
      {
        "date": "2022-01-01T00:00:00Z",
        "min": -3723984948428800,
        "max": 39240637987422210,
        "mean": 383672228362386.6,
        "the other statistic props": "..."
      },
      {
        "date": "2021-12-01T00:00:00Z",
        "min": -4058248831303680,
        "max": 34767369419096064,
        "mean": 434216209572398.4,
        "the other statistic props": "..."
      },
      {
        "date": "2021-11-01T00:00:00Z",
        "min": -3793642070212608,
        "max": 32763322646396930,
        "mean": 439771099336671.06,
        "the other statistic props": "..."
      }
    ],
    "error": null
  }
}
```


app/scripts/components/sandbox/request/index.js
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    requester.on('data', (data, index) => {
+      setRequestStatus((dataStore) =>
+        Object.assign([], dataStore, {
+          [index]: data
+        })
+      );
+    });
+  };
```
danielfdsilva, line 46: In the analysis page this will be achieved with a useEffect instead of a callback from a button click. This is just for demo purposes.

hanbyul-here:
Some kinds of formatting/processing would be needed (ex. date needs to be Y/m format? or Y/m/d? depending on the granularity I guess.) but the data format looks  👍

danielfdsilva:
@hanbyul-here I rebased this on top of the work @ricardoduplos is doing on `feature/scaffold-analysis`. While he's not finished with the layout there are things that look visually broken. I'll let you coordinate the next updates.

app/scripts/components/analysis/results/use-analysis-params.ts
```diff
@@ -0,0 +1,132 @@
...
+          'Invalid start or end date:',
+          {
+            start,
+            end
+          }
+        ];
+      }
+
+      // Create an array with all the dataset layers.
+      const allDatasetLayers = Object.values(deltaDatasets).flatMap(
```
hanbyul-here, line 78: Is it ok to aggregate all the bands? (I am assuming DeltaDatasets come as an object with band index as key like {1: []..2:[]..}. please correct me if I am wrong!)  I was not sure if we need to treat each band separately or not.

hanbyul-here:
FWIW, pushed a testing chart with requests, you can check in this preview (Chart will only show up when the request is succeeded) : https://deploy-preview-270--veda-ui.netlify.app/sandbox/request
Screen Shot 2022-10-07 at 10 52 40 AM

app/scripts/components/analysis/results/use-analysis-params.ts
danielfdsilva: Actually I was just returning the data from band 1.
I do not know if there's a specific band to use, but since in some datasets we're using band index 1, I defaulted to that.

danielfdsilva:
@hanbyul-here Chart is looking good!
I noticed you had branched from a old version of this branch (before rebasing), so I updated the `test-request` branch to include the new code. No major changes though.

Issue #259, opened 2022-09-22
Module not found and MDX content


danielfdsilva:
It may happen that while you're updating the content MDX files (datasets, thematic areas, discoveries) you refresh the page and are presented with an error message similar to the following:

![image](https://user-images.githubusercontent.com/1090606/191801797-c467cb0e-7d18-4d10-b672-bb54f4d12d98.png)

Instead of this image you can also get our standard critical error message (with the dashboard design), but if you expand the error details you'll still see a "Error: Cannot find module" message.

## The problem

This only happens if you're editing the frontmatter of one of the files (updating the MDX content does not cause the same problem). You make an edit and the page does not update ([this is a parcel bug](https://github.com/parcel-bundler/parcel/issues/8090), that I don't know if it's related or not), you refresh and the page errors.

You can try this by looking at the homepage, editing the title of a thematic area, and you'll see that the hot reload kicks in (Console gets cleared), but the title does not update. If you refresh the page the app will crash with that error. The only solution is to stop the server and restart.

## What we know so far

This is caused by the our virtual module `delta/thematics` in conjunction with react's lazy loading.

If the `delta/thematics` is used both in a lazy loaded module and outside it, then it crashes, otherwise it doesn't.

### Examples

a.js
```js
import deltaThematics, { thematics } from 'delta/thematics';

// more code
```

b.js
```js
import deltaThematics, { thematics } from 'delta/thematics';

// more code
```

main-crashing.js
```js
// Different load will crash.
import A from '$components/a';
const B = lazy(() => import('$components/b'));
```

main-working.js
```js
// Same load. Works fine.
const A = lazy(() => import('$components/a'));
const B = lazy(() => import('$components/b'));
```

main-working2.js
```js
// Same load. Works fine.
import A from '$components/a';
import B from '$components/b';
```

It seems that the bundle splitting generates `delta/thematics` with different internal ids.

danielfdsilva:
#306 seems to have fixed this, but reopening until tests are conducted.

danielfdsilva:
This issue never came up again. Closing

PR #260, opened 2022-09-22
Set parcel-resolver-thematics filepath to a real file - solve mdx module not found

danielfdsilva:
I did not do extensive testing but this seems to solve the problem outlined in #259.

I still don't know exactly why, but setting the virtual module filepath to an actual file with actual module content seems to make a difference, even though this is not what a **virtual** module is supposed to do.


hanbyul-here (approved):
oh this PR does seem to fix the problem..! 
I am also not sure the reasons behind.. but I think this will enhance editor's flow 🙏

PR #261, opened 2022-09-22
export chart as jpg 

hanbyul-here:


app/scripts/components/common/chart/analysis/index.tsx
```diff
@@ -26,6 +26,7 @@ export default function AnalysisChartProps(props: AnalysisChartProps) {
       chartData={chartData}
       renderLegend={false}
       renderBrush={true}
+      renderExport={true}
```
hanbyul-here, line 29: I have a question for you @danielfdsilva .

so I was wondering if it will be better to bring up `<ExportPNG />` component to this component, so ExportButton doesn't have to be affected by chart's wrapper style. so I did something like``

```
// chart/index.tsx
LineChartWithRef = React.forwardRef(RLineChart);
```
```
// chart/analysis/index.tsx
const chartRef = useRef(null);
<>
<LineChartWithRef ref={chartRef} ... />
<ExportPNG svgWrapperRef={chartRef} />
</>
```
But `ExportPNG` component doesn't seem to be able to get the update of `chartRef` 🤔 I couldn't really find a solution about it. I think current approach is ok, but I would like to hear how you would solve this?  (and if you think this is a better spot for `ExportPNG` component?)

danielfdsilva (changes requested):
I did a more in depth review and here are some thoughts:

### Export button
I totally agree that the export button should be separate from the chart.
You can create a ref with `useRef` and pass it to both the chart and the button component (Like your example). I'd just use a property different that `ref` for the chart, for example: `chartRef`.

However I'd encapsulate the export code in a function to be called by any button. Something like `exportChartToImg({ svgRef })`. Expanded on next section.


### Function instead of hook
I'd recommend the use of a function instead of a hook to generate the export image.
There doesn't seem to be a need to constantly generate the export image for all the charts, since this has performance impact. This seems to be an action that can be done on demand after the user requests it (ex: clicks a button).

I'd convert `useChartToPNG` to a non hook approach where you just run things sequentially without the need to set states. If by any reason you need this to run on a hook, you can always call it within a useEffect.


### Use react to create elements
I'd recommend the use of React and its `renderToStaticMarkup` function to create and manage the code for the image legend. There's nothing wrong with the current approach, but using react would keep things cleaner and more readable.
You can use components and only render to string at the last minute.
Here's an example:

```jsx
// import { renderToStaticMarkup } from 'react-dom/server';

export function getLegendStringForScreenshot({ uniqueKeys, lineColors }) {
  const legendWidth = 80;
  const legendHeight = 40;

  return renderToStaticMarkup(
    <g width={legendWidth * uniqueKeys.length} height={legendHeight}>
      {uniqueKeys
        .filter((k) => k.active)
        .map((entry, idx) => (
          <g key={entry.label} transform={`translate(${idx * legendWidth}, 10)`}>
            <rect width={12} height={12} fill={lineColors[idx]} />
            <text x={20} y={10}>
              {entry.label}
            </text>
          </g>
        ))}
    </g>
  );
}
```

It also seems that you could move the `getLegendStringForScreenshot` logic to the export function itself. I don't understand exactly why this is being passed via the button, but maybe there's a reason that I'm not seeing.


### Legend not showing on Firefox
The legend is being displayed correctly on chrome, but it doesn't show up on firefox.
After some debugging, it seems that it happens because the legend SVG doesn't have a `width/height` set, and the canvas context can't draw it.

### Other
- My editor is complaining about some typescript errors. Not sure you were still going to address them, just letting you know.
- I think we should make the exported chart pretty big as to not lose quality, perhaps even 1920px wide.

Let me know if there's anything you want to go over together.

hanbyul-here:
I believe I handled all the concerns. I will merge it.

PR #263, opened 2022-09-28
Sync charts on analysis page

hanbyul-here:
Opening to share the preview
https://deploy-preview-263--veda-ui.netlify.app/sandbox/analysis-chart

I just noticed that Recharts tries to sync brush too, which doesn't seem to make too much sense on our page. 🤔 I wonder if I can turn this feature off only for Brush, or maybe it is fine? not sure.


hanbyul-here:
Another thing I noticed while trying to line up the charts: If the chart gets smaller than a specific size, the tooltip becomes so big that it interferes with the interaction on the chart. A good thing to keep in mind: 

Screen Shot 2022-09-28 at 2 47 53 PM

aboydnw:
@hanbyul-here just a note for us to watch out for feedback as we roll this out, do we want charts to sync in every use case? aka are there potentially discoveries where scientists won't want the charts to sync.

hanbyul-here:
@aboydnw We are handling charts on the analysis page separately from other charts.  Charts embedded in discoveries ex. the ones on this page https://deploy-preview-263--veda-ui.netlify.app/sandbox/mdx-chart  will keep their current behavior.

package.json
```diff
@@ -136,7 +136,7 @@
     "react-router": "^6.0.0",
     "react-router-dom": "^6.0.0",
     "react-transition-group": "^4.4.2",
-    "recharts": "^2.1.13",
+    "recharts": "2.1.12",
```
hanbyul-here, line 139: pinned version to 2.1.12 for now because of this tooltip rendering issue: https://github.com/recharts/recharts/issues/2943

danielfdsilva (changes requested):
The charts are synchronizing correctly but there are 2 behaviors that are strange:

1) If you resize the brush on one of the charts, the other charts will update but their brush will stay the same size. Not sure if all the brushes should update or if the brushes should be something not synced.

2) The second is a bit harder to explain but using the image may help:
image

If we change the brush size on chart B so that only 1 data point is visible it will synchronize chart A, but since the point for chart B (`2021/12`) is not available on chart A, this one stays on `2021/03`. If we then hover chart A, it will activate chart B as well, but on a different date.

cc @ricardoduplos to do additional UX checks

hanbyul-here:
Ah right, I talked about Brush briefly with @ricardoduplos before, and then kind of forgot about it. Current behavior doesn't make sense in many ways but it's been hard for me to think what is synced brush with different timeframes/data points. I will make Brush behavior constant with tooltip and we can prob start from there. (Also thanks for the other comments! will handle them too..)

app/scripts/components/common/chart/tooltip.tsx
```diff
@@ -38,23 +38,25 @@ const TooltipItem = styled(ListItem)`
...
+          <strong>{timeFormatter(label, dateFormat)}</strong>
         </div>
-        {Object.keys(payload[0].payload)
-          .filter((key) => key !== xKey && !inactiveKeys.includes(key))
-          .map((key, idx) => {
-            const point = payload[0].payload[key];
+        {uniqueKeys
+          .map((e,idx) => ({label: e.label, color: colors[idx]}))
+          .filter((key) =>!inactiveKeys.includes(key.label))
+          .map((key) => {
```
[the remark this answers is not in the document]
hanbyul-here: Maybe, but I couldn't think how since I will lose the index of the key after filtering. All of this made me think It is better to bind the colors to unique key. so I moved the logic to higher level (in analysis/index)

app/scripts/components/common/chart/index.tsx
```diff
@@ -199,13 +221,32 @@ function RLineChart(props: RLineChartProps, ref: RefObject<HTMLDivElement>) {
...
-              lineColors,
-              dateFormat
-            })}
+            <Brush
+              data={chartData}
+              dataKey={xKey}
+              height={brushHeight}
+              tickFormatter={(t) => timeFormatter(t, dateFormat)}
+              onChange={handleBrushChange}
+              startIndex={brushStartIndex}
```
hanbyul-here, line 230: I thought setting index with state value will fix Brush Handle problem but it doesn't. :/ 
I looked up the repo and this bug is supposed to be fixed: https://github.com/recharts/recharts/issues/1796 but I still see the problem (ex. simplified demo that I copied and tweaked: https://codesandbox.io/s/rechart-brush-sync-z9bk04) I will file an issue, but not sure if I can do anything more about it.

danielfdsilva (approved):
Please open an issue with the brush problem and let's merge!

hanbyul-here:
I opened an issue and forgot to update here: https://github.com/recharts/recharts/issues/3027 !

PR #266, opened 2022-09-30
Disable date picker when there's only 1 date

danielfdsilva:


hanbyul-here:
lgtm 👍 
Screen Shot 2022-09-30 at 10 22 14 AM

hanbyul-here (approved):

PR #269, opened 2022-10-07
Number in scientific notation format

hanbyul-here:
The number from statistic seems to be quite big, scientific notation would help


danielfdsilva (changes requested):
This is a good call!

We already have some functions to handle large numbers and rounding in `app/scripts/utils/format.ts`. Perhaps they could be used here?

danielfdsilva (approved):
Good to go

PR #271, opened 2022-10-10
TS errors

danielfdsilva:


nerik:
Looking good, thank you. 

I'm wondering if at some point that could be interesting to enable type checking at build time (see https://parceljs.org/languages/typescript/#experimental-validator-plugin)

nerik (approved):
Looking good, thank you.

I'm wondering if at some point that could be interesting to enable type checking at build time (see https://parceljs.org/languages/typescript/#experimental-validator-plugin)

danielfdsilva:
@nerik Very interesting. A bit on the fence because of the experimental plugin, but I think we can give it a try!
I'm merging this one, but we can set up a PR for that :)

PR #272, opened 2022-10-10
zoom to bounds

nerik:
This is my attempt at fixing https://github.com/NASA-IMPACT/delta-ui/issues/257
Probably way off, but one needs to open a first PR at some point :)

### DIFF
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -37,10 +37,16 @@ export type StacFeature = {
   bbox: [number, number, number, number];
 };
 
+enum STATUS_KEY {
+  Global,
+  Layer,
+  StacSearch
+}
+
 type Statuses = {
-  global: ActionStatus;
-  layer: ActionStatus;
-  'zoom-markers': ActionStatus;
+  [STATUS_KEY.Global]: ActionStatus;
+  [STATUS_KEY.Layer]: ActionStatus;
+  [STATUS_KEY.StacSearch]: ActionStatus;
 };
 
 export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
@@ -69,9 +75,9 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
   // A raster timeseries layer has a base layer and may have markers.
   // The status is succeeded only if all requests succeed.
   const statuses = useRef<Statuses>({
-    global: S_IDLE,
-    layer: S_IDLE,
-    'zoom-markers': S_IDLE
+    [STATUS_KEY.Global]: S_IDLE,
+    [STATUS_KEY.Layer]: S_IDLE,
+    [STATUS_KEY.StacSearch]: S_IDLE
   });
 
   const changeStatus = useCallback(
@@ -80,17 +86,17 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
       context
     }: {
       status: ActionStatus;
-      context: 'zoom-markers' | 'layer';
+      context: STATUS_KEY.StacSearch | STATUS_KEY.Layer;
     }) => {
       // Set the new status
       statuses.current[context] = status;
 
       const layersToCheck = [
-        statuses.current['zoom-markers'],
-        statuses.current.layer
+        statuses.current[STATUS_KEY.StacSearch],
+        statuses.current[STATUS_KEY.Layer]
       ];
 
-      let newStatus = statuses.current.global;
+      let newStatus = statuses.current[STATUS_KEY.Global];
       // All must succeed to be considered successful.
       if (layersToCheck.every((s) => s === S_SUCCEEDED)) {
         newStatus = S_SUCCEEDED;
@@ -107,8 +113,8 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
       }
 
       // Only emit on status change.
-      if (newStatus !== statuses.current.global) {
-        statuses.current.global = newStatus;
+      if (newStatus !== statuses.current[STATUS_KEY.Global]) {
+        statuses.current[STATUS_KEY.Global] = newStatus;
         onStatusChange?.({ status: newStatus, id });
       }
     },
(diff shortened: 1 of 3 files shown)


nerik:
Moved the fetch to the Stac search endpoint up so that it is called for layers with both markers or no markers. The bounding box considered for the layer will now be made of all bboxes in the case of having markers.

app/scripts/components/common/mapbox/layers/utils.ts
```diff
@@ -258,3 +262,95 @@ export function resolveLayerTemporalExtent(
...
+    Number.NEGATIVE_INFINITY,
+    Number.NEGATIVE_INFINITY
+  ] as [number, number, number, number];
+  features.forEach((feature) => {
+    if (feature.bbox[0] < mergedBBox[0]) mergedBBox[0] = feature.bbox[0];
+    if (feature.bbox[1] < mergedBBox[1]) mergedBBox[1] = feature.bbox[1];
+    if (feature.bbox[2] > mergedBBox[2]) mergedBBox[2] = feature.bbox[2];
+    if (feature.bbox[3] > mergedBBox[3]) mergedBBox[3] = feature.bbox[3];
+  });
+  return mergedBBox;
```
danielfdsilva, line 331: I'm often torn between a more imperative approach, and one where you exclusively use high-order function without any sort of variable reassignement.

Modifying a variable makes code easier to read, but forces you to keep track of where a variable is modified (especially tricky with `const` arrays and objects where you modify the contents.)

The same approach without var reassignment would be:
```js
return features.reduce(
    (acc, feature) =>
      acc.map((mergeCoord, idx) => {
        const featCoord = feature.bbox[idx];

        if (idx < 2) {
          // SW
          return featCoord < mergeCoord ? featCoord : mergeCoord;
        } else {
          // NE
          return featCoord > mergeCoord ? featCoord : mergeCoord;
        }
      }),
    mergedBBox
  );
```

- However, isn't this approach harder to reason about?
- Should it be argued that modifying variables is ok for a simple function but a behavior to be avoided in general?
- How do we draw the line?
- Is there a happy middle ground? Like:
```js
return features.reduce(
    (acc, feature) => [
      feature.bbox[0] < acc[0] ? feature.bbox[0] : acc[0],
      feature.bbox[1] < acc[1] ? feature.bbox[1] : acc[1],
      feature.bbox[2] > acc[2] ? feature.bbox[2] : acc[2],
      feature.bbox[3] > acc[3] ? feature.bbox[3] : acc[3]
    ],
    mergedBBox
  );
```

@nerik Would like your thoughts.
cc @hanbyul-here

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -114,9 +71,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
   const statuses = useRef<Statuses>({
     global: S_IDLE,
     layer: S_IDLE,
-    // If there's no marker layer (i.e. there's no min zoom extent), set it to
-    // succeeded to be ignored.
-    'zoom-markers': !minZoom ? S_SUCCEEDED : S_IDLE
+    'zoom-markers': S_IDLE
```
danielfdsilva, line 74: Given that the markers have nothing to load now, maybe it is a good idea to rename this status? Something like `'bbox-data'`?

danielfdsilva (changes requested):

app/scripts/components/common/mapbox/layers/utils.ts
hanbyul-here: my 💭 : Not changing the variable is a generally great rule of thumb (and it often feels more exciting? to write in a declarative way). However, considering iteration methods still have different performances, we might need to be flexible on the rule. As far as I checked last time, `foreach` still has better performances than declarative methods such as `reduce`?

app/scripts/components/common/related-content.tsx
```diff
@@ -105,7 +105,10 @@ function formatContents(relatedData: Array<RelatedContentData>) {
         'Something went wrong. Check the related content frontmatter.'
       );
 
-    const { name, description, pubDate, media } = matchingContent;
+    const { name, description, pubDate, media } = matchingContent as Record<
+      string,
+      any
+    >;
```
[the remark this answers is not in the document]
nerik: Ah, very nice, missed that!

app/scripts/components/common/mapbox/layers/utils.ts
nerik: @danielfdsilva @hanbyul-here Thank you! I adopted the "happy middle ground approach". 
- using `reduce` over `forEach` does feel more elegant
- the first approach is just too difficult to read IMHO
  - thinking of me re-reading that in 6 months
  - we have 4 items, with the need for logical branching for half of them, so in reality you are abstracting away logic for 2 items. Worth it?
  - or put more simply: "simplifying" the code by writing more lines of code: probably not needed and/or premature 
- re performance: I believe where you really see a difference is between `for` loops and all other declarative approaches (ie `reduce`, `map`, `forEach`, etc), but even though it doesn't really matter in this order of magnitude of iterations

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -376,6 +339,11 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
           'admin-0-boundary-bg'
         );
         changeStatus?.({ status: S_SUCCEEDED, context: 'layer' });
+        const layerBounds = getMergedBBox(stacCollection);
+
+        if (checkFitBoundsFromLayer(layerBounds, mapInstance)) {
+          mapInstance.fitBounds(layerBounds, { padding: FIT_BOUNDS_PADDING });
+        }
```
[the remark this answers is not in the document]
nerik: Thanks, adressed <a href="https://github.com/NASA-IMPACT/delta-ui/pull/272/commits/7f10e8a67bb9389e167d777562e14830dfb1fb29#diff-859ada2fe95f22e3a018b9d8e1403db207fb23c99394e6164f1d524acfa22bc6R384-R394">here</a>

danielfdsilva (approved):

PR #275, opened 2022-10-12
Sticky header

### LINKED ISSUE #237: Strange behavior with sticky header on scroll
When scrolling the page shakes as the sticky header is being hidden. This seems to only happen in firefox and is a frustrating experience because the page feels like it doesn't want to scroll and is pushing back.

I think it has to do with the fact that the browser tries to compensate the negative margin on the sticky element by repositioning the rest of the page, but it is not doing it well.

Video to demonstrate:

https://user-images.githubusercontent.com/1090606/187896169-cdd266c7-dd57-42d9-9cad-f143662f6731.mp4

danielfdsilva:
**tldr;** This PR re-implements the sticky header functionality to allow the implementation of the secondary sticky header needed in the analysis page and in the process fixes #237 since the header doesn't use margin anymore.

The analysis header contains the controls for the analysis page and will need to stick to the top when the user scrolls and will also need to adapt its content when it is stuck (not done in the context of this PR).
Having this header stick to the top can be achieved with CSS and position sticky but it is not enough because of the already existent sticky page header. To complicate things further, the page header gets hidden when the user scrolls.
The solution is to change how the analysis page header behaves depending on the status of the global header.

In general terms there a `requestAnimationFrame` that checks the status and sizes of the header and stores those values in the context (Layout props context), which can then be accessed via a hook by the relevant elements.

@nerik @hanbyul-here Let me know your thoughts on the approach. It is a bit of a complex problem to explain and handle.


nerik (commented):
The behaviour looks really great. I need more time and probably a lot more context to wrap my head around the logic, so I might not be the best reviewer out there.

I do however have a general comment on performance - not necessarily to be addressed before merging, maybe to be kept in mind for later:
This type of situation really begs for a solution based on <a href="https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API">Intersection Observer</a>. While `requestAnimationFrame` has benefits over a `setInterval` approach ie it avoids spamming background tabs, syncs with the browser repaint etc (I found <a href="http://www.javascriptkit.com/javatutors/requestanimationframe.shtml">this refresher</a> quite useful), a rAF callback it will still run **on every frame**, so typically every 16ms. As far as I understand the <a href="https://github.com/NASA-IMPACT/delta-ui/pull/275/files#diff-ad9ae6d0b2d0f8d3629d8b89a06449c0a6f83f66b9652da6ea418bbfd3fb6451R24">logic</a> on the callback is potentially not cheap (I think it might trigger a complete reflow but I'm not sure), potentially noticeable on mobile.

An IntersectionObserver guarantees that whatever you run on the callback:
- only runs when the user is scrolling
- only runs when the relevant DOM elements are in the viewport, which in this case is beyond the fold, more or less

Sorry for the blabla, let me know what you think.

app/scripts/utils/use-sliding-sticky-header.ts
```diff
@@ -0,0 +1,88 @@
...
+  const [wrapperHeight, setWrapperHeight] = useState(wrapperHeightQueried);
+
+  useEffect(() => {
+    let reqId;
+    let lastTs = 0;
+    let prevY = window.pageYOffset;
+    let scrollUpDelta = 0;
+
+    function tick(ts) {
+      if (!lastTs || ts - lastTs >= 10) {
```
nerik, line 31: Just curious, why that value of `10` ? As far as I understand rAF generally runs at 60fps so one frame every 16ms, have you consistently seen values below 10?

danielfdsilva: ah yes, a leftover :p

I had 100, but was not fast enough, so I removed a 0 to try and then forgot to remove the if entirely.

danielfdsilva: good catch :)

danielfdsilva:
@nerik You're correct that the code on the callback causes a reflow 🤔 
However I don't think we can do this with an InteractionObserver. We need to know the scroll value, to understand by how much the user has scrolled and what's the page position. (If you have suggestions to try let me know)

But now that I look at this, I think we should use the old onScroll event, because that's exactly what we want to respond to: When the user scrolls we want to know what happens, not on every frame.
I'll re-implement using this, but would like to know your thoughts.

danielfdsilva:
@nerik Valid points. To mitigate that I like the solution that @hanbyul-here used originally:

```js
useEffect(() => {
  let ticking = false;

  function tick() {
    // Code.
  
    ticking = false;
  }

  function onScroll() {
    if (!ticking) {
      // instead of setting a specific number of ms to wait (throttling),
      // pass it to the browser to be processed on the next frame, whenever that may be.
      window.requestAnimationFrame(tick);
      ticking = true;
    }
  }
  
  window.addEventListener('scroll', onScroll);

  return () => {
    window.removeEventListener('scroll', onScroll);
  };
}, []);
```

It is not perfect but an improvement. I'd still very much like to hear how you'd go about using IntersectionObserver. For the threshold parameter you'd likely need an element that's the height of the viewport, to be able to get updates, or am I completely missing the point?

app/scripts/utils/use-sliding-sticky-header.ts
```diff
@@ -86,7 +86,7 @@ export function useSlidingStickyHeader() {
       ticking = false;
     }
 
-    function onScroll() {
+    function onEvent() {
```
nerik, line 89:
```suggestion 
    function onViewportPositionChange() {
```

PR #277, opened 2022-10-18
Analysis header

danielfdsilva:
Includes the structural work done on the analysis page header.
- Refactor of the sticky header
- Behavior when the header gets stuck

The content is still dummy but we can address the remaining changes in other PRs


nerik (approved):
👌

PR #278, opened 2022-10-18
Fixed crash on dataset exploration launch

nerik:
🤦‍♂️

### DIFF
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -385,7 +385,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
   // FitBounds when needed
   //
   useEffect(() => {
-    if (!stacCollection || !stacCollection.length) return;
+    if (!stacCollection.length) return;
     const layerBounds = getMergedBBox(stacCollection);
 
     if (checkFitBoundsFromLayer(layerBounds, mapInstance)) {
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -385,6 +385,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
   // FitBounds when needed
   //
   useEffect(() => {
+    if (!stacCollection || !stacCollection.length) return;
     const layerBounds = getMergedBBox(stacCollection);
 
     if (checkFitBoundsFromLayer(layerBounds, mapInstance)) {


app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -385,6 +385,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
   // FitBounds when needed
   //
   useEffect(() => {
+    if (!stacCollection || !stacCollection.length) return;
```
danielfdsilva, line 388:
```suggestion
    if (!stacCollection?.length) return;
```

hanbyul-here: I am usually a fan of `?` but with the combination of `!` it might be hard to read?

nerik: Good catch, thank you. 

Somehow I keep forgetting that optional chaining exists in TS. So I suggest adding <a href="https://typescript-eslint.io/rules/prefer-optional-chain/">this eslint rule</a>.

Which results in
Screenshot 2022-10-19 at 17 17 01

I suggest we do as the eslint recommends, which is using `plugin:@typescript-eslint/strict"` instead of `"plugin:@typescript-eslint/recommended"`

I tried and it would yield a very manageable number of errors (warnings actually) which I suggest I tackle on another PR

Screenshot 2022-10-19 at 17 22 13

danielfdsilva: @nerik YES to the rule! :) Please go ahead with the PR

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -385,6 +385,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
   // FitBounds when needed
   //
   useEffect(() => {
+    if (!stacCollection.length) return;
```
danielfdsilva, line 388: @nerik I think now you removed the optional chaining altogether 🙃

danielfdsilva (changes requested):

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
nerik: Yes, but the reason is:
Screenshot 2022-10-19 at 20 26 55
Screenshot 2022-10-19 at 20 28 36


...another rule triggered by `@typescript-eslint/strict` (`stacCollection` is a state initialized with an empty array)

danielfdsilva: ooooh I see. Never mind then :)

PR #282, opened 2022-10-19
[DO NOT MERGE] Analysis header IntersectionObs POC

nerik:
Okay so I got a little bit lost trying to wrap my head around how context is structured in the app and how DOM elements are placed, so I kinda gave up but I still wanted to demo this: track page scrolling by adding an IntersectionObserver to the `main` section of the document, using multiple `threshold`s to update frequently (pros and cons of this approach discussed <a href="https://github.com/NASA-IMPACT/delta-ui/pull/275#pullrequestreview-1141184838">here</a>)

Let me know if you can make any use of this @danielfdsilva

### DIFF
--- a/app/scripts/components/analysis/define/index.tsx
+++ b/app/scripts/components/analysis/define/index.tsx
@@ -105,15 +105,11 @@ export default function Analysis() {
   const thematic = useThematicArea();
   if (!thematic) throw resourceNotFound();
 
-  const [scrollRatio, setScrollRatio] = useState<number>(0);
-
   useEffect(() => {
     // Generate enough thresholds to get regular updates from the observer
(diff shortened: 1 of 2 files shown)


app/scripts/components/analysis/define/index.tsx
```diff
@@ -105,6 +105,22 @@ export default function Analysis() {
...
+    }, {
+      threshold
+    });
+
+    // TODO use a ref on PageMainContent
+    if (document.getElementsByTagName('main')[0]) {
+      obs.observe(document.getElementsByTagName('main')[0]);
+    }
+
+  }, []);
```
danielfdsilva, line 122: It is a bit confusing that this can't go inside `app/scripts/utils/use-sliding-sticky-header.ts`, since when that mounts the `main` is not yet available.

PR #283, opened 2022-10-19
Fix Eslint configuration

nerik:
This is a review of the project's eslint config:
- Renamed the package.json script from `lint:js` to `lint:scripts` and actually check ts files as well 
- Configured eslint so that it uses a distinct parser for js and ts files as it should be on js/ts mixed codebases (explicitly set `@babel/eslint-parser` for JS files, before that it was using `@typescript-eslint/parser` to lint JS files) 
- I did not activate eslint typescript in _strict_ mode, as it generates 180 warnings. I did not realize that when writing <a href="https://github.com/NASA-IMPACT/delta-ui/pull/278#discussion_r999614894">this</a> because running eslint from yarn did not actually check TS files.
  - Ways to tackle this is either we leave it as is for now, or try to do it progressively, or wait to have a relatively low number of open PRs and fix all affected files in one go
- Fixed any errors left in either JS or TS files, keeping `react-hooks/exhaustive-deps` as discussed

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,15 +1,13 @@
 {
-  "parser": "@typescript-eslint/parser", // Specifies the ESLint parser
+  "parser": "@babel/eslint-parser",
   "plugins": [
     "react",
     "react-hooks",
-    "@typescript-eslint",
     "inclusive-language",
     "jest"
   ],
   "extends": [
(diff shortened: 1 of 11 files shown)


hanbyul-here (approved):
Thanks for the fix!

PR #284, opened 2022-10-20
Wire up analysis refine with stac endpoint

nerik:
~- [x] Sync start/end dates with state and URL~
~- [ ] Use URL as a single source of truth / remove analysis state~
~- [ ] Sync AOI URL and UI~
- [ ] Use Google-polyline for encoding of AoI in URL?
- [x] Query Stac search endpoint with UI parameters
- [x] Refine Stac results with local datasets
~- [ ] Sync URL with selected datasets~
- [x] UI checks for start/end dates
- [x]  switch between infobox and datasets lists 
- [x] Update tooltip when datasets are selected 
- [x] allow saving to navigate to results  
- [x] Make date selectors active
- [x] reset state on AoI delete


app/scripts/components/analysis/results/use-analysis-params.ts
```diff
@@ -9,40 +9,41 @@ import { polygonUrlDecode, polygonUrlEncode } from '$utils/polygon-url';
...
+
+type AnyAnalysisParamsKey = keyof AnalysisParams;
+type AnyAnalysisParamsType = Date | DatasetLayer[] | Feature<MultiPolygon>;
+
+const initialState: AnalysisParamsNull = {
+  start: undefined,
+  end: undefined,
+  datasetsLayers: undefined,
+  aoi: undefined,
+  errors: []
```
danielfdsilva, line 32: There are some conflicting types for `errors`.
I'd say we do `undefiend` when there are no errors, and `[]` when there's something. Let's avoid, empty error arrays. thoughts?

app/scripts/components/analysis/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+      {
+        op: 'in',
+        args: [{ property: 'collection' }, collections]
+      }
+    ]
+  };
+  return filterPayload;
+}
+
+export function multiPolygonToPolygon(
```
danielfdsilva, line 47: This function's name seems a bit misleading. You're effectively removing part of the multipolygon to return a polygon, not doing a direct conversion.

app/scripts/components/analysis/define/index.tsx
```diff
@@ -101,21 +120,170 @@ export default function Analysis() {
...
+    const toMultiPolygon: Feature<MultiPolygon> = {
+      type: 'Feature',
+      properties: { ...aoiDrawState.feature.properties },
+      geometry: {
+        type: 'MultiPolygon',
+        coordinates: [aoiDrawState.feature.geometry.coordinates]
+      }
+    };
+    setAnalysisParam('aoi', toMultiPolygon);
+  }, [aoiDrawState]);
```
danielfdsilva, line 263: Should include `setAnalysisParam`

app/scripts/components/analysis/define/index.tsx
```diff
@@ -1,6 +1,13 @@
-import React from 'react';
+import React, {
```
danielfdsilva, line 1: I'd say we can split each block (aoi, date, datasets) into a different file to have this one not be do long

app/scripts/components/analysis/define/index.tsx
```diff
@@ -101,21 +120,170 @@ export default function Analysis() {
...
+          (l) => l.id === id
+        );
+        if (newDatasetLayer) {
+          newDatasetsLayers.push(newDatasetLayer);
+        }
+      } else {
+        const removeAt = newDatasetsLayers.findIndex((l) => l.id === id);
+        newDatasetsLayers.splice(removeAt, 1);
+      }
+      setAnalysisParam('datasetsLayers', newDatasetsLayers);
```
danielfdsilva, line 186: It would be better not to use methods that modify the array directly.
`concat` to add, `filter` to remove

danielfdsilva (changes requested):
As you mentioned the url state is not working properly, but you're also updating the analysis params, which will be a problem if you want to get the initial parameters back for the "cancel" option. This was also the reason why `useAnalysisParams` did not export a setter.

app/scripts/components/analysis/define/aoi-selector.tsx
```diff
@@ -59,17 +61,9 @@ export default function AoiSelector(props: AoiSelectorProps) {
   // Despite the query parameters support for multiple features on the aoi, the
   // AOI drawing tool only supports one.
   // Keeping just the first one.
-  const polygon: Feature<Polygon> | null = useMemo(() => {
+  const polygon: Feature<Polygon> | null = useDeepCompareMemo(() => {
```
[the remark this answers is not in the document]
nerik: Good catch, will fix

app/scripts/components/analysis/results/use-analysis-params.ts
```diff
@@ -128,15 +128,25 @@ export function useAnalysisParams(): AnalysisParams {
...
 export function analysisParams2QueryString(
-  params: Omit<AnalysisParams, 'errors'>
+  params: Omit<AnalysisParams | AnalysisParamsNull, 'errors'>
 ) {
   const urlParams = qs.stringify({
-    start: params.date.start ? userTzDate2utcString(params.date.start) : undefined,
-    end: params.date.end ? userTzDate2utcString(params.date.end) : undefined,
+    start: params.start ? userTzDate2utcString(params.start) : undefined,
+    end: params.end ? userTzDate2utcString(params.end) : undefined,
     datasets: params.datasetsLayers
```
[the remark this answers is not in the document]
nerik: I think so, yes.

app/scripts/components/analysis/define/index.tsx
nerik: Since `newDatasetsLayers` is a shallow copy of `datasetsLayers` there's no risk mutating it. But you're right, it's good practice anyways, will fix.

app/scripts/components/analysis/results/use-analysis-params.ts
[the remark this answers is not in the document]
danielfdsilva: I defer to you though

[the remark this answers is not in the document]
nerik: I [settled](https://github.com/NASA-IMPACT/delta-ui/pull/284/commits/76fd6ae15c7ae67a297c8d88c2946f8385e29fe1) on using `datasetsLayers` in the url, allright with you?

app/scripts/components/analysis/utils.ts
nerik: I agree that's not 100% accurate, but honestly I'm struggling to find a decent name for that 😄 (`firstMultiPolygonPolygonToPolygon` ? `trimMultiPolygonToPolygon` ?)
I'd argue that there's no ambiguity since there's no way to _convert_ a Polygon to a MultiPolygon, so one could infer that it's taking the first MultiPolygon's polygon?

app/scripts/components/analysis/define/index.tsx
nerik: https://github.com/NASA-IMPACT/delta-ui/pull/284/commits/1eb5f54785a09fe6061c1463a6e98f83e013d33a

app/scripts/components/analysis/results/use-analysis-params.ts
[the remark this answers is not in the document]
danielfdsilva: perfect!

app/scripts/components/analysis/utils.ts
danielfdsilva: fair point!

danielfdsilva: We don't need to do that, but I imagine that this function could exist with a second parameter to indicate the index of the multipolygon's polygon to keep. :D

danielfdsilva (changes requested):
I noticed that some datasets are not being removed from the "active list" if not actively unselected. Steps:

- Draw aoi over Netherlands and dates from 2015 to today
- Select no2 and ightlights
- go to results
- click refine
- change aoi to portugal
- from the available datasets unselect no2 and select no2-diff
- go to results

You'll notice that nightlights is still present. Since it is not available over portugal, the user did not actively unselect it and it was not removed from the list.

app/scripts/components/analysis/utils.ts
nerik: Yes, but [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) 😁

app/scripts/components/analysis/define/page-hero-actions.tsx
```diff
@@ -0,0 +1,99 @@
...
+          variation='achromic-outline'
+        >
+          <CollecticonXmarkSmall /> Cancel
+        </Button>
+      )}
+      {showTip ? (
+        <Tip
+          visible
+          placement='bottom-end'
+          content='To get results, define an area, pick a date and select datasets.'
```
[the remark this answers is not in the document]
nerik: https://github.com/NASA-IMPACT/delta-ui/pull/284/commits/a1599fe501a873d22d53f52b83e19a71e73b81af

nerik:
@danielfdsilva Sharp eyes 👀 
Fixed here: https://github.com/NASA-IMPACT/delta-ui/pull/284/commits/284d4bf837ef358e36058bf8addd898d498fb63f

app/scripts/components/analysis/define/use-stac-search.ts
```diff
@@ -0,0 +1,99 @@
...
+  const [stacSearchStatus, setStacSearchStatus] =
+  useState<ActionStatus>(S_IDLE);
+  useEffect(() => {
+    if (!readyToLoadDatasets || !allAvailableDatasetsLayers) return;
+
+    const load = async () => {
+      setStacSearchStatus(S_LOADING);
+      try {
+        // if (controller.current) controller.current.abort();
+        controller.current = new AbortController();
```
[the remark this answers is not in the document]
nerik: Nope. Just commented and forgot to uncomment, my bad.

danielfdsilva (approved):
All good! :shipit:

nerik:
🥳

PR #291, opened 2022-10-27
Add chart to analysis page

danielfdsilva:
Includes loading and error messages as well.

Here's a url which shows a bit of everything:

```
localhost:9000/covid-19/analysis/results?start=2015-01-01T00:00:00.000Z&end=2022-01-01T00:00:00.000Z&datasetsLayers=no2-monthly|no2-monthly-diff|blue-tarp-planetscope|dev-fail&aoi=-9.60205,36.72127|-7.03125,36.87962|-6.85546,39.45316|-6.52587,40.93011|-5.55908,42.11452|-9.38232,42.22851|-8.89892,40.84706|-9.93164,38.85682|-9.47021,38.08268|-8.96484,38.09998
```


hanbyul-here:
wah, the page looks great. I haven't looked at the code yet, but since URL is getting longer and longer, is this a good time to think about shortening it? using something like object-hash?

nerik:
@hanbyul-here @danielfdsilva Maybe have a look at https://www.npmjs.com/package/google-polyline ?

app/scripts/components/analysis/results/chart-card.tsx
```diff
@@ -0,0 +1,90 @@
...
+              ]}
+              xKey='date'
+              dates={data.timeseries.map((e) =>
+                dateFormatter(new Date(e.date), '%Y/%m')
+              )}
+              dateFormat='%Y/%m'
+              altTitle='alt title'
+              altDesc='alt desc'
+              xAxisLabel='x axis label'
+              yAxisLabel='y axis label'
```
nerik, line 81: Are we able to address that now? (I think I recall we don't have any anything about units)

app/scripts/components/analysis/results/index.tsx
```diff
@@ -159,7 +145,7 @@ export default function AnalysisResults() {
           </Button>
         )}
       />
-      <Fold>
+      <Fold style={{ height: '200vh' }}>
```
nerik, line 148: I'm pretty new to styled components so asking naively: what's the rule on using inline style along styled components? On a case-by-case basis?

app/scripts/components/common/loading-skeleton.tsx
```diff
@@ -128,6 +130,30 @@ const ContentLoadingSelf = styled.div`
...
+
+  &::before {
+    position: absolute;
+    inset: 0;
+    display: flex;
+    content: '';
+    z-index: -1;
+    padding: ${variableGlsp()};
+    background: ${themeVal('color.base-100')};
+    animation: ${pulse} 0.8s ease 0s infinite alternate;
```
nerik, line 153: 💅 
Can we use the occasion to store the whole string in a const? (it's repeated 4 times)

nerik (approved):
Looking great 💪 
Just a bunch of nitpicks and comments.

nerik:
Screenshot 2022-10-31 at 16 40 38

General comment on the design: it's very crowded around the graph axis(es), the y axis label is cut off, and title, chart and brush are not aligned horizontally. Overall there's probably a bit of whitespace that can be used to widen the graph horizontally, and the whole block could be bigger vertically to accommodate for everything.

app/scripts/components/analysis/results/index.tsx
danielfdsilva: Ups. This was not supposed to be here.
Answering your question: no inline styles

app/scripts/components/analysis/results/chart-card.tsx
danielfdsilva: Added a more generic description for now

PR #292, opened 2022-10-27
Image export

danielfdsilva:
The image export feature was tied to a button inside the chart, but the download button on the cards is in the header.

Given that the image export requires a chart reference, I ended up creating a imperative handle attached to the analysis chart that allows us to have extra programmatic methods.

I put this on a separate PR to be easier to review. There seem to be a lot of changes but it's mostly typing and indentation.
Let me know your thoughts.


app/scripts/components/common/chart/analysis/index.tsx
```diff
@@ -18,45 +20,68 @@ interface AnalysisChartProps extends CommonLineChartProps {
...
+        <Chart
+          {...props}
+          ref={chartRef}
+          syncId={syncId}
+          chartData={chartData}
+          renderLegend={false}
+          renderBrush={true}
+          colors={lineColors}
+        />
+        {/* <ExportImage
```
nerik, line 77: Cleanup?

app/scripts/components/common/chart/analysis/index.tsx
```diff
@@ -18,45 +20,68 @@ interface AnalysisChartProps extends CommonLineChartProps {
...
+      });
+    }, [timeSeriesData, dates, uniqueKeys, dateFormat, xKey]);
+
+    const lineColors = useMemo(() => {
+      return getColors({ steps: uniqueKeys.length, colorScheme: 'viridis' });
+    }, [uniqueKeys]);
+
+    useImperativeHandle(ref, () => ({
+      instance: chartRef.current,
+      saveAsImage: async () => {
```
nerik, line 52: 👍

app/scripts/components/analysis/results/chart-card.tsx
```diff
@@ -17,20 +21,60 @@ import {
...
+      .current.state as any;
+    // The indexes expect the data to be ascending, so we have to reverse the
+    // data.
+    const data = reverse(chartData.data.timeseries);
+    const dFormat = 'yyyy-MM-dd';
+    const startDate = format(new Date(data[dataStartIndex].date), dFormat);
+    const endDate = format(new Date(data[dataEndIndex].date), dFormat);
+
+    const filename = `chart.${id}.${startDate}-${endDate}`;
+    chartRef.current?.saveAsImage(filename);
```
danielfdsilva, line 76: @nerik To achieve the date on the filename I had to do some hacky magic.
Because the dates that get exported depend on what the brush is set to I needed the brush indexes. Instead of reworking the brush state in the [Chart](https://github.com/NASA-IMPACT/delta-ui/blob/12c4c612ac444753dbcbc2ea474a370a3a3d7383/app/scripts/components/common/chart/index.tsx#L109-L117) and having to pass it all up the chain, I accessed the state of the component directly.
Not very clean, but having to pass the brush index up the chain was not super clean either. Thoughts?

app/scripts/utils/utils.ts
```diff
@@ -44,3 +48,31 @@ export function isEqualObj(a, b) {
...
+      }
+    };
+  })`
+    ${({ visuallyDisabled: vd }) =>
+      vd &&
+      css`
+        ${visuallyDisabled()}
+      `}
+  `;
+}
```
danielfdsilva, line 78: @nerik To show a tooltip over a disabled button, it cannot be disabled, otherwise the pointer events are removed and the tooltip is not triggered. Visually disabling the button has the same visual result, but the click event still fires. This function allows us to compose components, disabling the click when visually disabled.

Perhaps this should also add the `aria-disabled` role, even though the element can still be interacted with to show a tooltip. @ricardoduplos thoughts here?

nerik: @danielfdsilva I had the same issue and used a [very mediocre hack](https://github.com/NASA-IMPACT/delta-ui/pull/284/commits/e269b4e67694873fe13e4118c2e15ccffd77e6e1#diff-111077f0776e65e2cf1a431485f118e091b99e82eb240ab5ad118c557de28323R18-R109).

Isn't there a simpler fix by just using a wrapper element? [Like this](https://codesandbox.io/s/withered-grass-yzhcry?file=/src/App.js)? Not sure it works with tippy

hanbyul-here:
Charts on other pages seem to be broken: https://deploy-preview-292--veda-ui.netlify.app/sandbox/mdx-chart

I am not exactly sure why the current change is introducing this bug, but it looks like there are 2 problems. 1. The data can be empty and `getNumForChart` is not handling the case when x is undefined.  2. It seems that there is a timing at the chart is rendered but data is not loaded yet, making y-axis range from -Infinity to Infinity. Both can be handled by adding checks like 
```
  if (x === undefined) return 'N/A';
  else if (!isFinite(x)) return;
  ``` 
to `getNumForChart` function, but maybe `formatAsScientificNotation` should handle the case when the number is infinite?

app/scripts/utils/utils.ts
danielfdsilva: I had not noticed that hack. 😅
Click events should only be used on interactive elements, like buttons, links, and the like. In this way the element is still focusable with keyboard navigation and can still be used, it just doesn't do anything. It is also semantically correct.

A couple of articles on the topic:
- [Semantic HTML: The Unbearable Rightness of Being](https://ux.shopify.com/semantic-html-the-unbearable-rightness-of-being-9b3c493e1791)
- [How (Not) to Build a Button](https://benmyers.dev/blog/clickable-divs/)

Full disclosure: I did not fully read the articles, but checked that they have the important parts

hanbyul-here: regarding `aria` roles: Maybe tooltip can have `aria-live` role so it can notify users when it shows up?

danielfdsilva:
@hanbyul-here This bug seems to have been introduced with #291. Will handle it there.

danielfdsilva:
After merging the other PR, had to merge into this one to solve the conflicts. Rebasing was proving too complicated

app/scripts/components/analysis/results/chart-card.tsx
nerik: @danielfdsilva `const [brushStartIndex, setBrushStartIndex] = useState(0);` needs to move up the component tree and should live here in chart-card. Yes, it means passing down both state prop and setter down to the chart component, which is not really an issue IMHO, but if we worry about that we have a lot of tools at our disposal like context, etc.
But it is bad practice to break encapsulation and access a component internal state from outside. Let's save us some future headaches 😉
[Don't overuse refs](https://reactjs.org/docs/refs-and-the-dom.html#dont-overuse-refs)

app/scripts/utils/utils.ts
nerik: @danielfdsilva You are right. Will fix on an upcoming PR.

app/scripts/components/analysis/results/chart-card.tsx
danielfdsilva: Instead of moving the whole state setting up the change, opted for a read only change event that doesn't break the already existing encapsulation.

nerik (approved):
👍 🚀

PR #294, opened 2022-10-28
Chart legend and colors

danielfdsilva:
Chart and legend now share colors from the theme. 
It also includes metric selection functionality.

### DIFF
--- a/app/scripts/components/analysis/results/analysis-head-actions.tsx
+++ b/app/scripts/components/analysis/results/analysis-head-actions.tsx
@@ -1,9 +1,8 @@
 import React from 'react';
 import styled, { useTheme } from 'styled-components';
+import { glsp, themeVal } from '@devseed-ui/theme-provider';
 import {
   Dropdown,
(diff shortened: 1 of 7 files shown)


app/scripts/components/analysis/results/analysis-head-actions.tsx
```diff
@@ -0,0 +1,160 @@
...
+                <LegendSwatch disabled={!active}>
+                  <svg height='8' width='8'>
+                    <title>{theme.color[metric.themeColor]}</title>
+                    <circle
+                      cx='4'
+                      cy='4'
+                      r='4'
+                      fill={theme.color[metric.themeColor]}
+                    />
+                  </svg>
```
nerik, line 120: Just curious, why an SVG element rather than a div with `border-radius: 4` ?

nerik (approved):
Looking good! 🕶️

ricardoduplos (approved):

PR #295, opened 2022-10-28
Concurrent requests on analysis results

danielfdsilva:
The concurrency manager of the analysis results requests was not properly implemented. Not only it was setup to handle concurrency on a datasetLayer level instead of globally, and the requests would not stop if the user clicked "Refine" while the charts were being loaded.

The solution in to clear the concurrency queue if an abort is triggered.


app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -87,16 +87,26 @@ export function requestStacDatasetsTimeseries({
...
   layers: DatasetLayer[];
   queryClient: QueryClient;
+  signal: AbortSignal;
 }) {
   const eventEmitter = EventEmitter();
 
+  const concurrencyManager = ConcurrencyManager();
+
+  // On abort clear the queue.
+  signal?.addEventListener('abort', () => {
```
nerik, line 105: Shouldn't we call `removeListener` when aborting?

nerik (approved):
Thanks for the fix!

app/scripts/components/analysis/results/timeseries-data.ts
danielfdsilva: Good catch! Setting up the event as a `once` thing.

PR #296, opened 2022-10-31
More eslint rules

nerik:
- Enforces the `import/order` rule and fixes file generating errors
- Enables TypeScript strict mode - for now only triggers warnings, let's fix things progressively until we reach a more manageable number of issues (~200 for now)
- Enables a [rule to prevent mutating objects](https://github.com/jfmengels/eslint-plugin-fp/blob/master/docs/rules/no-mutation.md)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -57,7 +57,7 @@
     "no-dupe-else-if": 0,
     "no-setter-return": 0,
     "prefer-promise-reject-errors": 0,
-    "import/order": 1,
+    "import/order": 2,
     "react/button-has-type": 2,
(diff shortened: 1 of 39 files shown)


danielfdsilva:
@nerik This is great! Would you rather fix some more warning, or merge and fix as we go?

nerik:
@danielfdsilva I think the best for now is to merge as is, brace for some light rebases due to applying `import/order`, then wait a few weeks to fix most warnings, then move from warnings to errors.

hanbyul-here (approved):
I love that we are making the lint rule more robust! Thanks!!!!

PR #302, opened 2022-11-03
Avoid useEffectPrevious on checking date compare

danielfdsilva:
The main problem here was that `isComparing && availableActiveLayerCompareDates?.length` were together. The fact that `availableActiveLayerCompareDates?.length` evaluates to `true` does not mean that `setSelectedCompareDatetime(null);` should happen.

Separating them solves the problem without the need to resort to any previous value.

Your solution works perfectly. This seems simple to reason about, but up to you which one to keep. :)

### DIFF
--- a/app/scripts/components/datasets/s-explore/index.tsx
+++ b/app/scripts/components/datasets/s-explore/index.tsx
@@ -398,20 +398,28 @@ function DatasetsExplore() {
     }
   }, [activeLayer, setSelectedCompareDatetime]);
 
+  const initialDatetime = activeLayer?.baseLayer.data?.initialDatetime;
+
   // Deselect compare dates when compare toggle changes.
-  useEffectPrevious(([prevIsComparing]) => {
-    // Only trigger this when user is toggling comparison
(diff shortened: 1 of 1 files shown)


app/scripts/components/datasets/s-explore/index.tsx
```diff
@@ -398,20 +398,28 @@ function DatasetsExplore() {
...
-    if (isComparing && availableActiveLayerCompareDates?.length) {
-      setSelectedCompareDatetime(selectedCompareDatetime ?? availableActiveLayerCompareDates.last);
+  useEffect(() => {
+    if (isComparing) {
+      if (
+        !selectedCompareDatetime &&
+        availableActiveLayerCompareDates?.length
+      ) {
+        setSelectedCompareDatetime(
+          getInitialDate(availableActiveLayerCompareDates, initialDatetime)
```
danielfdsilva, line 411: There's another minor fix I added here. The initial date should be dictated by the config in the `mdx` and not be `.last`

nerik (approved):
Thanks!

Issue #305, opened 2022-11-04
Cannot find module error


danielfdsilva:
When accessing the analysis page at https://veda-ui.netlify.app/covid-19/analysis, A module not found error is presented:

<img width="1000" alt="image" src="https://user-images.githubusercontent.com/1090606/200044719-a68343f7-ec17-4e8f-a78a-4b9e8c7db84a.png">

I did some debugging around this issue and I have a general idea about what is happening, but couldn't figure out how to solve this.

The module in question refers to the file [`app/scripts/utils/format.ts`](https://github.com/NASA-IMPACT/delta-ui/blob/75095baf3b82f42d559d9d096c2777ea38b13ba4/app/scripts/utils/format.ts) which parcel is internally naming `aV6JG`. There's nothing special about this file as far as I can tell, it just happened to be this one.

There are several files using functions exported by this file, like:
- app/scripts/components/home/index.tsx
- app/scripts/components/common/mapbox/map.tsx
- app/scripts/components/common/mapbox/layer-legend.tsx
- app/scripts/components/common/chart/utils.ts
- app/scripts/components/common/blocks/scrollytelling/chapter.tsx
- app/scripts/components/common/blocks/block-map.tsx
- app/scripts/components/common/aoi/utils.ts

To avoid duplicating the code multiple times, parcel will just package it once, name it something (`aV6JG`) and then reference it every time.

It seems that this behavior is clashing with code splitting. Each of our route components (Home, Analysis, About, etc) is being separated into a different file, and only loaded when needed. (Done automatically when using [`import('path')`](https://parceljs.org/features/code-splitting/) in [`app/scripts/main.js`](https://github.com/NASA-IMPACT/delta-ui/blob/75095baf3b82f42d559d9d096c2777ea38b13ba4/app/scripts/main.js))

The problem here is that parcel is placing the code for our `utils/format`, inside Home (Home uses the `zeroPad` function from the file). When we load the analysis page, the file for Home is not loaded, hence the "module not found". This also explains why if you start at the [site root](https://veda-ui.netlify.app), select a thematic area and then Analysis it works. The Home file was loaded and so the module is registered.

The next thing I'm going to try is to update parcel to the most recent version and see if it does anything.
cc @nerik @hanbyul-here (@olafveerman if you're curious)

danielfdsilva:
[...]

PR #306, opened 2022-11-07
Update parcel and switch bundler

### LINKED ISSUE #259: Module not found and MDX content

danielfdsilva:
Updating to the experimental bundler seems to solve ﻿﻿#305 and also seems to have fixed #259 🤩


danielfdsilva:
Accessing the analysis page directly via https://deploy-preview-306--veda-ui.netlify.app/covid-19/analysis works, meaning that the problem is fixed. Comparing with https://veda-ui.netlify.app/covid-19/analysis which fails

nerik (approved):
👏

Issue #307, opened 2022-11-07
Analysis results page does not show data.


danielfdsilva:
Even though the analysis refine page allows the datasets to be selected, the analysis results does not show anything.

![image](https://user-images.githubusercontent.com/1090606/200308852-195629fc-0c88-4ca5-a7a6-fab0245668fb.png)

----

## Explanation

This is happening because the analysis definition and analysis results page are making requests using a different filter format:

<table>
<thead>
  <tr>
    <th>Definition page</th>
    <th>Results page</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>

```js
{
    op: 'and',
    args: [
      {
        op: 't_intersects',
        args: [
          { property: 'datetime' },
          {
            interval: [
              '2022-05-07T00:00:00.000Z',
              '2022-08-05T00:00:00.000Z'
            ]
          }
        ]
      },
      // ... other params
    ]
  }
```

</td>
<td>

```js
{
          op: 'and',
          args: [
            {
              op: '>=',
              args: [
                {
                  property: 'datetime'
                },
                date.start
              ]
            },
            {
              op: '<=',
              args: [
                {
                  property: 'datetime'
                },
                date.end
              ]
            },
            // ... other params
          ]
        }
```

  </td>
  </tr>
</tbody>
</table>

Using the intersects filter, takes into account the start date and end date of a dataset, which in the NO2 case is a whole month. By using a greater/less than filter we only look at the datetime, which is set to be the first of the month

PR #308, opened 2022-11-07
Fixes some analysis related issues

danielfdsilva:
- Analysis results does not show data even though analysis definition allows dataset selection (Fix https://github.com/NASA-IMPACT/delta-ui/issues/307)
- Updating the dates does not update results https://github.com/NASA-IMPACT/delta-ui/issues/211
- Visual glitches with map controls
- Loading not shown when available datasets are being fetched
- AOI Draw internal sate not being cleared when url changes

cc @olafveerman

### DIFF
--- a/app/scripts/components/analysis/define/aoi-selector.tsx
+++ b/app/scripts/components/analysis/define/aoi-selector.tsx
@@ -71,6 +71,9 @@ export default function AoiSelector(props: AoiSelectorProps) {
       onAoiEvent('aoi.set-feature', { feature: polygon });
       const featureBbox = bbox(polygon) as [number, number, number, number];
       mapRef.current?.instance?.fitBounds(featureBbox, { padding: 32 });
+    } else {
+      onAoiEvent('aoi.clear');
+      mapRef.current?.instance?.flyTo({ zoom: 1, center: [0, 0] });
     }
   }, [onAoiEvent, polygon]);
 
--- a/app/scripts/components/analysis/define/index.tsx
+++ b/app/scripts/components/analysis/define/index.tsx
@@ -112,7 +112,7 @@ export default function Analysis() {
       if (!e.target.value || e.target.value === '') {
         setAnalysisParam('start', null);
         return;
-      } 
+      }
       setAnalysisParam('start', inputFormatToDate(e.target.value));
     },
     [setAnalysisParam]
@@ -120,7 +120,7 @@ export default function Analysis() {
 
   const onEndDateChange = useCallback(
     (e) => {
-      if (!e.target.value || e.target.value === '') { 
+      if (!e.target.value || e.target.value === '') {
         setAnalysisParam('end', null);
         return;
       }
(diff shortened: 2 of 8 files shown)


app/scripts/components/analysis/define/aoi-selector.tsx
```diff
@@ -71,6 +71,9 @@ export default function AoiSelector(props: AoiSelectorProps) {
       onAoiEvent('aoi.set-feature', { feature: polygon });
       const featureBbox = bbox(polygon) as [number, number, number, number];
       mapRef.current?.instance?.fitBounds(featureBbox, { padding: 32 });
+    } else {
+      onAoiEvent('aoi.clear');
+      mapRef.current?.instance?.flyTo({ zoom: 1, center: [0, 0] });
```
danielfdsilva, line 76: When refining an analysis and clicking the "Analysis" button the AOI is removed from the url but the map was not being cleared

app/scripts/components/analysis/define/index.tsx
```diff
@@ -185,12 +191,12 @@ export default function Analysis() {
...
     } else {
-      if (!selectableDatasetLayers.length) {
-        return 'No datasets available in that thematic area for the currently selected dates and area.';
-      } else if (stacSearchStatus === S_LOADING) {
+      if (stacSearchStatus === S_LOADING) {
         return 'Loading...';
       } else if (stacSearchStatus === S_FAILED) {
         return 'Error loading datasets.';
+      } else if (!selectableDatasetLayers.length) {
+        return 'No datasets available in that thematic area for the currently selected dates and area.';
```
danielfdsilva, line 199: The order of the if statements was causing the Loading message not to be displayed

app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -155,36 +159,32 @@ async function getDatasetAssets(
...
+                { property: 'datetime' },
                 {
-                  property: 'collection'
-                },
-                id
+                  interval: [
+                    date.start,
+                    date.end
+                  ]
+                }
```
danielfdsilva, line 187: See #307 for explanation.

app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -254,7 +254,7 @@ async function requestTimeseries({
 
   try {
     const layerInfoFromSTAC = await queryClient.fetchQuery(
-      [TIMESERIES_DATA_BASE_ID, 'dataset', id],
+      [TIMESERIES_DATA_BASE_ID, 'dataset', id, aoi, start, end],
```
danielfdsilva, line 257: Even though the dates were changing on the analysis page, the query key (used for caching) was not including the data, so the app was acting like nothing changed.

app/scripts/components/analysis/results/timeseries-data.ts
nerik: Too bad we are duplicating logic here. Is there a way we could reuse `analysis/utils/getFilterPayload` here (sending a single id as the `collections` arg).

Otherwise great use case for https://github.com/developmentseed/ds-business/issues/1310

nerik (approved):
Thanks for the fixes 👍

app/scripts/components/analysis/results/timeseries-data.ts
danielfdsilva: @nerik Indeed we can! done! :)

nerik: @danielfdsilva Perfect!

PR #315, opened 2022-11-09
Typescript

danielfdsilva:
I got started with moving some files to typescript.
I also added the styled-components types which resulted in having to add types to a lot of them.

There are still many files to move. This PR tackles most of the most straightforward ones.
I think I didn't change any significant logic, but a quick test is needed


danielfdsilva:
Pinging @Catalina-Moller @gmarichalgxc to also do a quick test on the preview url to ensure nothing broke

app/scripts/components/analysis/define/page-hero-actions.tsx
```diff
@@ -19,7 +18,7 @@ import { composeVisuallyDisabled } from '$utils/utils';
 const SaveButton = composeVisuallyDisabled(Button);
 
 interface PageHeroActionsProps {
-  size: string;
+  size: ButtonProps['size'];
```
nerik, line 21: I wasn't aware of this syntax ("indexed access types"). 
I feel though that we'd be better off with a distinct ButtonSize type.

app/scripts/components/analysis/results/analysis-head-actions.tsx
```diff
@@ -65,7 +62,7 @@ const MetricList = styled.ul`
   }
 `;
 
-const MetricSwitch = styled(FormSwitch)`
+const MetricSwitch = styled(FormSwitch)<{ metricThemeColor: string }>`
```
nerik, line 65: Below in line 74, we might be able to avoid casting to `any` - see https://medium.com/@KevinBGreene/fun-with-typescript-template-literal-types-95580e27e0e1

nerik (approved):
Looking good, thanks.
It seems a new TypeScript pass on the ui library might be useful as well.

Catalina-Moller:
> Pinging @Catalina-Moller @gmarichalgxc to also do a quick test on the preview url to ensure nothing broke

@danielfdsilva I did a quick test on the preview website and everything seems to work as expected

app/scripts/components/analysis/define/page-hero-actions.tsx
danielfdsilva: Indeed. The ui library definitely needs a typescript improvement. For the time being this is what's possible.

@nerik Maybe we can do some UI lib cleanup during flex week. I'd appreciate your input a lot :)

app/scripts/components/analysis/results/analysis-head-actions.tsx
danielfdsilva: Not sure it is in this case, and once again it has to do with not phenomenal types on the UI library.

The parameter for `themeVal(path)` must be a valid theme object path. Even though something like `color.infographicA` would be valid in this app's theme, the list of theme paths are manually [defined in the ui library](https://github.com/developmentseed/ui-library-seed/blob/5f00bf13e66c6afa7f46f22c46a76eaf9fa1e18a/packages/theme-provider/src/utils.d.ts#L8-L197).

I know that there are ways to [dynamically generate the paths](https://dev.to/bwca/create-object-property-string-path-generator-with-typescript-13e3) so could be something to explore.

danielfdsilva: Nevertheless, let me know if you have any suggestion on how to approach it.

PR #320, opened 2022-11-11
make idKey optional for mdx chart

hanbyul-here:
Close https://github.com/NASA-IMPACT/delta-ui/issues/316
- Let Chart component for mdx handle csv&json without idkey (I named the branch as two-column, but now that I think... the number  of columns doesn't really matter)
- Hide legend when there is not more than one line  on Chart
- Parcel's json loading doesn't really work for mdx path pattern. Added `@parcel/transformer-raw` to handle this case. The file should follow the naming pattern `raw*.json` to be handled by the transformer. (This info needs to go to documentation)


danielfdsilva (changes requested):
I'm wondering if we should remove the legend when there's only one data dimension. It may be that the combination of axis labels is not enough. 

Allowing the user to provide a value would be an option, but the behavior would be different if there are multiple dimensions.
@ricardoduplos You feel like the legend is needed even if there's only one dimension?

image

PR #322, opened 2022-11-11
Include image caption when exporting chart

danielfdsilva:


hanbyul-here (approved):
👍 
I wondered if it would make more sense to have a separate propa for title instead of reusing altTitle. I think it would be fine to use altTitle since this prop will be handled by devs

PR #328, opened 2022-11-15
Analysis presets

nerik:
Screenshot 2022-11-15 at 12 27 46
Screenshot 2022-11-15 at 13 10 00

Adds options to automatically select date ranges and regions in the analysis page.
Mostly I wanted to be able to debug the analysis page more quickly, but this should serve as a base for when we decide which presets we want to offer (for now  `'thisYear'   | 'last30Days'  | 'lastYear'   | 'last10Years'` for dates and `'world'` for regions) .

TODO for later:
- decide on presets
- figure out a way to use multipolygons (it will be needed for anything like 'USA' as a region)

### DIFF
--- a/app/scripts/components/analysis/define/aoi-selector.tsx
+++ b/app/scripts/components/analysis/define/aoi-selector.tsx
@@ -1,4 +1,10 @@
-import React, { useEffect, useMemo, useRef } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState
+} from 'react';
 import styled from 'styled-components';
 import { Feature, MultiPolygon, Polygon } from 'geojson';
 import bbox from '@turf/bbox';
@@ -22,6 +28,7 @@ import {
   CollecticonUpload2
 } from '@devseed-ui/collecticons';
 import { multiPolygonToPolygon } from '../utils';
+import { FeatureByRegionPreset, RegionPreset } from './constants';
 import {
   Fold,
   FoldHeader,
@@ -50,8 +57,11 @@ interface AoiSelectorProps {
   onAoiEvent: AoiChangeListenerOverload;
 }
 
-export default function AoiSelector(props: AoiSelectorProps) {
-  const { onAoiEvent, qsFeature, aoiDrawState } = props;
+export default function AoiSelector({
+  onAoiEvent,
+  qsFeature,
+  aoiDrawState
+}: AoiSelectorProps) {
   const { selected, drawing, feature } = aoiDrawState;
   const mapRef = useRef<MapboxMapRef>(null);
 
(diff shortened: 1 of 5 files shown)


app/scripts/components/common/aoi/types.d.ts
```diff
@@ -34,11 +34,11 @@ export type AoiChangeListener = (
...
+export interface AoiChangeListenerOverload {
+  (action: 'aoi.draw-click' | 'aoi.clear', payload?: never): void;
+  (
+    action: 'aoi.draw-finish' | 'aoi.set-feature' | 'aoi.update',
+    payload: { feature: AoiFeature }
+  ): void;
   (action: 'aoi.selection', payload: { selected: boolean }): void;
-  (action: 'aoi.update', payload: { feature: AoiFeature }): void;
-};
+}
```
danielfdsilva, line 44: Nice! 👍

app/scripts/utils/date.ts
```diff
@@ -143,4 +151,26 @@ export function inputFormatToDate(inputFormat: string) {
   return parse(inputFormat, 'yyyy-MM-dd', new Date());
 }
 
-
+export type DateRangePreset =
+  | 'thisYear'
```
danielfdsilva, line 155: instead of `thisYear`, `yearToDate` is a more common pattern, but as you said these are still TDB

app/scripts/components/analysis/define/index.tsx
```diff
@@ -253,16 +267,36 @@ export default function Analysis() {
                 <DropTitle>Select a date preset</DropTitle>
                 <DropMenu>
                   <li>
-                    <DropMenuItem href='#'>Preset A</DropMenuItem>
+                    <DropMenuItem
+                      role='button'
+                      onClick={() => onDatePresetClick('thisYear')}
+                    >
+                      This year
+                    </DropMenuItem>
```
danielfdsilva, line 275: `DropMenuItem` being `a`s need an `href`

app/scripts/components/analysis/define/aoi-selector.tsx
```diff
@@ -50,32 +57,46 @@ interface AoiSelectorProps {
...
     return qsFeature
       ? { ...multiPolygonToPolygon(qsFeature), id: 'qs-feature' }
       : null;
   }, [qsFeature]);
 
-  // Use the feature from the url qs as the initial state to center the map.
+  const [currentRegionPreset, setCurrentRegionPreset] =
+    useState<RegionPreset | null>(null);
+  const onRegionPresetClick = useCallback((preset: RegionPreset) => {
+    setCurrentRegionPreset(preset);
```
danielfdsilva, line 81: I'd say there's no need to store this in state and causing a rerender. You an trigger the aoi event and the maps reposition directly.

app/scripts/components/analysis/define/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+          [-180, -89],
+          [180, -89],
+          [180, 89],
+          [-180, 89],
+          [-180, -89]
+        ]
+      ],
+      type: 'Polygon'
+    }
+  }
```
danielfdsilva, line 21: To work with the AOI editor, the feature needs an `id`, otherwise if you try to edit it, the app will crash

danielfdsilva (changes requested):
This is a great start. I'd suggest you reorganize the code a little, to reduce some repetition and moved preset related stuff to dedicated files.

app/scripts/components/analysis/define/index.tsx
nerik: See https://github.com/NASA-IMPACT/delta-ui/pull/323#discussion_r1023828190

nerik:
Laying the groundwork for https://github.com/NASA-IMPACT/delta-ui/issues/212

app/scripts/components/analysis/define/aoi-selector.tsx
```diff
@@ -115,19 +133,15 @@ export default function AoiSelector(props: AoiSelectorProps) {
...
-                  <DropMenuItem href='#'>Country name C</DropMenuItem>
-                </li>
-                <li>
-                  <DropMenuItem href='#'>Country name D</DropMenuItem>
+                  <DropMenuItem
+                    role='button'
+                    onClick={() => onRegionPresetClick('world')}
+                  >
+                    World
+                  </DropMenuItem>
```
danielfdsilva, line 144: This one also needs the `href` for the time being

danielfdsilva (changes requested):

danielfdsilva:
@nerik I had a quick check to see what it would take to switch the dropdown menu items to buttons.
Code is pretty straightforward. I'll leave it here if you want to make the change:

```js
const DropMenuButton = styled(DropMenuItem).attrs({
  as: 'button'
})`
  background: none;
  border: none;
  width: 100%;
  cursor: pointer;
`;
```

Additionally, and I forgot to mention this, I'd recommend we add `data-dropdown='click.close'` to the `DropMenuItem`. This will cause the dropdown to close when an option is selected.

nerik:
@danielfdsilva Excellent suggestions, thank you

danielfdsilva (approved):
Superb!

PR #334, opened 2022-11-15
Fix tooltip sync with brushed range

hanbyul-here:
Close https://github.com/NASA-IMPACT/delta-ui/issues/326

### DIFF
--- a/app/scripts/components/common/chart/utils.ts
+++ b/app/scripts/components/common/chart/utils.ts
@@ -1,5 +1,6 @@
 import { timeFormat, timeParse } from 'd3';
 import * as d3ScaleChromatic from 'd3-scale-chromatic';
+import { NumberLiteralTypeAnnotation } from '@babel/types';
 import { UniqueKeyUnit } from '.';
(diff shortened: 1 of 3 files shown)


danielfdsilva (approved):

app/scripts/components/common/chart/utils.ts
```diff
@@ -1,5 +1,6 @@
 import { timeFormat, timeParse } from 'd3';
 import * as d3ScaleChromatic from 'd3-scale-chromatic';
+import { NumberLiteralTypeAnnotation } from '@babel/types';
```
hanbyul-here: This is an unused dependency which I thought is part of the lint rule we have. I thought GitHub action will catch this but it doesn't seem like it. Am I understanding how GitHub Action works now? @danielfdsilva ?

danielfdsilva: It should be. I also noticed that rule not throwing an error lately. Perhaps was disabled by mistake. I'll have a look

PR #341, opened 2022-11-17
Enable no-unused-vars lint rule

danielfdsilva:


.eslintrc
```diff
@@ -85,22 +85,29 @@
...
     {
       "files": ["**/*.ts", "**/*.tsx"],
-      "plugins": [
-        "@typescript-eslint"
-      ],
+      "plugins": ["@typescript-eslint"],
       "rules": {
         "prettier/prettier": "off",
-        "no-unused-vars": "off"
+        "@typescript-eslint/no-unused-vars": "error"
```
nerik, line 102: It would be nice to settle on how we express rule levels, either `0|1|2` or `off|warning|error` and stick to it. Personal preference for the numerical version but up to you.

nerik (changes requested):
Yep, that's cleaner, thank you!

.eslintrc
danielfdsilva: No preference! let's do numbers!

nerik (approved):

PR #343, opened 2022-11-17
Make one daterange to control charts on analysis page

hanbyul-here:
👉  [Preview to analysis result page with charts](https://deploy-preview-343--veda-ui.netlify.app/air-quality/analysis/results?start=2020-04-29T00%3A00%3A00.000Z&end=2022-11-12T00%3A00%3A00.000Z&datasetsLayers=no2-monthly%7Cno2-monthly-diff&aoi=paxfRyk%7DmFitqkBfab%5Bpfus%40zorm%40f%60op%40prl%40)

- close https://github.com/NASA-IMPACT/delta-ui/issues/324 https://github.com/NASA-IMPACT/delta-ui/issues/326
- set up `activeBrushDates` on analysis result page level to control all the charts on the analysis page.
- I realized that recharts need to be forced to rerender when we sync charts manually. (using `key)`, passing props needed for key (`id`). 

- But this pr introduces **the other problem** that because the whole component is being rendered (or its state is being updated), brushing becomes jagged. like you change one date, then you lose control over the mouse action which I think is really annoying. You can compare sandbox brush on the main branch now from this pr when force rerendering is not in place: https://veda-ui.netlify.app/sandbox/analysis-chart - I will talk about this in detail at front-end standup
: **Update** - I worked around this problem by updating the brush index only when mouse gets released 🤔 Now the charts don't get synced up in real-time but the other chart 'catches up' when mouse is up.

- I replaced 'syncMethod' with 'value' for now, the chart will get synced only when there is the exact same value on x-axis. I did this because I am a bit fuzzy about how brush should work on different time densities + advanced syncing doesn't seem to be in demand based on the feedback we have collected so far. 🤔 we can revisit this in the future.

Misc
- Bigger handle
- dots for data points along the lines (This was brought up in the feedback session.)

### DIFF
--- a/app/scripts/components/analysis/results/chart-card.tsx
+++ b/app/scripts/components/analysis/results/chart-card.tsx
@@ -46,7 +46,7 @@ interface ChartCardProps {
   title: React.ReactNode;
   chartData: TimeseriesData;
   activeMetrics: DataMetric[];
-  activeBrushDates: object;
+  activeBrushDates: {start: number; end: number };
   setActiveBrushDates: (dates: { start: number; end: number }) => void;
 }
 
--- a/app/scripts/components/analysis/results/chart-card.tsx
+++ b/app/scripts/components/analysis/results/chart-card.tsx
@@ -41,12 +41,15 @@ import { Tip } from '$components/common/tip';
 import { composeVisuallyDisabled } from '$utils/utils';
 import { exportCsv } from '$components/common/chart/analysis/utils';
 import DropMenuItemButton from '$styles/drop-menu-item-button';
-
+export interface ActiveDates {
+  start: number; 
+  end: number;
+}
 interface ChartCardProps {
   title: React.ReactNode;
   chartData: TimeseriesData;
   activeMetrics: DataMetric[];
-  activeBrushDates: {start: number; end: number };
+  activeBrushDates: ActiveDates;
   setActiveBrushDates: (dates: { start: number; end: number }) => void;
 }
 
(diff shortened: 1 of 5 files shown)


danielfdsilva:
@hanbyul-here Interesting approach with the brush, but there's a problem with the syncing.
It seems that the syncing is happening via the brush index and not the date.

Check the video:

https://user-images.githubusercontent.com/1090606/203825252-a1c26345-b738-4f89-b407-08497d602b43.mov

As I change yearly, the daily shows data base on the same brush index (from 0 to whichever it is on). The result is that none of the dates match.

Re-rendering the whole chart every time is also not idea, like you pointed out. The brushes on the other charts also do not update, which was a confusing point.

I think we should all seat down and discuss a little what would the ideal behavior be with different time densities (it is tricky indeed), and then see what's possible. cc @ricardoduplos @nerik

hanbyul-here:
@danielfdsilva It is expected that sandbox page doesn't work. 1. This change requires changes from the parental level component, and I didn't touch Sandbox. 2. More importantly, I replaced 'syncMethod' with 'value' (only syncs when there is an exact matching value) apology that I didn't explain it more thoroughly in the original message. Let's say if there are two-time frames that range from 2012-2014 and 2015-2016. If we try to sync the brush, we cannot show the data for one of them at all.  (but maybe this is an unlikely scenario based on the feedback we gathered?) 

But I agree that this is a sit-down discussion topic. I will schedule something.

hanbyul-here:
We decided to move forward with a custom brush. I will cherry-pick style-related changes from this pr.

Issue #342, opened 2022-11-17
The map is not covered by a layer when selecting the same dataset where I am located


Catalina-Moller:
### Description:
When the user selects the same dataset he is in, the view is updated but the map is not covered with a layer and no default date is selected. 

**Steps to reproduce:** 
1. The user is in a dataset, for example Nitrogen Dioxide.
2. Click on the button that displays a list of all datasets in the Area in which he is.
3. Click on the dataset the user is in, in this case Nitrogen Dioxide.
4. The dataset is updated without specifying a default date and the dataset is displayed without a layer covering it. 

### Acceptance Criteria: 
When the user selects the same dataset where it is located, the view is updated, the map is covered with a layer and a default date is selected.

PR #346, opened 2022-11-18
Dataset related errors

danielfdsilva:
Addresses the error found in https://github.com/NASA-IMPACT/delta-config/pull/142 and #342

@hanbyul-here @nerik We have to be very careful with the "Unnecessary optional chain on a non-nullish value" or the "Values always truthy" suggestions of typescript. Some types may not be properly defined (like the data coming from the MDX frontmatter), and accepting those suggestions may lead to the app breaking. When faced with this we should double check the types and update accordingly.

@Catalina-Moller Can you double check the errors are fixed?

### DIFF
--- a/app/scripts/components/datasets/s-explore/index.tsx
+++ b/app/scripts/components/datasets/s-explore/index.tsx
@@ -90,10 +90,8 @@ const DatesWrapper = styled.div`
   }
 `;
 
-const isSelectedDateValid = (dateList, prevDateList, selectedDate) => {
-  const currDates = JSON.stringify(dateList);
-  const prevDates = JSON.stringify(prevDateList);
-  if (dateList && currDates !== prevDates) {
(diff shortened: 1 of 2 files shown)


Catalina-Moller:
Both issues no longer occur!
But I just noticed a strange behavior that may be related to issue https://github.com/NASA-IMPACT/delta-config/pull/142. When the dataset has more than one layer when the user changes layer it shows an error message indicating that it was not possible to load the layer but it is possible to see the layer.

I attach a [video ](https://www.screencast.com/t/UrK7rSjtD)with the evidence.

I have tried to find a pattern but I have not been able to, I think it depends on the date change but I am not sure what could be the cause.

danielfdsilva:
@Catalina-Moller I'm merging this PR to avoid blocking https://github.com/NASA-IMPACT/delta-config/pull/142, but please open an issue for the bug you just found.

PR #347, opened 2022-11-21
Analysis charts updates

danielfdsilva:
- Fixes an error happening with the requested data: Not all datasets have a `start_datetime` property. Defaulting to `datetime` property when not available.
- Scientific notation was not working with negative exponents, example: 1.5x10-3
- Updated the data in the sandbox page to better suit the demo (@hanbyul-here if you see this is not helpful we can easily remove the last commit)


nerik (approved):
LGTM 🚀

hanbyul-here (approved):
I think new datasets for the sandbox is helpful, but need more dataset (or different section) with the same time density to reflect the current dataset better. So go ahead with the pr, we can add it later.

danielfdsilva:
@hanbyul-here That's a very good point. Added another yearly dataset before merging
image

PR #373, opened 2022-12-02
Disconnect sync between charts

hanbyul-here:
This PR disconnects sync behavior between charts. This should be just a temporary solution until the new brush is ready. Feel free to close if not needed.


danielfdsilva:
Superseded by #374

Issue #356, opened 2022-11-25
Pre-release analysis page feedback


aboydnw:
## Description

This ticket is a subset of the feedback from #318 and is a proposal of the high value/low effort feedback to complete before a production release.

## Copy

- The tag-line (`Visualize insights from a selected area over a period of time.`) could be more "direct2, feel more actionable.  Something along the lines: `Generate timeseries data for your area of interest` or `Generate custom analysis for datasets in the VEDA catalog`
- `Save` . The text on the button doesn’t really reflect the action I’m taking. Based on the text, I didn’t expect to generate the analysis. This may be different when somebody is "refining" their initial search. In that case, `Cancel` and `Save` feel more natural

## Map toolbar
<img width="915" alt="image" src="https://user-images.githubusercontent.com/1090606/201096309-3fcd3d80-27a2-4df8-bf0d-30cad27dbfa4.png">

- It wasn't immediately obvious how to draw an area. I expected to be able to draw immediately on the map.
   💬 This was supposed to be the default behavior. Right @ricardoduplos?
   💬💬 If this is intended to be default and an easy fix, then let's fix. If not, I say we push off until later - AB
- Hidden Toolbar - I scrolled down to select a date first. When scrolling back up, the tools kept being covered by the Page Hero and/or the tooltip. It was only when I scrolled back up completely, that the tools appeared again. For a first-time user, that doesn’t know the toolbar exists, it might be unclear what to do.
   💬 Perhaps the map message, could include a edit action?

## Other

- All the charts should show the full selected date range even if there's no data.
- It would be good to show where the actual data points are. This is important when timesteps are skipped (example: there's data for January and March but not February)
  💬 It was suggested to show dots instead of lines, and have an option to "connect the dots" for when the users want to perceive trends.
  💬💬 If this is easy, otherwise can probably wait - AB
- Unit should be present on the chart
  💬 Units alone are not enough, we need a transformation function. (example: original data is in meters, but we want to show it in miles)
  💬💬 Units would be a good first step, transformation can probably come later - AB

aboydnw:
@danielfdsilva @hanbyul-here @ricardoduplos @nerik @faustoperez Here is my proposal for the pieces of feedback to address first. Let me know if you feel something is missing or if one of these items is more challenging that I am thinking.

danielfdsilva:
We discussed this a little and this is how we plan to address these items:

**Copy**:  
[...]

PR #374, opened 2022-12-05
Feature/new brush

nerik:
Replaces the old Recharts brush with a non-broken custom brush. Sorry it took longer than expected.
Capture d’écran 2022-12-05 à 12 10 54

This probably needs a round of design review first.
Also, I disabled the tooltip sync for now, as it needs a little bit of rework and I wasn't sure anymore what was the status of this, do we want to keep it or not?

Also fixes
> Full date range to be shown

From https://github.com/NASA-IMPACT/veda-ui/issues/356#issuecomment-1329378711


app/scripts/components/analysis/results/index.tsx
```diff
@@ -123,17 +123,32 @@ export default function AnalysisResults() {
...
+  const availableDomain: [Date, Date] | null = useMemo(
+    () => (start && end ? [start, end] : null),
+    [start, end]
+  );
+
+  const [brushRange, setBrushRange] = useState<[Date, Date] | null>(null);
+
+  useEffect(() => {
+    if (availableDomain && !brushRange) {
+      // TODO auto fit to available data? For now taking the whole user-defined range
```
danielfdsilva, line 142: This could be tricky when there are different data domains, not sure it is need, but good of you to consider it! :)

parcel-resolver-thematics/delta-thematic.out.js
```diff
@@ -0,0 +1,89 @@
+/**
```
danielfdsilva, line 1: This file is an artifact after the repo renaming. It should be deleted

danielfdsilva (changes requested):
There's still something strange with the xAxis as it is not showing dates.

image

Sorry if this is the same issue you just mentioned.


Additionally, can you check that the satndbox page (http://localhost:9000/sandbox/analysis-chart) is updated with the new brush?
image

danielfdsilva (approved):
🥇 :shipit:

PR #375, opened 2022-12-06
Fix map having the wrong size on the analysis header

### LINKED ISSUE #330: The image of the selected area is displayed smaller on the analysis results page
### Description:
At times when the user is in the Analysis view viewing the graphs the image of the selected area is displayed smaller. This happens at times when saving the selected data for the first time or after editing the selected data. 

**Evidence**: [Video](https://www.screencast.com/t/1HBrtiZHn) 
image

### Acceptance Criteria:
When the user is in the analysis view looking at the graphs, the image of the selected area should be displayed with the same size as the containing bar.

danielfdsilva:
Fix #330

### DIFF
--- a/app/scripts/components/analysis/page-hero-media.tsx
+++ b/app/scripts/components/analysis/page-hero-media.tsx
@@ -1,12 +1,14 @@
 import React, { useEffect, useRef, useState } from 'react';
 import styled, { useTheme } from 'styled-components';
-import { rgba, themeVal } from '@devseed-ui/theme-provider';
 import mapboxgl from 'mapbox-gl';
 import { Feature, MultiPolygon } from 'geojson';
 import bbox from '@turf/bbox';
 import { shade } from 'polished';
+import { rgba, themeVal } from '@devseed-ui/theme-provider';
+import { reveal } from '@devseed-ui/animation';
 
 import { SimpleMap } from '$components/common/mapbox/map';
+import { useEffectPrevious } from '$utils/use-effect-previous';
 
 const WORLD_POLYGON = [
   [180, 90],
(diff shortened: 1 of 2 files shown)


Catalina-Moller (approved):
With the change made it is now possible to see with the correct size the selected area of the map to perform the analysis!

app/scripts/components/analysis/page-hero-media.tsx
```diff
@@ -26,18 +28,45 @@ const mapOptions: Partial<mapboxgl.MapboxOptions> = {
...
+  // then expand but the map will not.
+  // The solution:
+  // If the header starts as stuck we delay the mounting of the map so by the
+  // time the map mounts, the parent already has its final size. In the case
+  // that the header is already unstuck when the page loads, this is not needed.
+  const [shouldMount, setShouldMount] = useState(!isHeaderStuck);
+  useEffectPrevious(
+    ([wasStuck]) => {
+      if (!shouldMount && wasStuck && !isHeaderStuck) {
+        const tid = setTimeout(() => setShouldMount(true), 350);
```
nerik, line 58: If I understood correctly this matches the animation duration defined [here](https://github.com/NASA-IMPACT/veda-ui/pull/375/files#diff-9fd873109ee581a07519c6310070b2505bcc1ad2df4e4d5dc7cec9da890d04f5L75-L76). Try to store this in a variable and reuse?

nerik (approved):
First class debugging, well done 🕵️  

This is a hack, so that's not ideal. I also believe this is a minor bug that doesn't deserve a clean(er) solution. Which, I'm not sure what it is. One thing that might or might not be better could be to sync the end of the animation with mounting the map, using the [animationend](https://developer.mozilla.org/en-US/docs/Web/API/Element/animationend_event) event, but not sure if worth the trouble. Maybe at least store and reuse the animation duration? (see comment)

PR #378, opened 2022-12-09
Ignore brush size when calculating chart size

danielfdsilva:
With the brush outside the recharts component, we no longer need to account for its size when converting to image.
This was causing the image to be clipped. Example:

image

cc @Catalina-Moller


nerik (approved):
Good catch, thanks!

danielfdsilva:
@nerik Rebased to fix conflicts :)

Issue #212, opened 2022-07-24
AOI Selection via Predefined Bounds


aboydnw:
## Description
As a basic scientist
I want to select an AOI using predefined bounds
So that I can more quickly select the area that I am interested in

Part of the Analysis page: https://www.figma.com/proto/SA0u0sVdw5vM7g9ulV5Xgw/Wireframes?page-id=1498%3A2972&node-id=1601%3A3246&viewport=-615%2C134%2C0.52&scaling=min-zoom

This ticket came to be based on some feedback from Alexey when reviewing wire of the analysis page. Essentially, many of their users will have predefined bounds that they care about (state, county, fire district, etc.) 

TBD on exactly which predefined bounds we will want to include in our first version.

## Acceptance Criteria
- [ ] Given I have clicked the Edit button on the Analysis page, when the sidebar opens up then I see pre-defined bounds 
- [ ] Given I have clicked the Edit button on the Analysis page, when I select a pre-defined AOI then the map and coordinates respond to reflect that selection

aboydnw:
Related to, but not dependent on #205

aboydnw:
Questions:
1. What sorts of pre-defined options would be most useful for scientists?
2. Are these options specific to the datasets they want to view?
3. Are these options specific to the region / continent that they are viewing?
4. Will scientists want to define their own regions to be referenced later?

ashiklom:
> What sorts of pre-defined options would be most useful for scientists?

The obvious one is standard political geographic units — e.g., countries, states, counties.

Beyond that, there are discipline- or application-specific ones. For example, an important unit for fires is [NIFC Predictive Service Areas](https://data-nifc.opendata.arcgis.com/maps/e580f3fd4d644366b121676714d69c2d), and an important one for air quality is [Air Quality Management Districts](https://ww2.arb.ca.gov/california-map-local-air-district-websites).

> Are these options specific to the datasets they want to view?

I think analysis context is more important than dataset — i.e., if you're analyzing fires, you might want to be able to look at as many datasets as possible, including ones that aren't _strictly_ about fires but may provide relevant context or downstream impacts (e.g., precipitation, weather, soil moisture, land cover). Not sure how to structure this from a UI standpoint without overwhelming users — maybe, have a "basic" set of units that are fairly common and universal (like political boundaries), and then an "advanced" menu that provides a longer searchable list of these.

> Are these options specific to the region / continent that they are viewing?

[...]

PR #381, opened 2022-12-12
Upload custom AoI

nerik:
Still have to deal with simplification and MultiPolygons, but I'd very much like some feedback on the UX of this :)

https://github.com/NASA-IMPACT/veda-ui/issues/212

https://user-images.githubusercontent.com/1583415/207591497-0aa21853-0ecf-45b8-b3b0-a23ca5260713.mov

### DIFF
--- a/app/scripts/components/analysis/define/aoi-selector.tsx
+++ b/app/scripts/components/analysis/define/aoi-selector.tsx
@@ -106,13 +106,11 @@ export default function AoiSelector({
 
   return (
     <Fold>
-      {aoiModalRevealed && (
-        <AoIUploadModal
-          setFeature={setFeature}
-          revealed={true}
-          onCloseClick={() => setAoIModalRevealed(false)}
-        />
-      )}
+      <AoIUploadModal
+        setFeature={setFeature}
+        revealed={aoiModalRevealed}
+        onCloseClick={() => setAoIModalRevealed(false)}
+      />
       <FoldHeader>
         <FoldHeadline>
           <FoldTitle>Area</FoldTitle>
--- a/app/scripts/components/analysis/define/aoi-selector.tsx
+++ b/app/scripts/components/analysis/define/aoi-selector.tsx
@@ -24,6 +24,7 @@ import {
 } from '@devseed-ui/collecticons';
 import { multiPolygonToPolygon } from '../utils';
 import { FeatureByRegionPreset, RegionPreset } from './constants';
+import useCustomAoI from './use-custom-aoi';
 import {
   Fold,
   FoldHeader,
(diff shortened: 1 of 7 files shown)


app/scripts/components/analysis/define/aoi-selector.tsx
```diff
@@ -94,8 +102,17 @@ export default function AoiSelector({
...
+
   return (
     <Fold>
+      {aoiModalRevealed && (
+        <AoIUploadModal
+          setFeature={setFeature}
+          revealed={true}
+          onCloseClick={() => setAoIModalRevealed(false)}
+        />
+      )}
```
danielfdsilva, line 115: It would be better to use the `revealed` prop, as in doing so the modal animated in/out.

app/scripts/components/analysis/define/use-custom-aoi.ts
```diff
@@ -0,0 +1,110 @@
...
+          geojson = JSON.parse(rawGeoJSON as string) as FeatureCollection;
+        } catch (e) {
+          setError('Error uploading file: Invalid JSON');
+          return;
+        }
+      } else {
+        geojson = await shp(reader.current.result);
+      }
+      const feature: Feature = geojson.features[0];
+      if (!feature) {
```
danielfdsilva, line 48: Can the type be more specific to avoid the warning: "Unnecessary conditional, value is always falsy"

danielfdsilva (changes requested):

danielfdsilva:
Functionality is really cool! :)

app/scripts/components/analysis/define/aoi-selector.tsx
nerik: Good point - that's what I had initially. But doing this, it's not necessary to reset state on modal re-opening (as the component is unmounted). But I guess this needs to be done anyways.

danielfdsilva: yeah, that is true. In the past I've just used `useEffect` with revealed.

nerik:
Now with MultiPolygon to Polygon logic and geometry simplification:

https://user-images.githubusercontent.com/1583415/207907426-16b3fa15-d4f5-4dd8-b592-f3f0a8a252f0.mov

app/scripts/components/analysis/define/aoi-selector.tsx
nerik: https://github.com/NASA-IMPACT/veda-ui/pull/381/commits/8e32aaf868d2ac1548d2b36e0ff022c728c95208

app/scripts/components/analysis/define/aoi-upload-modal.tsx
```diff
@@ -0,0 +1,158 @@
...
+  }, []);
+
+  const onConfirmClick = useCallback(() => {
+    if (!feature) return;
+    setFeature(feature);
+    onCloseClick();
+  }, [feature, setFeature, onCloseClick]);
+
+  useEffect(() => {
+    if (!revealed) reset();
```
danielfdsilva, line 80:
```suggestion
    if (revealed) reset();
```

I'd recommend resetting on reveal, otherwise any feedback message disappears before you see the modal fade out.

danielfdsilva (changes requested):

ricardoduplos (approved):

danielfdsilva (approved):
@ricardoduplos and I made a couple of changes related to the styling.
Have a look, and let me know if you have any questions. If all good merge!

nerik:
Looks fantastic ✨ 
Thank you, merging!

PR #393, opened 2022-12-16
Scrollable dropdown

danielfdsilva:
Create a scrollable dropdown using a shadow-scrollbar.

image

The srollbar appears as the user interacts with the dropdown. Shadows on top/bottom indicate that there is more data

### DIFF
--- a/app/scripts/components/common/dropdown-scrollable.tsx
+++ b/app/scripts/components/common/dropdown-scrollable.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import styled from 'styled-components';
+import { glsp } from '@devseed-ui/theme-provider';
+import {
+  Dropdown,
+  DropdownRef,
+  DropMenu,
+  DropTitle
+} from '@devseed-ui/dropdown';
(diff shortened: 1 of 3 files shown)


app/scripts/components/common/dropdown-scrollable.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ShadowScrollbar } from '@devseed-ui/shadow-scrollbar';
+
+/**
+ * Override Dropdown styles to play well with the shadow scrollbar.
+ */
+const DropdownWithScroll = styled(Dropdown)`
+  padding: 0;
+  overflow: hidden;
+
+  ${DropTitle} {
```
nerik, line 19: I had no idea you could do that! 👏

nerik (approved):
Looks good. Worth adding to `devseed-ui` ?

danielfdsilva:
On one hand the pieces are already part of devseed ui and they're very easy to combine, on the other maybe an extra export of `Dropdown` wouldn't hurt? 🤔 Or perhaps as a [recipe](http://ui.ds.io/latest/?path=/docs/recipe-overview--page)?

PR #394, opened 2022-12-19
Ts warnings

danielfdsilva:
This started out as fixing TS warning on the raster timeseries file which caused errors on other files

### DIFF
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -22,20 +22,20 @@ const LOG = true;
 
 const FIT_BOUNDS_PADDING = 32;
 
-export interface MapLayerRasterTimeseriesProps {
+interface MapLayerRasterTimeseriesProps {
   id: string;
   stacCol: string;
-  date: Date;
+  date?: Date;
   mapInstance: mapboxgl.Map;
   sourceParams: object;
-  zoomExtent: [number, number];
+  zoomExtent?: [number, number];
   onStatusChange?: (result: { status: ActionStatus; id: string }) => void;
   isHidden: boolean;
 }
 
-export type StacFeature = {
+export interface StacFeature {
   bbox: [number, number, number, number];
-};
+}
 
 enum STATUS_KEY {
   Global,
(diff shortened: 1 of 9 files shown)


app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -206,16 +207,16 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
...
+      changeStatus({ status: 'idle', context: STATUS_KEY.StacSearch });
     };
   }, [id, changeStatus, stacCol, date]);
 
   //
   // Markers
   //
   useEffect(() => {
-    if (!id || !stacCol || !date || !minZoom || !stacCollection) return;
+    if (!id || !stacCol || !date || !minZoom || !stacCollection.length) return;
```
danielfdsilva, line 219: This was an actual bug since `stacCollection` is always defined, just an empty array

nerik: Nice catch!

nerik: Careful with `minZoom` - `-Infinity` is truthy.

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -63,7 +63,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
 
   const theme = useTheme();
   const primaryColor = theme.color?.primary;
-  const minZoom = zoomExtent?.[0];
+  const minZoom = zoomExtent?.[0] ?? -Infinity;
```
nerik, line 66: Careful here - `0` would work just as well and this could cause trouble on line 219.

nerik (commented):
Please double check what's happening with `minZoom`, otherwise good to go!

nerik (approved):

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
danielfdsilva: You're absolutely right! Thank you :)

Issue #389, opened 2022-12-15
When changing layers inside a dataset, an incorrect error message is displayed


Catalina-Moller:
Description:
In datasets that have more than one layer it happens that when changing from one layer to another the error message "Failed to load layer co2-mean", for example, is presented indicating that the dataset was not loaded correctly when in fact it has been loaded correctly. 

Steps to reproduce it:

1. Enter a Dataset that has more than one layer.
2. Enter layer 1.
3. Enter layer 2.
4. Re-enter layer 1. 
5. The error message "Failed to load layer (layer name)" appears but the layer loads correctly.

This is one way to reproduce it, at other times only when changing from one layer to another the error message is presented. 

Evidence:
https://www.screencast.com/t/KbHtCgibk 

![image](https://user-images.githubusercontent.com/111894012/207935177-8f969aec-b6f6-4e8b-a974-1d94773a1a3e.png)

PR #401, opened 2022-12-22
Fix map error when switching layers

danielfdsilva:
The #389 error was caused by a race condition adding layers sources to the map.
This should fix it. @Catalina-Moller can you have a look?

@nerik If you have a different (cleaner) suggestion for handling cross dependencies I'm all ears. Explaining the problem and leaving the dependencies out of the array was what I was able to come up with, without re-writing the whole layer manager

### DIFF
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -373,17 +373,29 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
         mapInstance.removeSource(id);
       }
     };
-    // The showMarkers and isHidden dep are left out on purpose, as visibility
+  }, [
+    // The `showMarkers` and `isHidden` dep are left out on purpose, as visibility
     // is controlled below, but we need the value to initialize the layer
     // visibility.
-  }, [
-    id,
-    changeStatus,
-    stacCol,
+
     stacCollection,
-    date,
+    // This hook depends on a series of properties, but whenever they change the
+    // `stacCollection` is guaranteed to change because a new STAC request is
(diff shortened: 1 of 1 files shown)


Catalina-Moller (approved):
The changes fixed the issue!
When a user enters a dataset that has more than one layer and changes the view between them, the message "Failed to load layer (layer name)" no longer appears, when it has loaded correctly. 

Evidence from the preview: [Video ](https://www.screencast.com/t/doz2ApNkpk9)

nerik (approved):
This is a good fix and thank you for taking the time to explain the solution.
I don't have a better short-term fix to suggest.

If a layer management revamp is planned at some point, I think the whole approach would benefit from two things:
- get MapboxGL out of the way completely in the React realm (ie having the Mapbox GL ref constantly around, listening to events and all sorts of imperative logic, dealing with layer types, mixing layer styles and state handling, etc). [react-map-gl](https://visgl.github.io/react-map-gl/) works by using a style.json object as a single source of truth in a purely declarative way (it does diffing of the style.json internally the same way React does virtual DOM diffing). In app land the only thing needed is a purely deterministic function that converts layers to a style json.
- use a state management library. Ideally we should not have to deal with fetch statuses, instantiating abort controllers, caching, URL sync, etc. Could also be a nice opportunity to try something new, I'm personally kinda stuck at Redux and I would love to learn something new. 

Thoughts?

danielfdsilva:
I agree that something that abstracts away layer management would be handy. Perhaps we'll be able to play with something like react-map-gl soon!

We're using [react-query](https://tanstack.com/query/v4/docs/react/overview) for other requests, which takes care of caching and such. I don't remember why it isn't used for the layer stac requests, but maybe we can easily do that!?

Issue #385, opened 2022-12-13
Quickly Revisit Past Analysis


aboydnw:
## Description

In a similar vein to saving AOI selections for later use, scientists would like the ability to revisit certain search criteria that they had before. 

Disclaimer: This ticket is not intended to have us set up user profiles. It might be used as part of the conversation for adding user profiles, should that decision ever be made, but is not the intent of this specific ticket.

## Acceptance Criteria
1. Given I am a scientist using the analysis page in the Dashboard, when I have certain search parameters that I would like to revisit, then I can revisit at a later date without having to enter in the same search parameters or having to copy and paste the URL in a separate notepad

## Feedback
Pasted from #318 

- A way to go back to a previous search I did. Some of this requires trial and error, and going back to a previous search would be great. This doesn’t have to be fancy. Just a list with x last searches would be great already.

aboydnw:
## Implementation Notes
- We can start with a quick way for users to access recent searches, and also notify users that if they want to specifically save something for later they can bookmark the URL
-

PR #422, opened 2023-01-12
Added retrieve previous settings feature

nerik:
🛑 Given this feature's dependency on functionalities from https://github.com/NASA-IMPACT/veda-ui/pull/438, it can only be merged after it.

----

Screenshot 2023-01-12 at 15 09 29

Adds very bare bone functionality for https://github.com/NASA-IMPACT/veda-ui/issues/385

Need to discuss:
- How to set up the Dropdown's `DropMenu` for long labels?
- Proper placement and labeling of this button
- Should this button also be present on the results page?
- Do we need to save previous settings __per thematic area__?
- Should we try to represent the AoI on the saved preset label?


danielfdsilva:
- How to set up the Dropdown's DropMenu for long labels?
I imagine this would be something more complex than a drop menu, and I like the idea of showing the AOI. Perhaps something along the following. @ricardoduplos Can you help with designing this? Handling long names is definitely a concern. It should be clear, so not sure we can hide it behind ellipsis.


image

- Should this button also be present on the results page?
For quick access it may be a good idea. However, what would the behavior be? On click open the analysis immediately? On the define page it populates the different fields. cc @ricardoduplos 

- Do we need to save previous settings per thematic area?
Yes. Since the datasets depend on the thematic area, they may not be available in another.

ricardoduplos (approved):

danielfdsilva:
QA Expected behavior:

From the analysis definition or results page it is possible to access the last 5 analysis. This list is specific to each thematic area, since each has its own analysis.
- Selecting a past analysis from the definition page prefills all the form elements
- Selecting a past analysis from the results page shows the analysis immediately

karitotp:
This is working with expected behavior in all thematic areas!

> From the analysis definition or results page it is possible to access the last 5 analysis. This list is specific to each thematic area, since each has its own analysis.
> - Selecting a past analysis from the definition page prefills all the form elements
> - Selecting a past analysis from the results page shows the analysis immediately

danielfdsilva (approved):
Good to go after #438

Issue #414, opened 2023-01-05
Thematic Area Welcome Page


aboydnw:
## Description
We have received feedback a few times from NASA and EIS that the thematic area welcome page would be more useful if it contained the information & configurability currently housed in the thematic area about pages. We are in discussions with NASA about the long term direction of the dashboard, how thematic areas might become less prominent, and there may have to be alternative solutions to this. (see this [slack thread](https://developmentseed.slack.com/archives/C0136FF156X/p1672957296183759)) However, at the moment, EIS is sending people directly to https://www.earthdata.nasa.gov/dashboard/eis

As a short-term remedy, we would like to include some space for additional information on the thematic area welcome page. Whether this means deprecating the thematic area about pages and shifting that content to the welcome page, providing more space for configurable content like text, images, videos, or an alternative solution is up in the air. We have some latitude to propose the best way forward, so long as users landing on the url above don't have to click Learn More to view information about EIS.

This should free us up to move towards a longer term solution for thematic areas, navigation, welcome pages, etc.

## Acceptance Criteria
- [ ] Given I am a user of the VEDA Dashboard, when I visit the EIS thematic area welcome page then I see some content (that might include text, images, and videos) that has been defined by the EIS group describing the EIS project

aboydnw:
@danielfdsilva @hanbyul-here @nerik @ricardoduplos @faustoperez take a look at this ticket and let me know what you think. This came up during a conversation today with the EIS team as a short-term compromise so we can continue moving towards our long-term vision. We can discuss during backlog refinement. cc @michaelgsuttles

danielfdsilva:
Moving the configurability of the about page to the homepage is pretty easy. The only thing we have to decide is what to do with the content we currently have there (feat discoveries/datasets and other thematic areas).

A couple of options are:
- Remove them entirely
- Add the configurable content, before, after or at any point between what's there
- Provide "non configurable components" (like `<FeaturedDatasets />`) so that users can decide themselves where to place them.

@ricardoduplos @faustoperez thoughts?

michaelgsuttles:
[...]

PR #428, opened 2023-01-16
Move thematic about content to welcome page

danielfdsilva:
Based on the discussion we had about #414 , this PR moves the user created content from the about page to the welcome page.

Notwithstanding all the consideration around the need/relevance of "team pages", could this be useful as a solution while the new vision is shaped up? cc @aboydnw


aboydnw (approved):
This looks good to me! Seems to meet the need that EIS is talking about in the short term.

PR #432, opened 2023-01-19
Add Google Tag Manager

### LINKED ISSUE #425: Google Tag Manager
## Description

I pulled up the [quickstart guide](https://developers.google.com/tag-platform/tag-manager/web) for google tag manager. Comparing to the original site tag in #209  , it LOOKS different. If so, I'd love to install google tag manager as well. it would allow us to get much more granular with our analytics. So we can understand specific actions users are taking. Right now we are stuck at just “page view” and that level of specificity.

hanbyul-here:
Add Google Tag Manager. (I think GA can be connected through Google Tag Manager console, so I got rid of it from the code base.)
Close #425


danielfdsilva:
I'm not super familiar with google tag manager. How do we test if this is working? I don't see any network request on the network tab.

- Is some other configuration needed in the tag manager console? Who's setting this up?
- Is this enough to track pageviews for example?
- Do we need to track other events as well?

hanbyul-here:
I am still trying to familiarize myself with GTM, but I at least see the network request for `gtm.js` from the page, can you confirm if the script is not being loaded? ( Full URL for the script is `https://www.googletagmanager.com/gtm.js~`)

My understanding is that GTM operator can set up custom events through the dashboard without the developer's help, and this is why we are trying to move to GTM. Ex. They can set up a 'browser history change' event for pageview through the container. (from  https://support.google.com/tagmanager/answer/12131703?hl=en#zippy=%2Cmanage-automatic-event-detection)

danielfdsilva:
@hanbyul-here I can confirm that the script is being loaded, but no events are being sent. I assume this means that GTM is not yet configured. Should we configure it properly before merging? Otherwise we risk loosing events.
Who can/will configure this?

hanbyul-here:
I just checked GTM, it looks like the basic events are configured, including page views, scrolls, outbound clicks, site search (not sure what it means?), video engagement, file downloads. ( I will share the gtm container URL with you through Slack) But the event is supposed to stream from our production, so I don't think we will see the events on our preview URL. 

This is why I mentioned the need to set up dev environment for analytics. 🤔 I wanted to wait for Anthony to come back to set up a dev environment but will look into how much effort it takes to set up a dev env.

hanbyul-here:
After a struggle, I found a useful tool that I could use without messing with the dashboard too much. https://tagassistant.google.com/


Then, the tool says that our container is not published yet. When I use the Analytics ID (which is also our Tag ID), the preview tool works, and I can't figure out where this container id is from (I got it from Anthony). I will wait for Anthony to come back to coordinate this.

hanbyul-here:
Hey @aboydnw , are the tags set up? We can push this, but we are going to lose events if tags are not set up yet.

hanbyul-here:
@danielfdsilva , network request part, @nerik shared this with me :  https://stackoverflow.com/questions/64665422/google-tag-manager-network-request 

Debugger preview:  https://tagassistant.google.com/#/?source=TAG_MANAGER&id=GTM-KHK5WJQ&gtm_auth=bsx_OITtJ12rqLurB6QmRw&gtm_preview=env-10&cb=7410430937401725

.env
```diff
@@ -17,5 +17,7 @@ MAPBOX_STYLE_URL='mapbox://styles/covid-nasa/ckb01h6f10bn81iqg98ne0i2y'
 # Google form for feedback
 GOOGLE_FORM = 'https://docs.google.com/forms/d/e/1FAIpQLSfGcd3FDsM3kQIOVKjzdPn4f88hX8RZ4Qef7qBsTtDqxjTSkg/viewform?embedded=true'
 
-# Google analytics tracking code
-GOOGLE_ANALYTICS_ID='G-CQ3WLED121'
+# Google Tag Manager tracking code
+GOOGLE_TAG_MANAGER_ID='GTM-KHK5WJQ'
+GOOGLE_TAG_AUTH='tqd01PxWYuaeIneNeoNm1g'
+GOOGLE_TAG_PREVIEW='env-1'
```
nerik, line 23: Aren't these supposed to be secrets, in which case shouldn't we have them in `env.local` ?

nerik (commented):
Still trying to figure out whether we need to explicitly trigger events... Otherwise looks good, please make sure the GA codes are not sensitive and can be placed in the .env file.

.env
hanbyul-here: Hey yeah so I am not sure tag manager id is sensitive or not, but it is a better idea to move them when I am not sure! so I moved them to Netlify, and github action. I took this as an opportunity to set up a different key for dev environment.

nerik (approved):

Issue #404, opened 2023-01-03
AoI Selection via Custom-defined polygons 


nerik:
Adds support for upload of a custom AoI polygon. Currently limited to single polygon shapefile or geojson files.

Catalina-Moller:
Hi @nerik, I'm not quite clear on what I should be testing in this task. Could you help me with a bit more context? Thanks!

danielfdsilva:
Basic behavior:
- Possibility to upload a geojson or shapefile to define an AOI
- If the given file is invalid display and error message
- If the given file has more than 1 polygon, use the first one and warn the user
- If the uploaded polygon is very complex, simplify it and warn the user

The uploaded polygon should be displayed on the map.
The [test-files.zip](https://github.com/NASA-IMPACT/veda-ui/files/10449345/test-files.zip) contains some files that can be used to test the behavior. https://geojson.io can be used to generate additional test data.

cc @Catalina-Moller

Catalina-Moller:
[descargar (7).zip](https://github.com/NASA-IMPACT/veda-ui/files/10457915/descargar.7.zip)
This is the zip with the Shapefile format that I used for the first observation, maybe it can be useful for you :)

PR #434, opened 2023-01-20
Add support for multiple polygons in the AOI tool

### LINKED ISSUE #423: Multiple Polygons
## Description

As a user I want to select multiple polygons for analysis so that I can conduct the most accurate analysis possible

Currently, you can only select one polygon in the analysis page. This came up during user research in a couple different ways.

1. Analyzing islands (this feels like the most straightforward use case)
2. Uploading a shapefile that contains all the watersheds in North America. This feels like a slightly different use case, because I would assume that scientists then want to narrow down their selection to 1 or 2 watersheds. They just have a shapefile that contains all of them, because that's the format that USDA (or similar agency) publishes (Moved to #426 )

There are probably other use cases, but those two came up during user research this week.

## Acceptance Criteria

- [ ] Given I am creating a search on the analysis page, when I create multiple polygons using the AOI drawing tool then the resulting time-series charts show the total spatial aggregation of both polygons as one (I am assuming the scientists won't want separate spatial aggregations for each polygon, basing off the islands use case above)
- [ ] Given I am creating a search on the analysis page, when I create multiple polygons then I am able to remove individual polygons as needed

danielfdsilva:
Fix #423 

This took longer and touched way more files than expected. We need to test this pretty well to ensure nothing broke.

Files linked in https://github.com/NASA-IMPACT/veda-ui/issues/404#issuecomment-1397074923 can be used to test.

@ricardoduplos We also need to do a review of the actions to understand if they make sense.

### DIFF
--- a/app/scripts/components/analysis/define/use-custom-aoi.ts
+++ b/app/scripts/components/analysis/define/use-custom-aoi.ts
@@ -4,7 +4,7 @@ import shp from 'shpjs';
 import simplify from 'simplify-js';
 import { multiPolygonToPolygons } from '../utils';
 
-import { featureCollection } from '$components/common/aoi/utils';
+import { makeFeatureCollection } from '$components/common/aoi/utils';
 import { round } from '$utils/format';
 
 const extensions = ['geojson', 'json', 'zip'];
@@ -45,14 +45,15 @@ function useCustomAoI() {
   const [uploadFileError, setUploadFileError] = useState<string | null>(null);
   const [uploadFileWarnings, setUploadFileWarnings] = useState<string[]>([]);
   const reader = useRef<FileReader>();
-  const [fc, setFc] = useState<FeatureCollection | null>(null);
+  const [featureCollection, setFeatureCollection] =
+    useState<FeatureCollection | null>(null);
 
   useEffect(() => {
     reader.current = new FileReader();
 
     const setError = (error: string) => {
       setUploadFileError(error);
-      setFc(null);
+      setFeatureCollection(null);
       setUploadFileWarnings([]);
     };
 
@@ -118,7 +119,6 @@ function useCustomAoI() {
         return;
       }
 
-
       // Simplify features;
       const originalTotalFeaturePoints = features.reduce(
         (acc, f) => acc + getNumPoints(f),
@@ -128,7 +128,7 @@ function useCustomAoI() {
       let tolerance = 0.001;
       let simplifiedFeatures = [...features];
 
-      while (numPoints > 200 || tolerance > 5) {
+      while (numPoints > 200 && tolerance < 5) {
         simplifiedFeatures = simplifiedFeatures.map((feature) =>
           simplifyFeature(feature, tolerance)
         );
@@ -150,8 +150,8 @@ function useCustomAoI() {
 
       setUploadFileWarnings(warnings);
       setUploadFileError(null);
-      setFc(
-        featureCollection(
+      setFeatureCollection(
+        makeFeatureCollection(
           simplifiedFeatures.map((feat, i) => ({
             id: `aoi-upload-${i}`,
             ...feat
@@ -203,14 +203,14 @@ function useCustomAoI() {
   }, []);
 
   const reset = useCallback(() => {
-    setFc(null);
+    setFeatureCollection(null);
     setUploadFileWarnings([]);
     setUploadFileError(null);
     setFileInfo(null);
   }, []);
 
   return {
-    fc,
+    featureCollection,
     onUploadFile,
     uploadFileError,
     uploadFileWarnings,
--- a/app/scripts/components/analysis/define/use-custom-aoi.ts
+++ b/app/scripts/components/analysis/define/use-custom-aoi.ts
@@ -1,8 +1,11 @@
-import { Feature, MultiPolygon, Polygon } from 'geojson';
+import { Feature, FeatureCollection, MultiPolygon, Polygon } from 'geojson';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import shp from 'shpjs';
 import simplify from 'simplify-js';
-import { multiPolygonToPolygon } from '../utils';
+import { multiPolygonToPolygons } from '../utils';
+
+import { featureCollection } from '$components/common/aoi/utils';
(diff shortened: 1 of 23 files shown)


app/scripts/components/analysis/page-hero-media.tsx
```diff
@@ -74,18 +76,16 @@ function PageHeroMedia(props: PageHeroMediaProps) {
...
 
-    // Quick copy.
-    const featureInverse = JSON.parse(
-      JSON.stringify(feature)
-    ) as Feature<MultiPolygon>;
+    // Convert to multipolygon to use the inverse shading trick.
+    const aoiInverse = combineFeatureCollection(aoi);
     // Add a full polygon to reverse the feature.
-    featureInverse.geometry.coordinates[0].unshift(WORLD_POLYGON);
+    aoiInverse.geometry.coordinates[0].unshift(WORLD_POLYGON);
```
nerik, line 82: I'd follow eslint advice here:
```suggestion
    aoiInverse.geometry.coordinates[0] = [WORLD_POLYGON, ...aoiInverse.geometry.coordinates[0]];
```

app/scripts/components/analysis/results/use-analysis-params.ts
```diff
@@ -22,16 +22,25 @@ type AnalysisParamsNull = Omit<Partial<AnalysisParams>, 'errors'> & {
...
-  start: undefined,
-  end: undefined,
+  start: new Date('2020-01-01'),
+  end: new Date('2020-03-01'),
   datasetsLayers: undefined,
   aoi: undefined,
   errors: null
 };
 
+export class ValidationError extends Error {
```
nerik, line 35: 👍

app/scripts/utils/polygon-url.ts
```diff
@@ -1,27 +1,29 @@
...
 
 /**
- * Decodes a multi polygon string converting it into a MultiPolygon feature.
+ * Decodes a multi polygon string converting it into a FeatureCollection of
+ * Polygons.
  *
- * lon,lat|lon,lat||lon,lat|lon,lat
- * || separates polygons
- * | separates points
+ * lon,lat|lon,lat||lon,lat|lon,lat || separates polygons | separates points
```
nerik, line 9: This comment is not true anymore since we used `google-polyline`

nerik (changes requested):
Great work, I was also surprised by the amount of files that needed changing, but this is good technical foundation.

There are two comments that needs to be addressed and/or discussed (the while loop and calling `_controls` on the map), the rest is code style.
One overall code style comment: I've seen instances of `fc`, `featureColl`, and `featureCollection`: can we try to stick to one of those? (strong personal preference for `featureCollection`).

nerik:
Also: do we still want to evaluate Mapshaper as a more robust way to simplify geometries at some point? If so worth opening an issue?

app/scripts/components/common/aoi/use-aoi-controls.ts
```diff
@@ -16,49 +21,87 @@ export function useAoiControls(initialState: Partial<AoiState> = {}) {
...
   }, [initialState]);
 
   const onAoiEvent = useCallback((action, payload) => {
     switch (action) {
+      case 'aoi.trash-click': {
+        // Find the mapDraw control by checking for properties.
+        // We need to programmatically access the mapbox draw trash method which
+        // will do different things depending on the selected mode.
+        // @ts-expect-error _controls does exist but it is an internal property.
+        const mbDraw = mapRef.current?.instance?._controls.find(
```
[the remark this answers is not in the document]
danielfdsilva: Good call!

app/scripts/components/analysis/define/use-custom-aoi.ts
```diff
@@ -75,52 +80,84 @@ function useCustomAoI() {
...
+        0
+      );
+      let numPoints = originalTotalFeaturePoints;
       let tolerance = 0.001;
-      while (numPoints > 2000) {
-        feature = simplifyFeature(feature as Feature<Polygon>, tolerance);
-        numPoints = getNumPoints(feature as Feature<Polygon>);
+      let simplifiedFeatures = [...features];
+
+      while (numPoints > 200 || tolerance > 5) {
```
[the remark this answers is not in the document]
danielfdsilva: Indeed. 💯

app/scripts/components/analysis/utils.ts
```diff
@@ -44,15 +46,31 @@ export function getFilterPayload(
   return filterPayload;
 }
 
-export function multiPolygonToPolygon(
-  feature: Feature<MultiPolygon>
-): Feature<Polygon> {
+export function multiPolygonToPolygons(feature: Feature<MultiPolygon>) {
```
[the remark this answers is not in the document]
danielfdsilva: It is inferred by the return type. You think we should be verbose?

image

app/scripts/components/analysis/page-hero-media.tsx
danielfdsilva: We're already mutating the object, my reasoning is that this way just makes it clearer. Nevertheless, no strong opinion on it.

app/scripts/utils/polygon-url.ts
danielfdsilva: indeed 💯

danielfdsilva:
@nerik Replaced the instances of `fc` to makes it more consistent as you suggested.
I'd say we can re-evaluate the use of mapshaper later if this approach is causing too much trouble

app/scripts/components/analysis/page-hero-media.tsx
nerik: I have a vaguely strong opinion that this is easier to read (I mean, `unshift` really is a badly named method)

app/scripts/components/analysis/utils.ts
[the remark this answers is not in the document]
nerik: I meant useful as documentation. A JSDoc block would serve the same purpose.

app/scripts/components/analysis/page-hero-media.tsx
danielfdsilva: yes it is! terrible really :p

Issue #430, opened 2023-01-18
Chart not displaying the first value


danielfdsilva:
The charts in the analysis page are showing the first value outside the available chart area and it is not possible to access it.
In the example below it is not possible to select `2021/09`

<img width="699" alt="image" src="https://user-images.githubusercontent.com/1090606/213211215-5d72b77f-f2c0-41bc-95b1-2f4b6e2fec25.png">

URL: http://localhost:9000/covid-19/analysis/results?start=2021-09-22T00%3A00%3A00.000Z&end=2023-01-06T00%3A00%3A00.000Z&datasetsLayers=no2-monthly&aoi=znpsMmytzFe`fMgjAdyFljcC|~sK`dY

danielfdsilva:
@nerik The data is fetched taking into account the item's start and end date (internal stac item properties). Since the item represents a monthly time period it starts a `2021-09-01` and ends at `2021-09-30` that is why t shows up.
This would also happen if the dataset was yearly. In that case you'd get back a dataset from  `2021-01-01`. This is technically correct but a tricky issue.

My guess is that the brush area should be expanded to include the fetched data. Thoughts?

PR #435, opened 2023-01-23
Include min and max dates of all datasets in brush domain

nerik:
Fix for https://github.com/NASA-IMPACT/veda-ui/issues/430

Screenshot 2023-01-23 at 11 16 51

### DIFF
--- a/app/scripts/components/analysis/results/index.tsx
+++ b/app/scripts/components/analysis/results/index.tsx
@@ -133,16 +133,28 @@ export default function AnalysisResults() {
 
   const availableDomain: [Date, Date] | null = useMemo(() => {
     if (!start || !end) return null;
+    let minDate = +start;
+    let maxDate = +end;
+    requestStatus.forEach((item) => {
+      if (item.data?.timeseries) {
+        const itemDates = item.data.timeseries.map((t) => +new Date(t.date));
+        const itemMin = min(itemDates) ?? Number.POSITIVE_INFINITY;
+        const itemMax = max(itemDates) ?? Number.NEGATIVE_INFINITY;
+        if (itemMin < minDate) minDate = itemMin;
+        if (itemMax > maxDate) maxDate = itemMax;
+      }
+    });
+
     const onlySingleValues = requestStatus.every(
       (rs) => rs.data?.timeseries.length === 1
     );
-    if (!onlySingleValues) return [start, end];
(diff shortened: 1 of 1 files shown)


app/scripts/components/analysis/results/index.tsx
```diff
@@ -133,16 +133,28 @@ export default function AnalysisResults() {
...
+    let maxDate = +end;
+    requestStatus.forEach((item) => {
+      if (item.data?.timeseries) {
+        const itemDates = item.data.timeseries.map((t) => +new Date(t.date));
+        const itemMin = min(itemDates) ?? Number.POSITIVE_INFINITY;
+        const itemMax = max(itemDates) ?? Number.NEGATIVE_INFINITY;
+        if (itemMin < minDate) minDate = itemMin;
+        if (itemMax > maxDate) maxDate = itemMax;
+      }
+    });
```
danielfdsilva, line 146: @nerik Thoughts on using reduce to not mutate values?

```suggestion

    const { minDate, maxDate } = requestStatus.reduce(
      (acc, item) => {
        if (item.data?.timeseries) {
          const itemDates = item.data.timeseries.map((t) => +new Date(t.date));
          const itemMin = min(itemDates) ?? Number.POSITIVE_INFINITY;
          const itemMax = max(itemDates) ?? Number.NEGATIVE_INFINITY;

          return {
            minDate: itemMin < acc.minDate ? itemMin : acc.minDate,
            maxDate: itemMax > acc.maxDate ? itemMax : acc.maxDate
          };
        }
        return acc;
      },
      { minDate: +start, maxDate: +end }
    );
```

nerik: In that specific case (correct me if I'm wrong), there are no mutating values as minDate and maxDate are primitive data types and thus immutable. I avoid using `reduce` when not forced to, as it is slightly harder to reason about than other looping structures, but it's really down to personal preference. Maybe it's good to make a habit of using reduce for the majority of cases where you'd actually be mutating something. 

You have the final word.

danielfdsilva: You're right, nothing is mutating just re-assigning values. I tend to avoid this since it may be really hard to know how may times things get changed, and I like to consistently not do it unless necessary. But then again, this is such a small case that it really is matter of personal preference.
Will update just to keep consistent with other parts of the app.

danielfdsilva (approved):

PR #438, opened 2023-01-24
Add support for multiple polygons in the AOI tool

danielfdsilva:
This is exactly the same pull request as https://github.com/NASA-IMPACT/veda-ui/pull/434

I merged the other one, but then reverted since this is a big feature and would really benefit from some more formal QA.
Apologies for the confusion.

Original issue  #423

### DIFF
--- a/app/scripts/components/analysis/define/use-custom-aoi.ts
+++ b/app/scripts/components/analysis/define/use-custom-aoi.ts
@@ -4,7 +4,7 @@ import shp from 'shpjs';
 import simplify from 'simplify-js';
 import { multiPolygonToPolygons } from '../utils';
 
-import { featureCollection } from '$components/common/aoi/utils';
+import { makeFeatureCollection } from '$components/common/aoi/utils';
 import { round } from '$utils/format';
 
 const extensions = ['geojson', 'json', 'zip'];
@@ -45,14 +45,15 @@ function useCustomAoI() {
   const [uploadFileError, setUploadFileError] = useState<string | null>(null);
   const [uploadFileWarnings, setUploadFileWarnings] = useState<string[]>([]);
   const reader = useRef<FileReader>();
-  const [fc, setFc] = useState<FeatureCollection | null>(null);
+  const [featureCollection, setFeatureCollection] =
+    useState<FeatureCollection | null>(null);
 
   useEffect(() => {
     reader.current = new FileReader();
 
     const setError = (error: string) => {
       setUploadFileError(error);
-      setFc(null);
+      setFeatureCollection(null);
       setUploadFileWarnings([]);
     };
 
@@ -118,7 +119,6 @@ function useCustomAoI() {
         return;
       }
 
-
       // Simplify features;
       const originalTotalFeaturePoints = features.reduce(
         (acc, f) => acc + getNumPoints(f),
@@ -128,7 +128,7 @@ function useCustomAoI() {
       let tolerance = 0.001;
       let simplifiedFeatures = [...features];
 
-      while (numPoints > 200 || tolerance > 5) {
+      while (numPoints > 200 && tolerance < 5) {
         simplifiedFeatures = simplifiedFeatures.map((feature) =>
           simplifyFeature(feature, tolerance)
         );
@@ -150,8 +150,8 @@ function useCustomAoI() {
 
       setUploadFileWarnings(warnings);
       setUploadFileError(null);
-      setFc(
-        featureCollection(
+      setFeatureCollection(
+        makeFeatureCollection(
           simplifiedFeatures.map((feat, i) => ({
             id: `aoi-upload-${i}`,
             ...feat
@@ -203,14 +203,14 @@ function useCustomAoI() {
   }, []);
 
   const reset = useCallback(() => {
-    setFc(null);
+    setFeatureCollection(null);
     setUploadFileWarnings([]);
     setUploadFileError(null);
     setFileInfo(null);
   }, []);
 
   return {
-    fc,
+    featureCollection,
     onUploadFile,
     uploadFileError,
     uploadFileWarnings,
--- a/app/scripts/components/analysis/define/use-custom-aoi.ts
+++ b/app/scripts/components/analysis/define/use-custom-aoi.ts
@@ -89,12 +89,12 @@ function useCustomAoI() {
         ];
       }
 
-      if (feature.geometry.type === 'MultiPolygon') {
+      if (feature.geometry.type === 'MultiPolygon' || (feature.geometry.type === 'Polygon' && feature.geometry.coordinates.length > 1)) {
         warnings = [
           ...warnings,
           'Files containing multiple polygons are not supported. Only the first polygon will be used.'
         ];
-        feature = multiPolygonToPolygon(feature as Feature<MultiPolygon>);
+        feature = multiPolygonToPolygon(feature as Feature<MultiPolygon | Polygon>);
       } else if (feature.geometry.type !== 'Polygon') {
         setError(
           'Wrong geometry type. Only polygons or multi polygons are accepted.'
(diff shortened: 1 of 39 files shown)


danielfdsilva:
The remaining thing on this feature is to improve the AOI controls. I had a discussion with @ricardoduplos about this and will implement as follows:

- The aoi controls will be placed on top of the map, so that they're placed within their context of use.
- There will be a button group for edit/selection modes, so it is clear which one of th two is active
- To delete features and vertices a popup with a trash icon will be shown when said elements are selected. (bringing the action to the source)

danielfdsilva:
This is now ready for a formal review and QA.

Besides supporting drawing multiple polygons, the map controls and behaviour were improved.

image
The buttons were placed on top of the map for ease of access.
- The button group allows the user to toggle between selection and edition mode
- The clock back button will clear the whole map.

@ricardoduplos Do these icons make sense?

popover
The delete option was moved to a map popover. When selecting a vertex the popover appears on top of it and only that vertex is deleted. When selecting a whole feature, the popover appears in its centroid and the whole feature is removed.

---
### Custom popover

The popover is custom built, instead of using the mapbox one. This has some advantages:
- Easier to style and allows us to more easily use React to add contents.
- Includes in and out animations.
- Since it is not inside the map, it can overflow it. Not relevant for a trash button, but it would be if it was a popover with content.

I had created this popover for another project and just improved the code / migrated to typescript. There are probably improvements that can be done to it. Suggestions are welcome.

image

danielfdsilva:
@karitotp Adding you as a reviewer for QA

danielfdsilva:
Addressed #444. cc @Catalina-Moller

hanbyul-here:
The custom popover seems a bit clunky. When I scroll up and down, the popover jumps. 

https://user-images.githubusercontent.com/4583806/215212251-22d06a6e-e6db-402b-bdcf-b771a435e477.mov

danielfdsilva:
@hanbyul-here This was happening because when the header shrunk the body height changed and the popover's position was not being updated. Added a `ResizeObserver` that tracks the change and the issue is fixed.

ricardoduplos (approved):

app/scripts/components/common/mapbox/mb-popover/styled.ts
```diff
@@ -0,0 +1,241 @@
+import styled, { css } from 'styled-components';
+import { glsp, multiply, themeVal } from '@devseed-ui/theme-provider';
```
hanbyul-here, line 2:
```suggestion
import { glsp, themeVal } from '@devseed-ui/theme-provider';
```

app/scripts/components/analysis/define/use-custom-aoi.ts
```diff
@@ -75,52 +81,83 @@ function useCustomAoI() {
...
+      let numPoints = originalTotalFeaturePoints;
       let tolerance = 0.001;
-      while (numPoints > 2000) {
-        feature = simplifyFeature(feature as Feature<Polygon>, tolerance);
-        numPoints = getNumPoints(feature as Feature<Polygon>);
+      let simplifiedFeatures = [...features];
+
+      while (numPoints > 200 && tolerance < 5) {
+        simplifiedFeatures = simplifiedFeatures.map((feature) =>
+          simplifyFeature(feature, tolerance)
```
hanbyul-here, line 133: I tested with this file (state parks in Hawaii): https://gist.github.com/hanbyul-here/fd587e3cb106a3458b39f78adb69249f , I got a critical error.

I think `simplifyFeature` simplifies polygons to two coordinates level (line), and made `union` function impossible. It might make sense to validate simplified polygons.

danielfdsilva:
@karitotp I can't see your video. It says the `file is corrupt`. however from your description, it seems that you're referring to the following:
image

If that is the case, this is expected and is a feature. It's the behavior described in the comment above under "Custom popover".

app/scripts/components/analysis/define/use-custom-aoi.ts
danielfdsilva: @hanbyul-here You found something very interesting here. That is exactly what was happening and I fixed it by checking that each polygon doesn't get below 4 points.
The other thing that I saw with this dataset is that some of the polygon had rings (holes). That is not something currently supported by the editor, so I added some functionality to remove the holes and warn the user.

image

Even though I don't think this dataset represent a real use case (areas are very small) it was great to use it to test!

@nerik The fact that each feature may be simplified to the point of becoming invalid is really problematic. For now this works, but we should look more closely at that other library you suggested.

danielfdsilva:
@karitotp Once you've finished all your tests and the results are satisfactory please give your approval to the PR. :)

app/scripts/components/analysis/define/use-custom-aoi.ts
hanbyul-here: Yeah, I agree that the dataset is unlikely to be used for the dashboard. 
Thanks for handling the issue! Since the editor can't offer 'ring' type polygons, isn't it better to allow users to upload that type of polygons through the upload modal?

danielfdsilva: The idea is that the different options lead to the same outcome. You'd not be able to edit those polygons properly with the editor, so I don't think we should allow them.
Not sure if having rings would even be a use case when performing analysis.

karitotp:
The simplification of the nodes distorts a lot of the original shape and also generates new small polygons.

hanbyul-here (approved):

danielfdsilva:
I replaced the `simplify-js` module with `@turf/simplify`. Turf uses `simplify-js` under the hood but it can handle features instead of only points, making it behave a bit better.
@karitotp This should prevent new polygons from appearing. Can you check? Also, can you provide your test file for us to check?

I also changed the tolerance multiplier. It was too high and tolerance grew too quickly making changes very drastic. This will decrease performance but shouldn't be noticeable. I also increased the number of allowed points as it was too low. We may need to tweak these values further.

Geometry simplification is not easy, maybe we should think about different approaches like using bboxes for complicated polygons. I don't think that detailed polygons are necessary for this type of analysis as the data may not even have the needed resolution, but this is something that we should confirm.
I'll open an issue to track this. For the time being let's just ensure things don't break drastically.

cc @nerik

nerik:
> Geometry simplification is not easy, maybe we should think about different approaches like using bboxes for complicated polygons.

Have you looked into `mapshaper` eventually?

app/scripts/components/common/aoi/use-aoi-controls.ts
```diff
@@ -16,68 +22,130 @@ export function useAoiControls(initialState: Partial<AoiState> = {}) {
...
+        });
+        break;
+      case 'aoi.select-click':
+        setAoi((state) => {
+          return {
+            ...state,
+            drawing: false,
+            selected: false,
+            selectedContext: undefined,
+            actionOrigin: null
```
nerik, line 62: Throughout that switch statement there's a lot of resetting state going on, could we DRY it a little bit?

```suggestion
            ...DEFAULT_PARAMETERS
```

karitotp:
@danielfdsilva, Now the shape of the original polygons improved, and there are no new polygons. 

I just tested with two files, one with 8 polygons and the other one with only one polygon, and I see that when I tested with the file that has just one polygon the simplification of the geometry is less than a file with more polygons. Is this how simplification works?

Screenshot from 2023-02-01 10-10-35

Here the files that I am using:
[corridors_MX.zip](https://github.com/NASA-IMPACT/veda-ui/files/10558414/corridors_MX.zip)

Another thing, when the file name is big, the upload button gets distorted

Screenshot from 2023-02-01 10-16-19

danielfdsilva:
@karitotp yes. We need to ensure that the total amount of points is below a certain threshold. If there are 8 polygons each one will be simplified more. If there's only one, then then the amount of simplification needed to be below the threshold is less.

karitotp (approved):

nerik (approved):

PR #445, opened 2023-01-30
Add ADR with decision about dashboard v1 architecture

danielfdsilva:


.adr/001-dashboard-v1-architecture.md
```diff
@@ -0,0 +1,88 @@
...
+- Effort and time to implement
+- Support of needed functionalities
+
+## Considered Options
+
+- [1] Maintain current architecture
+- [2] CMS
+- [3] Hybrid
+- [4] Veda Datastore
+
```
hanbyul-here, line 30: I forgot to bring this up last time, but 'Replicability of the instance' (Somebody being able to fork veda-config repo and replace the contents with the same dashboard functionality) should be mentioned somewhere. Users will be able to use the search functionality without extra set up with option [1], but not with [2] or [3]

danielfdsilva: Good call! Will add it

.adr/001-dashboard-v1-architecture.md
```diff
@@ -0,0 +1,96 @@
+# Dashboard V1 architecture
```
hanbyul-here, line 1: nitpick: the title gives the impression that this doc is about the dashboard architecture in total. Can we narrow down the scope of the title to something like "Dashboard search functionality architecture"?

nerik:
Didn't know about ADR, this is great. All good for me content-wise. 

What's the reason for storing those documents under a folder with a dot prefix? Why not under `docs` ?

nerik (approved):

danielfdsilva:
@nerik no reason for the folder - needed one and named it such 😅
The [MADR](https://adr.github.io/madr/) docs use `docs/decisions`. I'm fine using the same, or maybe `docs/adr`?

.adr/001-dashboard-v1-architecture.md
hanbyul-here: Yeah, the decision's impact will be all over the dashboard, but the decision was about how to implement the search functionality. I will suggest 'Dashboard Catalog' or 'Dashboard Search'

hanbyul-here: yup sounds good 👍

danielfdsilva:
@nerik markdown footnotes?! I did not know this! Nice! 👏

hanbyul-here (approved):

danielfdsilva:
@michaelgsuttles You think this doc correctly captures our discussion around the architecture for next iteration?

Additionally, should we rename this to v2 as per our latest discussion?

michaelgsuttles:
This makes sense to me. Question: if we needed a complex site with, say, 300-500 datasets and the ability to search by metadata of the datasets, with some datasets linking to the Analysis Page and some linking to JupyterHub (and some to both), would this solution work for that? Or is there a situation (or tipping point) where we would want to move to a CMS?

danielfdsilva:
With that amount of datasets we'd move to solution 3 - having a third party search system.
We could still load data from the files as we're doing, but we'd probably need to be clever in how we structure our dataset lists.

A CMS becomes more relevant when we want to consider content authoring workflows, with forms, preview, and moving away from GH and a file based storage.

michaelgsuttles:
@danielfdsilva OK perfect! Thanks for the info... It all makes sense to me now :)

michaelgsuttles (approved):

PR #446, opened 2023-01-30
Notebook connect

danielfdsilva:
Contributes to https://github.com/NASA-IMPACT/veda-architecture/issues/174
This sets up the baseline for the notebook connect widget for veda.

Currently the notebook connect presents itself as a button in the dataset overview and explore pages:

| | |
|--|--|
|image|image|

Clicking the button will open a modal with the content that is currently on the usage page for each dataset.

image

> ⚠️ If the dataset has no `usage` defined on the MDX file, the button will not be available.

There is also an option to add a callout to a long form text page (like discoveries). Here the user will have to specify some text for the callout and the dataset for which to load the usage. Example:  
  

image

The code to generate this being:
```jsx
<NotebookConnectCallout datasetId='no2'>
  Lorem sed risus ultricies tristique nulla. Sed faucibus turpis in eu mi
  bibendum neque egestas congue. Mi proin sed libero enim sed. Odio aenean
  sed adipiscing diam donec adipiscing tristique risus.
</NotebookConnectCallout>
```

What are your thoughts @j08lue @abarciauskas-bgse @michaelgsuttles?

----

## Content updates

Feel free to suggest a name change for the button and modal title.  
The modal content is currently dynamically generated from the [`usage` property](https://github.com/NASA-IMPACT/veda-config/blob/develop/docs/CONTENT.md#datasets). We can change the common hardcoded portion if needed.

----

## Next steps

This is what we propose for a first release. Next up is having the button on top of charts and small maps in the long form pages. This requires some more complex technical changes, but wanted to have something to start with.

### DIFF
--- a/app/scripts/components/common/mapbox/map.tsx
+++ b/app/scripts/components/common/mapbox/map.tsx
@@ -16,6 +16,7 @@ import { aoiCursorStyles, useMbDraw } from './aoi/mb-aoi-draw';
 import ProjectionSelector from './projection-selector';
 import { useMapboxControl } from './use-mapbox-control';
 import { convertProjectionToMapbox } from './projection-selector/utils';
+
 import { round } from '$utils/format';
 
 mapboxgl.accessToken = process.env.MAPBOX_TOKEN ?? '';
--- a/app/scripts/components/common/mdx-content.js
+++ b/app/scripts/components/common/mdx-content.js
@@ -16,6 +16,7 @@ import {
   LazyScrollyTelling,
   LazyMap
 } from '$components/common/blocks/lazy-components';
+import { NotebookConnectCalloutBlock } from '$components/common/notebook-connect';
 
 function MdxContent(props) {
   const pageMdx = useMdxPageLoader(props.loader);
@@ -37,7 +38,8 @@ function MdxContent(props) {
           Map: LazyMap,
           ScrollytellingBlock: LazyScrollyTelling,
           Chart: LazyChart,
-          CompareImage: LazyCompareImage
+          CompareImage: LazyCompareImage,
+          NotebookConnectCallout: NotebookConnectCalloutBlock
         }}
       >
         <pageMdx.MdxContent />
--- a/app/scripts/components/common/notebook-connect.tsx
+++ b/app/scripts/components/common/notebook-connect.tsx
@@ -1,0 +1,151 @@
+import React, { useCallback, useState } from 'react';
+import styled from 'styled-components';
+import { multiply, themeVal } from '@devseed-ui/theme-provider';
+import { Button, ButtonProps } from '@devseed-ui/button';
+import { CollecticonCode } from '@devseed-ui/collecticons';
+import { Modal } from '@devseed-ui/modal';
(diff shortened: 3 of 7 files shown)


j08lue:
Great to see this feature taking shape, thanks, @danielfdsilva!

> Feel free to suggest a name change for the button and modal title.

### Dataset page

So we have

image

Could we do this:

image

The point is: It is not just code the user might want to use, it could also be links to load the data in QGIS or so. If you do not like the here/elsewhere distinction, I would just say:

image

### Exploration page

We have:

image

I think this button looks like a map control (zoom, layers, etc). I would not expect that this is my gateway to other places I can use this data. Can we place this somewhere else than on the map and also include the same label/text as on the Dataset page (Use data [elsewhere] or so), please? At the bottom of the side panel, for example?

danielfdsilva:
We are looking to have a consistent language for this feature so that it can be present throughout the application and can then be easily included in the [updated dashboard vision](https://docs.google.com/document/d/1UT1bK7wTivPvSsTGP5073PISwYjbSsSTrhg6NizmnLM/edit) (Dashboard V1).

So far we've identified essentially 3 different contexts where this notebook connect widget would appear:
- Dataset overview and explore
- Long form content (e.g. Discoveries) in the form of a callout
- Long form media content (charts, maps, figures)

The way the feature presents itself should be fairly similar as to be recognisable by the user. The current approach relied on having a trigger button that displays a modal, however this button is not visible or prominent enough being confused with map controls.

Some points from a discussion on this:
- Moving the list of links to the side panel leads to a solution that does not scale, nor works for the Dashboard V1.
- Having the button on top of the media content (charts, maps, etc) has the same issue of lack of clarity.
- Simply changing the color of the button can clash with the visual language of the dashboard.

A global cohesive solution has us going back to the drawing board.

In the meantime we can release this as is, which may allow us to gather additional feedback, unless it is not worth the confusion it may cause.

ricardoduplos (approved):

hanbyul-here (approved):
👍

nerik (approved):

j08lue:
> In the meantime we can release this as is, which may allow us to gather additional feedback, unless it is not worth the confusion it may cause.

Great, thank you for these reflections. As we discussed, an initial release as-is would be appreciated, so we can easily gather more feedback. 

I do not see a risk of confusion, since the usage information does not exist on many assets (and the new controls will be hidden when it is not) and we do not have (m)any production users yet who would trip on this.

michaelgsuttles:
@danielfdsilva @nerik @hanbyul-here Is this live somewhere to view? I'm still noodling a bit on exact wording for buttons but this is an awesome start!

danielfdsilva:
@michaelgsuttles Currently on staging, but not all datasets have usage info. Check [no2](https://visex.netlify.app/covid-19/datasets/no2)

michaelgsuttles:
@danielfdsilva Thanks! I think there may be some confusion with the broad meaning of "Use Data" -- would "Analyze Data in Python" be too long? Other options I can think of: "Analyze Data", "Launch Code", "Launch Jupyter Notebook"... Perhaps polling the scientists would reveal which one they find clearest?

PR #451, opened 2023-02-03
Address QA review of stage/v0.6.0

danielfdsilva:
# Problems covered by this PR
Report: https://www.dropbox.com/scl/fi/kjroygitvhc8f2kh2nhuw/Veda-Dashboard-QA-Stage-deploy-v0.6.0.paper?dl=0&rlkey=ls3fz1idxvq03hlzewfnblnnz

## Broken polygons in analysis page
We were using the double pipe `||` as a way to separate polygons in the url, however a polyline has naturally occurring double pipes, so when the polyline of a complex polgon had them, it was being broken into multiple polygons.
Replaced the separator with a semicolon `;` which is not present in the polyline.

## Overlay z-index
Updated the z-index of the dropdown and created a list of z-indices in the theme to simplify usage. Also converted the theme file to typescript which required updating the typings of `themeVal`

### DIFF
--- a/app/scripts/components/analysis/results/use-analysis-params.ts
+++ b/app/scripts/components/analysis/results/use-analysis-params.ts
@@ -107,9 +107,8 @@ export function useAnalysisParams(): {
         throw new ValidationError(sentences);
       }
 
-      // lon,lat|lon,lat||lon,lat|lon,lat
-      // || separates polygons
-      // | separates points
+      // polyline-encoding;polyline-encoding
+      // ; separates polygons
       const { geojson, errors: gjvErrors } = polygonUrlDecode(aoi);
 
       if (gjvErrors.length) {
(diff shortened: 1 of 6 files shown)


nerik (changes requested):
Good catch about the polyline encoding.

> Replaced the separator with a semicolon ; which is not present in the polyline.

Are you 100% sure of that? I couldn't find any info on what characters are included or not.
Additionally, I'm not sure semicolons are safe to use in URLs, [see](https://stackoverflow.com/questions/1178024/can-a-url-contain-a-semicolon-and-still-be-valid)

Maybe a safer bet would be to use a longer string? 
I'm also wondering if at this point we should provide retrocompatibility to URLs formed in a previous version (probably not; although something to keep in mind)

danielfdsilva:
@nerik According to the [polyline encoding](https://developers.google.com/maps/documentation/utilities/polylinealgorithm) they add 63 before converting to ASCII. The ASCII code for the semicolon is below 63, so it won't appear. 
Since the values get URI encoded I'm not too worried about using it in the url.

> Maybe a safer bet would be to use a longer string?
What do you mean?

Retro-compatibility is a good call. Agree that it is probably not needed at this point, but something we have to keep in mind!

nerik:
Oh ok, perfect then.

nerik (approved):

PR #453, opened 2023-02-03
Adapt brush size on viewport resize

nerik:


app/scripts/utils/use-resize-observer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+interface UseResizeObserverProps {
+  initialWidth?: number;
+}
+
+export const useResizeObserver = (
+  ref: MutableRefObject<Element | null>,
+  { initialWidth }: UseResizeObserverProps = {}
+) => {
+  const [width, setWidth] = useState<number>(initialWidth || 0);
```
danielfdsilva, line 11:
```suggestion
  const [width, setWidth] = useState<number>(initialWidth ?? 0);
```

app/scripts/components/common/chart/analysis/useBrush.ts
```diff
@@ -1,5 +1,6 @@
 import { scaleTime } from 'd3';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { useResizeObserver } from '$utils//use-resize-observer';
```
danielfdsilva, line 3:
```suggestion
import { useResizeObserver } from '$utils/use-resize-observer';
```

danielfdsilva (commented):
Two style related items, but all good otherwise! 💯

danielfdsilva (approved):

PR #457, opened 2023-02-07
Put y axis domain as auto

hanbyul-here:
Rechart's default behavior for y-axis is setting [0, 'auto'], which is not very helpful. Explicitly declare to use auto (from rechart doc: any element of the domain is set to be 'auto', comprehensible scale ticks will be calculated, and the final domain of axis is generated by the ticks).


nerik (approved):
LGTM 👍 
Is this needed because we have datasets with negative values?

danielfdsilva (approved):

Issue #442, opened 2023-01-26
Add support for reference information on map (coastlines, place labels)


michaelgsuttles:

danielfdsilva:
To implement this we need:

- [ ] Add the needed data to the style (via mapbox studio)
- [ ] Create the data selector control for the map (Use projection selector for inspiration)

nerik:
I figure this should be added to the map in the dataset exploration section, alongside other map controls:

<img width="559" alt="Screenshot 2023-02-06 at 11 01 57" src="https://user-images.githubusercontent.com/1583415/216942546-00cfb72d-d976-479a-9299-7d60b6ee7662.png">

Overall design questions:
- [ ] Is there a risk of confusion with the projection selector button? This planet icon pretty much already looks like a "basemap selector" icon, but maybe that's just me
- [ ] What are the context layers we want to add?
  - [ ] Place labels:
    - [ ] Should the layer mentioned above appear on top of the raster layer(s) (I'm guessing yes, same as the existing boundaries)
    - [ ] Labels will require some styling in Mapbox Studio. Who's owning that? (I can, but we'd get better results from a designer :)
  - [ ] Do we want to add administrative boundaries beyond countries/adm0? (I reckon adm1 would be useful at the very least for US States)
  - [ ] Add a graticules layer? (would also need design in MB Studio)
  - [ ] Any use for a basic topography + bathymetry basemap?
    - [ ] For UI considerations: take into account that this would _replace_ the satellite basemap, while the labels or graticules would be _added on top_ of the satellite basemap.
- [ ] Should we also add this to the Analysis AoI selection map?
<img width="291" alt="Screenshot 2023-02-06 at 11 14 28" src="https://user-images.githubusercontent.com/1583415/216945291-b2f79404-08ba-46b4-929e-eadeddd0a836.png">

As usual, I suggest we gather to define what's a good MVP for a first iteration and start building.

ricardoduplos:
Capturing the solution @nerik and I discussed for the MVP:

- The previous "Projections" action will become "Map Options", keeping the Globe icon for the button, but its scroll-able dropdown will be redesigned to harbor three groups of map-related options:
  - Style (exclusive option selection list):
    - Default light
    - Default dark
    - Satellite
  - Details (On/off togglable options list):
    - Labels
    - Boundaries
    - ...
  - Projection (exclusive option selection list)
    - Albers
    - Mercator
    - Globe
    - ...
[...]

PR #458, opened 2023-02-08
Map options dropdown

ricardoduplos:
https://github.com/NASA-IMPACT/veda-ui/issues/442

This contains markup and styles for the Map Options dropdown. ~Ready to be wired.~

https://user-images.githubusercontent.com/1583415/222201199-2b31ab16-8c3c-43c6-8c53-8a92913913aa.mov

Screenshot 2023-02-08 at 10 54 57

Related to [#basemap](https://github.com/NASA-IMPACT/veda-ui/issues/442)


app/scripts/components/common/mapbox/index.tsx
```diff
@@ -131,6 +133,15 @@ function MapboxMapComponent(props: MapboxMapProps, ref) {
   const [isMapLoaded, setMapLoaded] = useState(false);
   const [isMapCompareLoaded, setMapCompareLoaded] = useState(false);
 
+  const {
+    styleLoaded,
+    currentBasemapStyleId,
```
danielfdsilva, line 138: Only the basemap has a `current` prefix. Should this be normalized?

app/scripts/components/common/mapbox/map-options/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+              <DropMenu as='ol'>
+                {BASEMAP_STYLES.map((basemap) => (
+                  <li key={basemap.id}>
+                    <DropMenuItem
+                      href='#'
+                      active={currentBasemapStyleId === basemap.id}
+                      onClick={(e) => {
+                        e.preventDefault();
+                        if (onBasemapStyleIdChange)
+                          onBasemapStyleIdChange(basemap.id);
```
danielfdsilva, line 160:
```suggestion
                        onBasemapStyleIdChange?.(basemap.id);
```
more succinct?

app/scripts/components/common/mapbox/map-options/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+                <li>
+                  <DropMenuItem as='span'>
+                    <OptionSwitch
+                      name='labels'
+                      id='labels'
+                      value='labels'
+                      checked={labelsOption}
+                      onChange={(e) => {
+                        if (onOptionChange) {
+                          onOptionChange('labels', e.target.checked);
```
danielfdsilva, line 192:
```suggestion
                        onOptionChange?.('labels', e.target.checked);
```
more succinct?

app/scripts/components/common/mapbox/map-options/index.tsx
```diff
@@ -0,0 +1,271 @@
...
+                <li>
+                  <DropMenuItem as='span'>
+                    <OptionSwitch
+                      name='boundaries'
+                      id='boundaries'
+                      value='boundaries'
+                      checked={boundariesOption}
+                      onChange={(e) => {
+                        if (onOptionChange) {
+                          onOptionChange('boundaries', e.target.checked);
```
danielfdsilva, line 209:
```suggestion
                        onOptionChange?.('boundaries', e.target.checked);
```
more succinct?

app/scripts/components/common/mapbox/index.tsx
```diff
@@ -267,6 +278,7 @@ function MapboxMapComponent(props: MapboxMapProps, ref) {
           sourceParams={baseLayerResolvedData.sourceParams}
           zoomExtent={baseLayerResolvedData.zoomExtent}
           onStatusChange={onBaseLayerStatusChange}
+          styleLoaded={styleLoaded}
```
danielfdsilva, line 281: Adding the `styleLoaded` as a prop leads to the layer itself having to deal with this change, which may be easily forgotten in new layers.
What do you think of only mounting the layer if the `styleLoaded` is true. The effect should be the same.

Additionally, this is also needed for the compare layer.

app/scripts/components/common/mapbox/map-options/use-basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+  const styleUrl = useMemo(() => {
+    return currentBasemapStyleId
+      ? BASEMAP_STYLES.find((b) => b.id === currentBasemapStyleId)!.url
+      : BASEMAP_STYLES[0].url;
+  }, [currentBasemapStyleId]);
+
+  useEffect(() => {
+    if (!mapInstance || !styleUrl) return;
+
+    // Any BaseLayerComponent (ie raster-timeseries or others) must listen to styledata/styledataloading events
```
danielfdsilva, line 51: Is this comment still true?

app/scripts/components/common/mapbox/map-options/use-basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import {
+  BasemapId,
+  BASEMAP_STYLES,
+  GROUPS_BY_OPTION,
+  Option
+} from './basemaps';
+
+export function useBasemap(mapInstance: mapboxgl.Map | null) {
+  const [styleLoaded, setStyleLoaded] = useState(true);
```
danielfdsilva, line 11: Shouldn't this start as `false`? I think it may be participating in the error mentioned on line 99.

danielfdsilva (changes requested):
It is working well. There is some fine tuning to do, but overall it looks good.

### Url parameters

The projection is currently being stored in the url to be applied on load. Should we do the same for the base style and other options? Perhaps we could leave this for later, but the base style seems important. I imagine this requiring some changes, but should not be too tricky.

### Blocks
All the map blocks have properties to define the starting projection. Should this be extended to include a starting style and other options? Perhaps this is less critical and can be logged as an issue.

### Other
- There are some linting errors that need fixing
- I'd recommend rebasing from `main` to include latest changes.

app/scripts/components/common/mapbox/map-options/use-basemap.ts
nerik: Nope, good catch

app/scripts/components/common/mapbox/index.tsx
nerik: Yes, will do

app/scripts/components/common/mapbox/index.tsx
nerik: Well that's an excellent suggestion. Thanks

danielfdsilva (changes requested):
Fixes look good! There are still a couple of things to address:

- The style and other options also needs to be applied to the compare map and compare layer otherwise it looks strange.
image

- I think that we should have the selected style in the url and it may be easier to do now given that the implementation is fresh. However, I'm not opposed to log it as an issue and tackle later if you don't have the bandwidth right now.
- Options for blocks: Chatting about this we decided that this could be logged as a future enhancement. Can you open an issue describing the situation and link it to this pr?

nerik:
Thanks. I'll address 1), I totally overlooked that.

For 2) and 3) I created an [issue](https://github.com/NASA-IMPACT/veda-ui/issues/477) yesterday based on your initial comments.

nerik:
@danielfdsilva [This commit ](https://github.com/NASA-IMPACT/veda-ui/pull/458/commits/322189879a378c8de0169e9a8caa6bb52b00e875)probably deserves some explanation:

In order to fix the compare map problem, first idea was to either send the compare map ref to the useBasemap hook, or to instantiate two distinct hooks for the two maps.
This felt awkward as the responsibility of this hook is to manage the basemap state; so I decided to instantiate it once and remove the map ref(s) dependency(ies) completely. So that hook now loads the base style from the Mapbox style API and modifies it depending on the options. The resulting style is then directly used by the map(s). 
It felt so great to remove that imperative `map.setLayoutProperty` line, that now I want to do the same for all those `addLayer`s and `addSource`s 😁

app/scripts/components/common/mapbox/map-options/use-basemap.ts
```diff
@@ -0,0 +1,100 @@
...
+      } else {
+        setBoundariesOption(value);
+      }
+    },
+    [setLabelsOption, setBoundariesOption]
+  );
+
+  // Apply labels and boundaries options, by setting visibility on related layers
+  // For simplicity's sake, the Mapbox layer group (as set in Mapbox Studio) is used
+  // to determine wehether a layer is a labels layer or boundaries or none of those.
```
danielfdsilva, line 54: Should we document (in some dev doc), what is needed if we want to add a new base style? It seems that certain groups are required?! Or am I not getting this the right way?

danielfdsilva (changes requested):

app/scripts/components/common/mapbox/map-options/use-basemap.ts
nerik: Yes, you are right. I think the right place to document is directly in that file or basemap.ts.
Another option is to migrate the Mapbox styles directly in the source code, and store those groups as constants that we can reuse directly within the styles. But then we loose the ability to edit them within Mapbox Studio.

app/scripts/components/common/mapbox/map-options/use-basemap.ts
```diff
@@ -0,0 +1,100 @@
...
+      : BASEMAP_STYLES[0].mapboxId;
+
+    const url = getStyleUrl(mapboxId);
+    const controller = new AbortController();
+
+    const load = async () => {
+      const styleRaw = await fetch(url, { signal: controller.signal });
+      const styleJson = await styleRaw.json();
+      setBaseStyle(styleJson as Style);
+    };
```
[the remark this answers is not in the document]
nerik: [Now fetching basemap styles with react-query](https://github.com/NASA-IMPACT/veda-ui/pull/458/commits/777c38e91205a2fc90ec8b2c9feb61e6eb1b1efb)

However, it's unclear to me:
- whether caching actually works. Here style gets reloaded on basemap change, but could it be because of [dev tools](https://react-query-v3.tanstack.com/guides/important-defaults). Maybe something off on my side.
- what's the benefit of react-query when data gets cached on disk by the browser anyways
- whether we should manually cancel fetches when unmounting, since react-query seems to consider that keeping on loading the data is a [sensible default](https://react-query-v3.tanstack.com/guides/query-cancellation)

Thanks 🙏

nerik:
Hi @karitotp ! 
Would you mind doing a round of tests on this basemap options feature?
This feature is visible in the dataset exploration section, for example, https://deploy-preview-458--veda-ui.netlify.app/covid-19/datasets/no2/explore
We need to test whether this works properly in this section, also in "date comparison" mode, on several datasets. Would be nice to also do a quick round of testing in other sections where a map is visible (analysis header, analysis AoI tool,  discoveries, etc)

karitotp:
Hi @nerik! I took a look at the map in the different sections where it is and it looks good on all of them.

danielfdsilva (approved):

Issue #460, opened 2023-02-15
Make more helpful errors for MDX blocks


hanbyul-here:
Common errors from editors/datasets are:
- Incorrect combination of Block + Prose + Figure
- Incorrect date format for chart
- Incorrect column names 
- missing summaries 

We can offer more precise error messages per each error, hopefully helping editors configure the discoveries/datasets.

PR #476, opened 2023-03-01
Show a proper error message to users

hanbyul-here:
React's api `getDerivedStateFromError` that we are using to throw an error in `BlockErrorBoundary` doesn't catch async errors. This PR tries to catch async errors with a workaround - it works if  an error is thrown in useState hook. More in this thread: https://github.com/facebook/react/issues/14981 

I also put some verbose error message for Block + Figure combination error.

Close https://github.com/NASA-IMPACT/veda-ui/issues/460

### DIFF
--- a/app/scripts/components/common/blocks/index.js
+++ b/app/scripts/components/common/blocks/index.js
@@ -241,14 +241,16 @@ function BlockComponent(props) {
   );
 
   if (![defaultBlockName, wideBlockName, fullBlockName].includes(typeName)) {
-    throw new HintedError(`${blockTypeErrorMessage} '${typeName}'`);
+    throw new HintedError(`${blockTypeErrorMessage} '${typeName}'`, [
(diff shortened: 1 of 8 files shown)


danielfdsilva (approved):
@hanbyul-here Added a commit with a couple of improvements:

```jsx
// ❌ Invalid
<Block>
  <Image /> ## .... <Image />
</Block>
```
An image should not be a direct descendant of `Block`. For inline images you need a `Prose` block.


```jsx
// ✅ Valid
<Block>
  <Prose>
    <Image /> ## .... <Image />
  </Prose>
</Block>
```

I updated the message to reflect this. I also tweaked the error display so it is better formatted. It now supports `<pre>` tags and links to github docs.
image

Additionally I converted the hinted-error to typescript.

nerik (approved):
Very cool 🙇

PR #478, opened 2023-03-02
Sort alphanumerically using numeric option of collator

### LINKED ISSUE #459: Sort legend alphanumerically
Current legend doesn't consider numerical order, making the legend start with numbers sorted unexpectedly. Probably can be solved with : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/numeric 

Problem: 
Screen Shot 2023-02-15 at 10 46 30 AM

hanbyul-here:
Close #459 
Legend gets sorted  `a1,a11, a2` right now. 
This PR should fix this problem.

### DIFF
--- a/app/scripts/components/common/chart/utils.test.ts
+++ b/app/scripts/components/common/chart/utils.test.ts
@@ -1,8 +1,54 @@
 import { getFData } from '$components/common/chart/utils';
 
-describe('uniqueKey of formatted data', () => {
+describe('uniqueKey of formatted data from getFData', () => {
 
-  it('checks numeric character gets sorted', () => {
+  it('checks how alpha + nummeric characters get sorted', () => {
+    
+    const idValues = ['a6','b','a22','c', 'z'];
+
+    const uniqueKeyTestData = new Array(idValues.length)
+    .fill(0)
+    .map((_, idx) => ({
+      idKey: idValues[idx],
+      xKey: '2022',
+      yKey: '10'
+    }));
+    
+    const { uniqueKeys }  = getFData({
+      data: uniqueKeyTestData,
+      idKey: 'idKey',
+      xKey: 'xKey',
+      yKey: 'yKey',
+      dateFormat :'%Y'
+    });
+
+    expect(uniqueKeys).toEqual(['a6','a22','b','c', 'z']);
+  });
+
+  it('checks characters with diacritic marks don\'t impact the sorting', () => {
+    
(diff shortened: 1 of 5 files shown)


app/scripts/components/common/chart/utils.ts
```diff
@@ -92,8 +92,11 @@ export function getFData({
   yKey: string;
   dateFormat: string;
 }): { uniqueKeys: any[], fData: FormattedTimeSeriesData[] } {
+  // Check sensitivity value
+  const collator = new Intl.Collator(undefined, {numeric: true, sensitivity: 'base'});
```
hanbyul-here, line 96: Anybody can give me insight if it is reasonable to consider accent or other diacritic marks for sorting? (or any other values for sensitivity? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator. options are `base`, `case`, `variant`) I've been living accentless life so I have no sense.

danielfdsilva: I imagine that for the majority of use cases we don't need to worry about accents. However, I'd say to disregard accents when sorting. Consider `à` equal to `a`

app/scripts/components/common/chart/utils.test.ts
```diff
@@ -0,0 +1,80 @@
+import { getFData } from '$components/common/chart/utils';
```
danielfdsilva, line 1: yay tests! 🎉

jest.config.js
```diff
@@ -93,7 +93,8 @@ module.exports = {
         : acc;
     }, {}),
     '.+\\.(css|styl|less|sass|scss)$':
-      '<rootDir>/node_modules/jest-css-modules-transform'
+      '<rootDir>/node_modules/jest-css-modules-transform',
+    d3: '<rootDir>/node_modules/d3/dist/d3.min.js'
```
danielfdsilva, line 97: why was it needed to specify d3?

hanbyul-here: `chart/utils` imports d3 and jest has problems importing packages without `exports.require` option declared (https://github.com/facebook/jest/issues/12036#issuecomment-1039479569)

danielfdsilva: gotcha! 👍

hanbyul-here:
Flagging that https://github.com/NASA-IMPACT/veda-config/pull/193 <- legend for this data should be reconfigured with this change. (It should be formatted as `20% <` )

Issue #480, opened 2023-03-08
[Technical] remove Mapbox GL instance dependency from map layers 


nerik:
Based on this [comment](https://github.com/NASA-IMPACT/veda-ui/pull/458#issuecomment-1453340781):

Having the Mapbox GL instance ref used throughout the React tree is starting to introduce some headaches in the codebase, because it forces to call imperative methods on the Mapbox GL instance such as `addLayer` / `removeLayer`, which is not ideal in a React world.

My idea is to have a centralized method that takes in layer data and settings, and generates a style object that is used as a one-way single source of truth for the map instance. Here's a skeleton PoC of what I'd like to achieve: https://codesandbox.io/p/sandbox/map-layers-management-wi-803lc3?file=%2Fsrc%2FApp.tsx&selection=%5B%7B%22endColumn%22%3A28%2C%22endLineNumber%22%3A77%2C%22startColumn%22%3A28%2C%22startLineNumber%22%3A77%7D%5D

- [ ] Implement working version of PoC
- [ ] Remove map ref dependency from raster-timeseries

cc @hanbyul-here @danielfdsilva

hanbyul-here:
! It will be so neat to have a central place for all the layer changes. 

This may be a bit off-topic: you mentioned using react-mapbox-gl before. Can I ask what your opinion is like after working on Veda for a few months? I am not suggesting anything. The benefit that I can see from this change reminded me of what I thought we can get from using react-mapbox-gl.

ifsimicoded:
closing as stale

PR #488, opened 2023-03-14
Map style generation

nerik:
https://github.com/NASA-IMPACT/veda-ui/issues/480


app/scripts/components/common/mapbox/layers/styles.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const layersWithMeta = [
+      ...generatorParams.layers.map((layer) => {
+        const metadata = layer.metadata ?? {};
+        metadata.generatorId = generatorId;
+        return { ...layer, metadata };
+      })
+    ];
+
+    layers = [...layers, ...layersWithMeta];
+  });
```
danielfdsilva, line 65: To keep the current approach, I'd use a `foreach` instead of a `map`. Same outcome, but no return expectations.

To avoid reassigning variables, this could be a `reduce`.

danielfdsilva (commented):
This is really nice work @nerik!

Question:
Instead of getting the style out of the context (with `<Styles onStyleUpdate={onStyleUpdate}>`) and storing it in a state so that you can [pass it to the map](https://github.com/NASA-IMPACT/veda-ui/blob/e9e547c8a2ed962aa656911a8571aeef12da259b/app/scripts/components/common/mapbox/index.tsx#L454), have you considered wrapping each map in its respective context and then just using the hook from within `components/common/mapbox/map.tsx`?

nerik:
@danielfdsilva I'm not sure this is exactly what you had in mind, but [here it goes](https://github.com/NASA-IMPACT/veda-ui/pull/488/commits/f24356f3b30bd1cea548cfe6ad071fca5917f597). I think that works nicely.

PR #489, opened 2023-03-14
Index pages for datasets (future data catalog) and discoveries

danielfdsilva:
Lists content regardless of thematic area.

Since the datasets and discoveries must link to a thematic area (for the time being), they link to the first one available when one piece of content belongs to multiple.

### DIFF
--- a/app/scripts/components/data-catalog/index.tsx
+++ b/app/scripts/components/data-catalog/index.tsx
@@ -15,7 +15,7 @@ const allDatasets = Object.values(datasets)
   /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition */
   .filter((d) => !!d?.data && !!d.data.thematics?.length)
   .map((d) => d!.data)
-  .sort((a, b) => (a.name > b.name ? 1 : -1));
+  .sort((a, b) => (a.name.localeCompare(b.name));
 
 function DataCatalog() {
   return (
--- a/app/scripts/components/data-catalog/index.tsx
+++ b/app/scripts/components/data-catalog/index.tsx
@@ -15,7 +15,7 @@ const allDatasets = Object.values(datasets)
   /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition */
   .filter((d) => !!d?.data && !!d.data.thematics?.length)
   .map((d) => d!.data)
-  .sort((a, b) => (a.name.localeCompare(b.name));
+  .sort((a, b) => a.name.localeCompare(b.name));
 
 function DataCatalog() {
   return (
--- a/app/scripts/components/data-catalog/index.tsx
+++ b/app/scripts/components/data-catalog/index.tsx
@@ -1,0 +1,61 @@
+import React from 'react';
+import { datasets } from 'veda/thematics';
+
+import { LayoutProps } from '$components/common/layout-root';
+import PageHero from '$components/common/page-hero';
+import { Fold, FoldHeader, FoldTitle } from '$components/common/fold';
+import { Card, CardList } from '$components/common/card';
+import EmptyHub from '$components/common/empty-hub';
+
+import { PageMainContent } from '$styles/page';
+import { thematicDatasetsPath } from '$utils/routes';
+
+/* eslint-disable-next-line fp/no-mutating-methods */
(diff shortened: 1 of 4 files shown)


aboydnw (commented):
I'll leave for others to approve. Functionally, seems good to me. I wonder if it would be valuable to include links to the thematic areas from these pages to start, since we are still organized by thematic areas? I don't know if that's necessary, though. I'll lean on @ricardoduplos for an opinion there.

danielfdsilva:
@aboydnw What do you mean by including links to the thematic areas? Is it including the list of thematic areas on both pages?

aboydnw:
@danielfdsilva I just meant that the only way off the page was through the top nav or by clicking on a dataset/discovery. On second thought, though, I don't think it's very useful to put on this page. I was imagining having the thematic area slider until we get rid of thematic areas, but now I don't agree with that idea 😅 
image

app/scripts/components/data-catalog/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { LayoutProps } from '$components/common/layout-root';
+import PageHero from '$components/common/page-hero';
+import { Fold, FoldHeader, FoldTitle } from '$components/common/fold';
+import { Card, CardList } from '$components/common/card';
+import EmptyHub from '$components/common/empty-hub';
+
+import { PageMainContent } from '$styles/page';
+import { thematicDatasetsPath } from '$utils/routes';
+
+/* eslint-disable-next-line fp/no-mutating-methods */
```
nerik, line 13: I wish there was a way to disable this rule for `sort`, but there isn't :(
Only workaround is pretty ugly: https://github.com/jfmengels/eslint-plugin-fp/blob/master/docs/rules/no-mutating-methods.md#pass

app/scripts/components/data-catalog/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+const allDatasets = Object.values(datasets)
+  /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition */
+  .filter((d) => !!d?.data && !!d.data.thematics?.length)
+  .map((d) => d!.data)
+  .sort((a, b) => (a.name > b.name ? 1 : -1));
```
nerik, line 18:
```suggestion
  .sort((a, b) => (a.name.localeCompare(b.name));
```

app/scripts/components/root-discoveries/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+/* eslint-disable-next-line fp/no-mutating-methods */
+const allDiscoveries = Object.values(discoveries)
+  .map((d) => d!.data)
+  .sort((a, b) => {
+    const getTime = (d: string) => {
+      const millis = new Date(d).getTime();
+      return isNaN(millis) ? -Infinity : millis;
+    };
+
+    return getTime(b.pubDate) - getTime(a.pubDate);
```
nerik, line 22: What is the reason for sorting by pub date rather than discovery title? Shouldn't this be coherent with dataset sorting?

danielfdsilva: Since the discoveries are more akin to news articles, I think that novelty is important, so the most recent ones are shown first.
Datasets are more of exploratory nature so they're easier to find if alphabetical.

app/scripts/components/data-catalog/index.tsx
danielfdsilva: good call!

app/scripts/components/data-catalog/index.tsx
danielfdsilva: Well, sort is mutating and will be problematic if not used properly, so maybe it is better to intentional about it?!
The linter could be clever enough to understand that it is a copy and that it is ok! 😅

nerik: Yes that would be ideal, even `[...arr].sort()` won't make it shut up :/

app/scripts/components/root-discoveries/index.tsx
nerik: That's a sound rationale but in that case, the design should probably be different. Anyways, something for later.

PR #490, opened 2023-03-14
Redirect thematic /about to root

danielfdsilva:


hanbyul-here (approved):
This is a good reminder that we need to be careful when we rearrange URLs. Thanks for handling this!

PR #496, opened 2023-03-27
Add vector dataset support

danielfdsilva:
Add support for vector data https://github.com/NASA-IMPACT/veda-architecture/issues/60

Vector datasets are configured lie normal datasets, just needing to specify `type: vector`.

https://github.com/NASA-IMPACT/veda-ui/blob/14e02e08cf50842ee2a00742b412bfb84a1b9d47/mock/datasets/fire.data.mdx?plain=1#L13-L21

Since vector datasets are not yet in production, their api is different:
https://github.com/NASA-IMPACT/veda-ui/blob/14e02e08cf50842ee2a00742b412bfb84a1b9d47/.env#L7
@smohiudd Any idea when the vector datasets will make it to the staging api?

The styles applied to the vector tiles are the same for all datasets and non configurable:

|lines|polygons|
|--|--|
|image|image|


smohiudd:
@danielfdsilva I will get back to you about when the stac collections will move to staging. Will probably be by end of this week.

app/scripts/components/common/mapbox/layers/vector-timeseries.tsx
```diff
@@ -0,0 +1,242 @@
...
+        id: `${id}-line-bg`,
+        type: 'line',
+        source: id,
+        'source-layer': 'default',
+        layout: {
+          visibility: isHidden ? 'none' : 'visible'
+        },
+        paint: {
+          'line-color': theme.color?.['base-300'],
+          'line-width': 4
```
nerik, line 118: This could probably use some zoom-dependent styling, for instance:
```suggestion
          'line-width': [
              "interpolate",
              ["linear"],
              ["zoom"],
              minZoom,
              4,
              zoomExtent?.[1],
              10
            ]
```

app/scripts/components/common/mapbox/layers/vector-timeseries.tsx
```diff
@@ -0,0 +1,242 @@
...
+    featuresApiEndpoint,
+    minZoom,
+    isHidden,
+    haveSourceParamsChanged
+  ]);
+
+  //
+  // Listen to mouse events on the markers layer
+  //
+  useEffect(() => {
```
nerik, line 202: It would be interesting to extract that as a hook and reuse in `raster-timeseries` (provide a callback for the `onPointsClick` handler)

nerik (approved):
This looks very cool 👏 
Some code style remarks, you decide if they're worth addressing, nothing critical.
Reminder to update the `veda-config` documentation with the new `vector` type.

app/scripts/components/common/mapbox/layers/vector-timeseries.tsx
danielfdsilva: Great idea!

app/scripts/components/common/mapbox/layers/vector-timeseries.tsx
```diff
@@ -0,0 +1,242 @@
...
+              visibility: isHidden ? 'none' : 'visible',
+              'icon-offset': [0, -12]
+            },
+            maxzoom: minZoom,
+            metadata: {
+              layerOrderPosition: 'markers'
+            }
+          }
+        : undefined
+    ].filter(Boolean) as AnyLayer[];
```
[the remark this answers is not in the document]
danielfdsilva: Interesting! In this case we're not mapping anything, but constructing an array.
Maybe we could do:
```
const values = [1, 2, 3, condition ? 4 : []].flat()
```
Was this what you meant?

In this case I think this would be harder to read.

danielfdsilva:
@karitotp Can you do some tests to ensure that this behaves well?

karitotp (approved):

PR #501, opened 2023-04-02
Set up redirects for datasets/discoveries

nerik:


app/scripts/utils/routes.ts
```diff
@@ -1,11 +1,24 @@
...
+  ThematicData,
+  DiscoveryData
 } from 'veda/thematics';
 
 type ThematicOrString = VedaDatum<ThematicData> | string;
-type DatasetOrString = VedaDatum<DatasetData> | string;
+
+export const ABOUT_PATH = '/about';
+export const DISCOVERIES_PATH = '/discoveries';
+export const DATASETS_PATH = '/data-catalog';
```
danielfdsilva, line 12: I'm wondering if the path for the datasets shouldn't be `/datasets/:id`. thoughts?

danielfdsilva (changes requested):

danielfdsilva:
@nerik Check https://github.com/NASA-IMPACT/veda-ui/pull/502 first which is also part of the review. Since it was a bigger proposed change, didn't want to do it in a suggestion block

danielfdsilva (approved):

PR #502, opened 2023-04-03
Remove hasSeveralThematicAreas check

danielfdsilva:
The `hasSeveralThematicAreas` was implemented for scalability purposes. If the config only had a single thematic area, the whole app would be about it and the thematic areas would sort of be ignored. Now that we're removing them, this can also be removed, and it is a nice addition to #501


nerik (approved):
Looks good, thanks, I'll go ahead and merge.

Issue #504, opened 2023-04-06
Improve performace of Date Slider control


danielfdsilva:
On datasets that have a lot of time steps, the date slider has to render an absurd amount of points and that hinders the performance of the whole page.

- Example: https://visex.netlify.app/eis/datasets/global-reanalysis-da/explore (≈7000 time steps)

Video evidence:
https://www.dropbox.com/xframe/s/q5zc3tcddog1dqk/global.mp4?dl=0

PR #507, opened 2023-04-13
Improve date slider behavior and performance

danielfdsilva:
Related to #504

The behavior and performance improvement was achieved in essentially 2 ways:
- Only render the elements that are within the viewport extent (with some margin). This results in having little dom nodes, even on datasets with 7000+ timesteps.
- Add data breaks when there are 20 consecutive timesteps with no data.

image

@karitotp Can you run some tests to ensure this works properly?

cc @faustoperez @ricardoduplos Quick visual check?

### DIFF
--- a/app/scripts/components/common/dateslider/data-points.tsx
+++ b/app/scripts/components/common/dateslider/data-points.tsx
@@ -1,10 +1,11 @@
 import React, { useEffect, useRef } from 'react';
 import styled from 'styled-components';
-import { ScaleTime, select, Selection } from 'd3';
-import { startOfDay, startOfMonth, startOfYear } from 'date-fns';
+import { ScaleLinear, select, Selection } from 'd3';
 import { themeVal } from '@devseed-ui/theme-provider';
 
-import { DateSliderData, DateSliderTimeDensity } from './constants';
+import { findDate } from './utils';
+import { DateSliderDataItem, DateSliderTimeDensity } from './constants';
+
 import useReducedMotion from '$utils/use-prefers-reduced-motion';
 
 const StyledG = styled.g`
@@ -28,6 +29,11 @@ const StyledG = styled.g`
     stroke-width: 2px;
     stroke: ${themeVal('color.base')};
   }
+
+  .data-point-break {
+    fill: none;
+    stroke: ${themeVal('color.base-200')};
+  }
 `;
 
 const DataLineSelf = styled.line`
@@ -36,7 +42,7 @@ const DataLineSelf = styled.line`
 
 type HighlightCircle = Selection<
   SVGCircleElement,
-  Date,
+  DateSliderDataItem,
   SVGGElement | null,
   unknown
 >;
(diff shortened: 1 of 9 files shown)


karitotp:
@danielfdsilva, this is working well. Just one thing, when there is an interaction with the date slider on a layer, and you want to return to the first date, the slider is not showing the date of the left end on its own.

https://user-images.githubusercontent.com/11504548/231816990-8209407b-1548-401d-9078-a21de1f63df1.mp4

danielfdsilva:
@karitotp Problem addressed. Please approve the PR when everything is working well.

karitotp (approved):

app/scripts/components/common/dateslider/index.tsx
```diff
@@ -70,9 +74,28 @@ export default function DateSliderControl(props: DateSliderControlProps) {
...
+  const indexes = [
+    Math.max(Math.floor(x.invert(minX)), 0),
+    // Add one to account for when there's a single data point.
+    Math.ceil(x.invert(maxX)) + 1
+  ];
+
+  const dataToRender = useMemo(
+    () => data.slice(...indexes),
+    [data, ...indexes]
+  );
```
nerik, line 93: What are you trying to achieve with that `useMemo`? 
From what I understand from the whole block, this could be wrapped in a bigger useMemo with a dependency on data and zoomXTranslation. 

&lt;pedantic&gt;`indices` rather than `indexes`&lt;/pedantic&gt;

app/scripts/components/common/dateslider/data-points.tsx
```diff
@@ -78,15 +78,15 @@ export function DataPoints(props: DataPointsProps) {
     const rootG = select(container.current);
 
     const classAttr = (d, c) => {
-      return hoveringDataPoint === d.date ? `${c} over` : c;
+      return hoveringDataPoint === d.index ? `${c} over` : c;
```
nerik, line 81: Have you checked whether the performance issue was also caused by this? The whole `useEffect` block gets re-executed everytime the hovered point changes. The points could be rendered "statically" (this `useEffect`would only have a dependency on `data`), with another `useEffect` on top of it, with only a dependency on `hoveringDataPoint`, that only takes care of drawing the single hovered DOM element.

nerik (approved):
Great work 👍 

I'm not sure it's worth our time, definitely not a requirement for this PR, but it would be good to replace the d3 DOM wrangling logic with React native code placing the SVG elements in `data-points` and `data-axis`.

app/scripts/components/common/dateslider/data-points.tsx
danielfdsilva: The performance problems were particularly prominent on render, not so much on running. Adding all the DOM nodes to show 7000+ data points is what was causing the problems. Once they were in the DOM, it ran smoothly.

It is true that the useEffect would run multiple times, but once the nodes were there, d3 wouldn't do anything else to them.

app/scripts/components/common/dateslider/index.tsx
danielfdsilva: `zoomXTranslation` translation has a subpixel value, so the useEffect would run constantly.
The indexes are more stable (i.e. the `zoomXTranslation` would have to change "a lot" before having an index change).

Will update the name ;)

danielfdsilva:
@nerik Not including it as part of this PR, but a refactor of the rendering code would definitely be nice!

Issue #484, opened 2023-03-09
Detailed design of data catalog & discoveries page


aboydnw:
## PI Objective

[Implement dataset and discovery search](https://github.com/NASA-IMPACT/veda-architecture/issues/229)

## Description
We have a prototype that shows rough sketches of the direction of the Dashboard under v2. The first area we would like to develop is the data catalog. Our proposed process to get from current state to a new data catalog in production is:
1. Add a list view of all datasets currently on the dashboard, regardless of thematic area. This can look and behave the same as our current Dataset page under each thematic area, and can almost act as it's own thematic area, an "All" thematic area. This will eventually become our data catalog (maybe we just call it data catalog from the beginning?)
2. Link this view to the home page for users to jump right into it from the first page (and maybe top nav on the home page?
3. Repeat steps 1 & 2 for Discoveries
4. Layer on search, filter, and other capabilities that we want with this new data catalog view as well as the new discoveries search view
5. Deprecate current thematic area pages [this likely requires some change management from Product]

The goal of this ticket is to start designing step #4 above, so that when development gets through steps 1-3 we have these designs ready for implementation.

## Acceptance Criteria
- [ ] Detailed designs of the data catalog and discovery search views, whether represented as hi-def mockups or other assets
- [ ] Description of how our users will want search to function
- [ ] Description of different ways our users want to filter content in the data catalog and discovery view

aboydnw:
@ricardoduplos I tried to capture here what we discussed earlier today. check if you agree with the details on this ticket ahead of sprint planning please.

Also flagging for the rest of the team @nerik @danielfdsilva @hanbyul-here @faustoperez @anayeaye @ividito @j08lue as this could have cross-functional implications

danielfdsilva:
Wireframes: https://www.figma.com/proto/Tie1qnq4uZ9DS5IVjhgUGI/Dashboard-V2-Mockups?page-id=1%3A47&node-id=3-1436&viewport=-2538%2C-437%2C0.9&scaling=min-zoom&starting-point-node-id=3%3A1436

PR #508, opened 2023-04-14
Data Catalog filters/search

danielfdsilva:
Implement Data Catalog page according with the mockups (#484)

- Date indication on the card and date sorting options were not implemented (See #514 for context)
- Source is currently missing, therefore all cards day `By SOURCE` and the source filter only has the `All` options (More info on #511 )


ricardoduplos (approved):

app/scripts/components/common/card.tsx
```diff
@@ -214,18 +272,19 @@ const CardFigure = styled(Figure)`
 `;
 
 interface CardComponentProps {
-  title: string;
+  title: React.ReactNode;
```
nerik, line 275: Just out of curiosity: do we have a set practice on using React types directly, rather than importing them (`import React, { ReactNode } from 'react';`)

app/scripts/components/data-catalog/browse-controls.tsx
```diff
@@ -0,0 +1,233 @@
...
+          <DropdownOptions
+            prefix='Source'
+            items={sourcesOptions}
+            currentId={source}
+            size={isLargeUp ? 'large' : 'medium'}
+            onChange={(v) => onAction(Actions.SOURCE, v)}
+          />
+          <SearchField
+            size={isLargeUp ? 'large' : 'medium'}
+            placeholder='Title, description...'
```
nerik, line 122: Maybe add a hint here that at least 3 chars are needed.

app/scripts/components/data-catalog/index.tsx
```diff
@@ -1,22 +1,115 @@
...
+const allDatasets = Object.values(datasets).map((d) => d!.data);
+
+const DatasetCount = styled(Subtitle)`
+  grid-column: 1 / -1;
+  text-transform: uppercase;
+`;
+
+const topicsOptions = [
+  {
+    id: 'all',
```
nerik, line 41: (nit)
```suggestion
    id: TOPIC_OPTION_ALL,
```

app/scripts/components/data-catalog/index.tsx
```diff
@@ -1,22 +1,115 @@
...
+  // TODO: human readable values for Taxonomies
+  ...Array.from(new Set(allDatasets.flatMap((d) => d.thematics))).map((t) => ({
+    id: t,
+    name: t
+  }))
+];
+
+const sourcesOptions = [
+  {
+    id: 'all',
```
nerik, line 53: (nit)
```suggestion
    id: SOURCE_OPTION_ALL,
```

app/scripts/components/data-catalog/index.tsx
```diff
@@ -1,22 +1,115 @@
...
 function DataCatalog() {
+  const controlVars = useBrowserControls({
+    topicsOptions,
+    sourcesOptions
+  });
+
+  const { topic, source, sortField, sortDir, onAction } = controlVars;
+  const search = controlVars.search ?? '';
+
+  const displayDatasets = prepareDatasets(allDatasets, {
```
nerik, line 105: Might wanna memoize that?

nerik (changes requested):
Looking great! Love the attention to details like the collapsible search field on mobile 🥳 

A bunch of things:
- I had to uncomment the `thematics` field in the [urban heating dataset](https://raw.githubusercontent.com/NASA-IMPACT/veda-config/develop/datasets/urban-heating.data.mdx) for this to run. 
- (nit) cursor on the featured datasets card's thematic pill is reverting to the 'move' cursor on hover, and it has a hover style (while not clickable) which is a bit confusing. Also because it's not clickable, I'd give it a disabled style especially since the pills on the non featured cards _are_ clickable.
- Clicking on the card's thematic pills should scroll up to the filters bar IMHO. 
- Make search case-insensitive ('Car' yields results, 'car' does not)
- We probably don't need the card's top right 'DATASET' pill
- (nice-to-have) Also perform a search on thematic areas ('eis' in the search box should filter datasets with eis thematic)
- (nice-to-have) On the dataset count block: 
'Showing 2 datasets out of 45 <ins>clear filters</ins>' or 'Showing all 45 datasets'

app/scripts/components/data-catalog/index.tsx
danielfdsilva: 👍

app/scripts/components/common/card.tsx
danielfdsilva: I don't think we do. Do you favor one?

danielfdsilva:
Thank you for the very thorough review @nerik! :pray:

Some notes on your points:

> I had to uncomment the thematics field (...)

I made the thematics optional for now, but should they be required on every dataset?
Currently the Urban heating doesn't have them because it was a way to hide it (Can't find where this convo happened). They didn't want to create a dataset page but wanted to use it in a discovery. Should this be a thing? I'd vote against.

> cursor on the featured datasets (...)

Pill now behaves like any other text. I decided not to disabled it running the risk of making it too invisible. @ricardoduplos, any concerns on this point?

> Clicking on the card's thematic pills should scroll up to the filters bar IMHO.

Great idea! Done.

> Make search case-insensitive ('Car' yields results, 'car' does not)

Done

> We probably don't need the card's top right 'DATASET' pill

There are places where there is related data (like discoveries) and the cards show up. That pill is the way to tell them apart. To be consistent it is also shown here. I think @ricardoduplos was rethinking that, but maybe for later.

> Also perform a search on thematic areas (...)

Wouldn't this make the filter redundant? Or is this good redundancy? @ricardoduplos thoughts?

> 'Showing 2 datasets out of 45 clear filters' (...)

Implemented. @ricardoduplos to comment on design.
image


> Add hint to indicate needing 3 chars on txt search

@ricardoduplos Would the UI library form field hint work here?

nerik:
Thanks for addressing the feedback, and sorry for not commenting in-context, this makes things messy.

> I made the thematics optional for now, but should they be required on every dataset?
Currently the Urban heating doesn't have them because it was a way to hide it (Can't find where this convo happened). They didn't want to create a dataset page but wanted to use it in a discovery. Should this be a thing? I'd vote against.

I'm not sure but _ideally_, we'd have an explicit `hideInCatalog` or similar for something like that.

> We probably don't need the card's top right 'DATASET' pill
> There are places where there is related data (like discoveries) and the cards show up. That pill is the way to tell them apart. To be consistent it is also shown here. I think @ricardoduplos was rethinking that, but maybe for later.

Also deferring to @ricardoduplos - if we leave it I'd recommend disabling the hover styles similar to what you did for the thematic pills.

nerik (approved):

app/scripts/components/common/card.tsx
danielfdsilva: @nerik Did this in a separate PR because there were a lot of changes. https://github.com/NASA-IMPACT/veda-ui/pull/517

danielfdsilva:
@nerik @ricardoduplos Addressed all the items we discussed. :) See commits above

danielfdsilva:
@karitotp This is ready to undergo some QA testing.
If you want to use production data you'll need to checkout the `feature/remove-thematic-areas` branch from veda-config which includes the needed [changes in data](https://github.com/NASA-IMPACT/veda-config/pull/238).

karitotp (approved):

PR #517, opened 2023-04-27
Import types directly for react and mapbox

danielfdsilva:
All the React and mapbox types were not being imported, but used from the default export, for example: `React.ReactNode`. The same was true for some methods like `React.forwardRef`.

This PR standardizes the approach to always import everything. The only exception is made with `mapboxgl.Map` since it the constructor and the types have the same name ans they clash with js' `Map`.


nerik (approved):
Thanks 🙏 
One possible alternative to the Map situation:
`import mapboxgl, { Map as MapboxMap } from 'mapbox-gl';`

danielfdsilva:
@nerik good call. done!

PR #528, opened 2023-05-02
Discovery hub filters

danielfdsilva:
Implement Discovery Hub page according with the mockups (#484)

- Source is currently missing, therefore all cards day `By SOURCE` and the source filter only has the `All` options.


app/scripts/components/discoveries/hub/index.tsx
```diff
@@ -1,28 +1,159 @@
...
-    };
+  /* eslint-disable-next-line fp/no-mutating-methods */
+  filtered.sort((a, b) => {
+    if (!a[sortField]) return Infinity;
 
-    return getTime(b.pubDate) - getTime(a.pubDate);
+    return a[sortField]?.localeCompare(b[sortField]);
   });
 
+  if (sortDir === 'desc' || sortField === 'pubDate') {
```
nerik, line 118: Not sure I understand the `sortField === 'pubDate'` part. This makes sorting by date in ascending order impossible.

app/scripts/components/discoveries/hub/index.tsx
```diff
@@ -33,31 +164,138 @@ function DiscoveriesHub() {
...
+                          By SOURCE
+                        </Link>
+                        {!isNaN(pubDate.getTime()) && (
+                          <>
+                            <VerticalDivider variation='dark' />
+                            <Link
+                              to={`${DISCOVERIES_PATH}?${Actions.SORT_FIELD}=pubDate`}
+                              onClick={(e) => {
+                                e.preventDefault();
+                                onAction(Actions.SORT_FIELD, 'pubDate');
```
nerik, line 233: I'd scroll to top here too.

app/scripts/components/discoveries/hub/pub-date.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { format } from 'date-fns';
+
+function PublishedDate(props: { date: Date }) {
+  const { date } = props;
+
+  return (
+    <>
+      <span>published on</span>{' '}
+      <time dateTime={format(date, 'yyyy-MM-dd')}>
```
nerik, line 10: TIL about the <time> tag 🙌

nerik (changes requested):
Feel free to address the DRY comments on a separate PR (or not).
For the rest I see two blockers:
- The carousel is broken for me
- The sort by date issue (unless I'm missing something here)

app/scripts/components/discoveries/hub/index.tsx
danielfdsilva: Great catch! :D

app/scripts/components/discoveries/hub/index.tsx
danielfdsilva: Forgot this after adding the date

app/scripts/components/discoveries/hub/featured-discoveries.tsx
```diff
@@ -0,0 +1,117 @@
...
+  largeUp: 'content-start'
+};
+
+export const continuumFoldSpanCols = {
+  smallUp: 3,
+  mediumUp: 5,
+  largeUp: 8
+};
+
+function FeaturedDiscoveries() {
```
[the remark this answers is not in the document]
danielfdsilva: This is interesting! Not sure where it got the `type` property (`d.type === "dataset"`) as it doesn't exist.

@nerik I think the tricky thing here would be handling the fact that a discovery has a date and a dataset doesn't (maybe in the future it will have one but won't be called `pubDate`)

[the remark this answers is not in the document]
danielfdsilva: ideas?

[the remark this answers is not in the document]
danielfdsilva: I'll try something...

danielfdsilva:
@nerik I think I was able to address most of the items

app/scripts/components/discoveries/hub/index.tsx
```diff
@@ -33,31 +164,138 @@ function DiscoveriesHub() {
...
-                  parentName='Discovery'
-                  parentTo={DISCOVERIES_PATH}
-                  description={t.description}
-                  date={t.pubDate ? new Date(t.pubDate) : undefined}
-                  imgSrc={t.media?.src}
-                  imgAlt={t.media?.alt}
-                />
-              </li>
-            ))}
+            {displayDiscoveries.map((d) => {
```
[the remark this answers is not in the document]
nerik: Got it

nerik (approved):
Looking good, thanks!

Issue #498, opened 2023-03-29
Meeting with EIS Fire 2023.03.29


danielfdsilva:
Saadiq facilitated a meeting with a EIS Fire scientist to get some initial impressions/feedbach on #496.

Overall the feature was well received with the following comments:

**"homepage" for information**:  
It would be good to have a place where to information about the work EIS Fire is doing with the fire dataset. The current dataset overview page works for this, but when we split the different dataset layers (#493) there may be a lot of repeated content. A comment was also made about cross-linking to other data products and notebooks (reinforces the idea that a better notebook connection is needed (#450).

**conditional data visualization**:  
Currently the data is being filtered by time step (like all other datasets). For vector sources it would be good to allow additional filtering by other properties (in the fire case: fires over x area.). Data driven styling where a color scale is applies based on a property would also be very useful.

**accessing underlying data**:  
A way to download the data being viewed could be useful, but linking to documentation where it is explained how to do it would also suffice.

**compound datasets**:
The fire line dataset makes the most sense when viewed together with the fire perimeter, as the fire line is the part that is currently burning.

The orange here is the fireline and the blue the perimeter.

<img width="512" alt="image" src="https://user-images.githubusercontent.com/1090606/228615135-eaa49a27-2dd9-454c-89f9-5fbec9887202.png">

In this case the dataset is made of these vector 2 layers. Having a way to have an additional raster layer with the fire satellite view could also be useful.

**tooltip with information**:  
The features of a vector dataset have properties baked in that should be surfaced to the user. To avoid having machine names like the image below, scientists could provide a mapping dictionary - it would be a matter of setting up the proper config structure.

<img width="495" alt="image" src="https://user-images.githubusercontent.com/1090606/228615818-2a7a0025-5ab8-4f47-8a0c-2532b85a919f.png">

**seeing more data at low zoom levels**:  
The current approach with markers doesn't allow the user to understand where big fires are. Adjusting he zoom threshold could help with this, but other options could be considered.

@smohiudd is there anything I did not capture?

smohiudd:
Thanks @danielfdsilva, this is a great summary of the discussion.

One more point was about the option to download the vector data similar to the veda documentation (https://nasa-impact.github.io/veda-docs/example-notebooks/wfs.html). We're not sure how appropriate this is in the veda ui context but just something to note.

mccabete:
Thanks for the summary @danielfdsilva! That covers everything I can think of.

nerik:
[...]

PR #530, opened 2023-05-03
Use custom marker

danielfdsilva:
For the inclusion of the [fire dataset](https://github.com/NASA-IMPACT/veda-ui/issues/498#issuecomment-1491977465), the default blue color for markers and the vector data is not appropriate.

Given the pressing need to have this dataset in the app, and the fact that no other vector datasets exist, we're changing the default data color for all vector datasets to a shade of red. To achieve this for markers we need to use a custom image which supports [sdf](https://docs.mapbox.com/help/troubleshooting/using-recolorable-images-in-mapbox-maps/) and change its color.

⚠️ **This solution does not scale and it is temporary. We'll need to revisit the possible implementation of custom vector data styling if new vector datasets are added.**

image

image


nerik (approved):
LGTM 🚀 

Conceptually, I'd rather have images declared in the style json rather than using imperative `loadImage`/`addImage` etc. But sadly only one `sprite` containing all images can be declared, which requires generating a sprite image, which in my experience is a huge pain (for instance you can't have SDF and non SDF at the same time).

Anyways, a lot of words to say: this is a great solution, I can see myself reuse that in the future.

Issue #166, opened 2022-06-14
Zarr visualization on dashboard


hanbyul-here:
This is a placeholder ticket to investigate zarr viz on dashboard. note that our titiler doesn't support zarr format yet. (so it won't follow the pattern of how we display raster data now.) 

helpful links: 
- https://www.youtube.com/watch?v=YnTAuSOAl2c&ab_channel=opengeospatial
- https://www.npmjs.com/package/@carbonplan/maps

aboydnw:
Based on the outcome of this spike: https://github.com/NASA-IMPACT/veda-architecture/issues/57

PR #538, opened 2023-05-15
Ab/add zarr timeseries

abarciauskas-bgse:
This PR adds a component for Zarr timeseries layers which is similar too (but much simpler than) the raster timeseries component. 

Right now, the only Zarr datasets available via STAC are in the dev version of the catalog, which is why I temporarily point the STAC API URL to the dev stack. This change (to the STAC API URL) should not be merged. We should first determine if that zarr dataset in dev will be promoted to staging and / or a resolution to this ADR on including external datasets: https://github.com/NASA-IMPACT/veda-architecture/pull/258


app/scripts/components/common/mapbox/map-options/basemaps.ts
```diff
@@ -15,8 +15,8 @@ export const BASEMAP_STYLES = [
   {
     id: 'satellite',
     label: 'Satellite',
-    mapboxId: 'cldu1cb8f00ds01p6gi583w1m',
-    thumbnailUrl: `https://api.mapbox.com/styles/v1/covid-nasa/cldac5c2c003k01oebmavw4q3/static/-9.14,38.7,10.5,0/480x320?access_token=${process.env.MAPBOX_TOKEN}`
+    mapboxId: 'ckb01h6f10bn81iqg98ne0i2y',
+    thumbnailUrl: `https://api.mapbox.com/styles/v1/covid-nasa/ckb01h6f10bn81iqg98ne0i2y/static/-9.14,38.7,10.5,0/480x320?access_token=${process.env.MAPBOX_TOKEN}`
```
abarciauskas-bgse, line 19: I'm not sure if this change is necessary to have things working locally but apparently I needed it to get things working. Should it be updated?

mock/datasets/oco2-geos-l3-daily.data.mdx
```diff
@@ -0,0 +1,44 @@
+---
+id: oco2-geos-l3-daily
```
abarciauskas-bgse, line 2: This is just for demonstration purposes, not sure this dataset or its configuration will end up in the dashboard and if so it would go in veda-config anyways

danielfdsilva:
@abarciauskas-bgse Made some small changes and rebased on top of `main`.

Changes:
- The layer components now get the full layer data instead of just properties. Each component then uses what needs
- Fixed some TS typings

danielfdsilva:
Will fix https://github.com/NASA-IMPACT/veda-ui/issues/166

danielfdsilva (commented):
This should be good to go once the data makes it to the production environment.

abarciauskas-bgse:
I will update this PR to resolve conflicts soon, but more immediately need to address the titiler deployment returning an internal server error (with no logs at this time)

danielfdsilva:
@abarciauskas-bgse Is there any update on this? Are we planning to include this in the current format?

abarciauskas-bgse:
@danielfdsilva this should be good to merge. I have updated the branch with latest changes and tested locally. Waiting on the netlify preview at the moment.

abarciauskas-bgse:
@danielfdsilva it looks like there are some lint and ts-check errors unrelated to this PR - can you help me resolve them? Any advice on how to avoid this in the future would be appreciated.

abarciauskas-bgse:
Thank you @danielfdsilva 

@j08lue @anayeaye I would like to setup some automated alerts in case the xarray titiler goes down but then I think we should go ahead and add the zarr dataset to the dashboard (see https://github.com/NASA-IMPACT/veda-ui/blob/ab/add-zarr-timeseries/mock/datasets/oco2-geos-l3-daily.data.mdx) to demonstrate the capability, what do you think?

PR #540, opened 2023-05-18
Override support

danielfdsilva:
To kickstart the GHG development this PR adds override support for:
- theme
- homepage content
- page footer
- header brand (where the logo would go)
- meta tags

The meta tags override is the trickiest. I didn't find a very nice way of doing this, so currently it needs specific files in a specific folder (this part being not very configurable). Depending on how the configuration requirements move forward this would be something I'd like to discuss further.

Updated documentation is included in the PR. Check [veda-config-ghg](https://github.com/NASA-IMPACT/veda-config-ghg) for the implementation


hanbyul-here (approved):
Is there anything that needs to happen on veda-config? (It looks like we need to move our favicons and metadata to a static folder?) Can you provide a checklist?

danielfdsilva:
@hanbyul-here In this case nothing has to happen. veda-config will continue to use the files in veda-ui (the default) but can override if desired (the case of veda-config-ghg). So this PR is non-breaking

PR #541, opened 2023-05-25
Add Link component to mdx for internal links

danielfdsilva:
Avoids absolute links to other parts of the app in long form content pieces


hanbyul-here (approved):
👍 I am a bit concerned that not all the editors will have clear ideas about internal links /differentiating the components to use for internal vs. external. but the code lgtm!

Issue #542, opened 2023-05-25
Scrollytelling + raster-timeseries problems


hanbyul-here:
Currently raster-time-series component has `raster-timeseries` as its ID. This fixed ID leads to the problem when there needs to be more than one raster-timeseries layer on one map like scrollytelling component. 

Ex. 
We can observe that some layers (often the first one) fail to be displayed in this discovery: https://www.earthdata.nasa.gov/dashboard/discoveries/fire-life-cycle

PR #543, opened 2023-05-25
Add suffix props for RasterTimeseries component 

hanbyul-here:
Closes https://github.com/NASA-IMPACT/veda-ui/issues/542

This PR is a quick fix to add suffixes to rastetimeseris component for Scrollymap component. We would need a way of generating unique IDs if we are moving forward in the direction of displaying multiple layers simultaneously in the future. I feel like that is an overengineering at this moment so here is a quick fix.

### DIFF
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -358,42 +358,6 @@ function Scrollytelling(props) {
   return (
     <ScrollyMapWrapper>
       <TheMap topOffset={topOffset}>
-        <Styles>
-          <Basemap />
-          {isMapLoaded &&
-            resolvedLayers.map((resolvedLayer) => {
-              if (!resolvedLayer) return null;
-
-              const { runtimeData, Component: LayerCmp, layer } = resolvedLayer;
-
-              if (!LayerCmp) return null;
-
-              // Each layer type is added to the map through a component. This
-              // component has all the logic needed to add/update/remove the
-              // layer. Which component to use will depend on the characteristics
-              // of the layer and dataset.
-              // The function getLayerComponent() should be used to get the
-              // correct component.
-              return (
-                <LayerCmp
-                  key={runtimeData.id}
-                  id={runtimeData.id}
-                  mapInstance={mapRef.current}
-                  stacCol={layer.stacCol}
-                  date={runtimeData.datetime}
-                  sourceParams={layer.sourceParams}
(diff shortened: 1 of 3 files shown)


app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -358,42 +358,6 @@ function Scrollytelling(props) {
   return (
     <ScrollyMapWrapper>
       <TheMap topOffset={topOffset}>
-        <Styles>
```
hanbyul-here, line 361: I think I do not understand why two layers are repeating. Is this for a smooth transition?

danielfdsilva: This shouldn't happen. When the `SimpleMap` was wrapped with `Styles` the layer code was copied and this was not deleted. Great catch!

hanbyul-here:
The bug can be observed through any scrollymap component. Ex. Scrolly map component on this discovery failing to load the first layer: https://www.earthdata.nasa.gov/dashboard/discoveries/fire-life-cycle

danielfdsilva (changes requested):
I think this works as a quick fix to get the stories working again. 👍 
The suffix also needs to be include in the [unmount `useEffect`](https://github.com/NASA-IMPACT/veda-ui/pull/543/files#diff-859ada2fe95f22e3a018b9d8e1403db207fb23c99394e6164f1d524acfa22bc6L425-L435).

However this makes the layer component and the `generatorId` prop a bit confusing. Could you open an issue where you detail the problem, explain what the fix does and what could be next steps? I think we should clean this tech debt as soon as possible.

danielfdsilva (changes requested):
Missed the dependency in the unmount useEffect

hanbyul-here:
@danielfdsilva Yeah I can make a follow-up ticket but can you elaborate on how this makes `generatorId` confusing? I was thinking we can use some kinds of object-hash to dynamically generate ID (still not sure we need this at this point.) but want to check if we are on the same page

danielfdsilva:
@hanbyul-here hmm, I think the vector should also have the suffix. For now we don't have them in the scrolly, but they could be there.

The `generatorId` was referring to what "created" that layer i.e. the type of component. Perhaps renaming it to something else would suffice? Or maybe I'm just over complicating.
@nerik what was your reasoning when you created this?

hanbyul-here:
This PR is ready @danielfdsilva

danielfdsilva (approved):

Issue #468, opened 2023-02-23
Compare Map improvement


hanbyul-here:
(I will elaborate this later, writing super quick to capture the idea)
- separate compare layer from dataset configuration
- zoom/center

slesaad:
+ legends for both layers if they're different

danielfdsilva:
Considered done for both legends.
Other items will be addressed on different issues

PR #547, opened 2023-06-01
Put two legends when compare layer is not the same with the base layer

hanbyul-here:
related to #468 

Sometimes authors need to put two layers with different colormaps to compare. (ex. https://www.earthdata.nasa.gov/dashboard/discoveries/urban-heating the map at the top could be an interactive compare map.) This PR addresses the need. 

Currently, there is one discovery that uses different layers to compare from a config repo: /discoveries/tws-trends 

This PR stacks two legends. I am happy to iterate if there is a better way to handle multiple legends.
Screen Shot 2023-06-01 at 3 40 22 PM

### DIFF
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -34,7 +34,7 @@ import { S_FAILED, S_SUCCEEDED } from '$utils/status';
 
 import { SimpleMap } from '$components/common/mapbox/map';
 import Hug from '$styles/hug';
-import LayerLegendContainer from '$components/common/mapbox/layer-legend';
+import LayerLegendContainer, { LayerLegend } from '$components/common/mapbox/layer-legend';
 import MapMessage from '$components/common/mapbox/map-message';
 import { MapLoading } from '$components/common/loading-skeleton';
 import { HintedError } from '$utils/hinted-error';
@@ -402,14 +402,14 @@ function Scrollytelling(props) {
               }}
               classNames='reveal'
             >
-            <LayerLegendContainer
-              layers={[{
-                id:`base-${activeChapterLayer.layer.id}`,
-                description: activeChapterLayer.layer.description, 
-                title: activeChapterLayer.layer.name, 
-                ...activeChapterLayer.layer.legend}
-              ]}
-            />
+            <LayerLegendContainer key={activeChapterLayer.layer.name}>
+              <LayerLegend       
+                  id={`base-${activeChapterLayer.layer.id}`}
+                  description={activeChapterLayer.layer.description}
+                  title={activeChapterLayer.layer.name}
+                  {...activeChapterLayer.layer.legend}
+              />
+            </LayerLegendContainer>
             </CSSTransition>
           </SwitchTransition>
         )}
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -34,7 +34,8 @@ import { S_FAILED, S_SUCCEEDED } from '$utils/status';
 
 import { SimpleMap } from '$components/common/mapbox/map';
 import Hug from '$styles/hug';
-import LayerLegendContainer, {
+import {
+  LayerLegendContainer,
   LayerLegend
 } from '$components/common/mapbox/layer-legend';
 import MapMessage from '$components/common/mapbox/map-message';
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -67,6 +67,9 @@ const TheMap = styled.div<{ topOffset: number }>`
     top: ${topOffset}px;
     height: calc(100vh - ${topOffset}px);
   `}
+  .mapboxgl-canvas {
+    height: 100%;
+  }
 `;
 
 const TheChapters = styled(Hug)`
@@ -465,7 +468,11 @@ function Scrollytelling(props) {
             className='root'
             mapRef={mapRef}
             containerRef={mapContainer}
-            onLoad={() => setMapLoaded(true)}
+            onLoad={() => {
+              setMapLoaded(true);
+              // Fit the map to the container once  loaded.
+              mapRef.current?.resize();
+            }}
             mapOptions={mapOptions}
           />
         </Styles>
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -34,7 +34,7 @@ import { S_FAILED, S_SUCCEEDED } from '$utils/status';
 
 import { SimpleMap } from '$components/common/mapbox/map';
 import Hug from '$styles/hug';
-import LayerLegend from '$components/common/mapbox/layer-legend';
+import LayerLegendContainer from '$components/common/mapbox/layer-legend';
 import MapMessage from '$components/common/mapbox/map-message';
 import { MapLoading } from '$components/common/loading-skeleton';
 import { HintedError } from '$utils/hinted-error';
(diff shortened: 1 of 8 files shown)


ricardoduplos:
I would suggest to wrap multiple legend items in the same container. Stacked, these can be separated by each other using a 1px light gray bottom border. The container can be constrained to a max-height of 192px and be vertically scroll-able. An accordion behavior can be expected when interacting with the legend info: only one legend item can be expanded at a time.

hanbyul-here:
@ricardoduplos I was hoping to get design advice from you! Thank you!

parcel-resolver-veda/index.d.ts
```diff
@@ -62,13 +62,27 @@ declare module 'veda' {
...
   }
 
+  interface DatasetLayerCompareNoLegend extends DatasetLayerCompareBase {
+    legend?: never;
+    name?: never;
+    description?: never;
+  }
+      
+  interface DatasetLayerCompareWLegend extends DatasetLayerCompareBase {
+    legend: LayerLegendCategorical | LayerLegendGradient | undefined;
```
hanbyul-here, line 79: This passes ts check but does not real accurately reflect how typ should work. Because if the legend is undefined  - name and description should not exist. I tried several things but nothing seemed to work. Do you @danielfdsilva @nerik have an idea how I can type this better?

parcel-resolver-veda/index.d.ts
```diff
@@ -62,13 +62,27 @@ declare module 'veda' {
...
+    name: string;
+    description: string;
+  }
+
+  export type DatasetLayerCompareNormalized = DatasetLayerCompareNoLegend | DatasetLayerCompareWLegend
+
```
danielfdsilva, line 85: Perhaps it is not needed to create types for compare layers with and without legend.

After the compare layer is "normalized" it should have a legend since we need to display it.
```tsx
  export interface DatasetLayerCompareNormalized
    extends DatasetLayerCommonCompareProps {
    id: string;
    stacCol: string;
    type: DatasetLayerType;
    name: string;
    description: string;
    legend?: LayerLegendCategorical | LayerLegendGradient;
  }
```

This is ok when using a internal layer to compare (by providing a dataset and layer id), since the defined legend will be gotten from that layer.
When using a layer with a STAC definition (`DatasetLayerCompareSTAC`), we need to start requiring a legend, since it is not defined anywhere else.
```tsx
  export interface DatasetLayerCompareSTAC
    extends DatasetLayerCommonCompareProps {
    stacCol: string;
    type: DatasetLayerType;
    name: string;
    description: string;
    legend?: LayerLegendCategorical | LayerLegendGradient;
  }
```

hanbyul-here: ah I was thinking in a way that if comparelayer is fetching the layer from stac endpoint, there is no way to know about the legend but I guess we should ask in this case. Thank you for 💭 !

app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -402,12 +402,14 @@ function Scrollytelling(props) {
               }}
               classNames='reveal'
             >
-              <LayerLegend
-                id={`base-${activeChapterLayer.layer.id}`}
-                description={activeChapterLayer.layer.description}
-                title={activeChapterLayer.layer.name}
-                {...activeChapterLayer.layer.legend}
+            <LayerLegendContainer key={activeChapterLayer.layer.name}>
```
hanbyul-here, line 405: @danielfdsilva  Legend does not update inside of the scrollymap and I am not sure why. Do you have a better insight?

danielfdsilva: @hanbyul-here Added a commit to fix the animation. This was happening because the legend animates in and out with the help of `CSSTransition` which will wait for the transition to finish.
```
<CSSTransition
addEndListener={(node, done) => {
  if (!activeChapterLayer) return;
  node?.addEventListener('transitionend', done, false);
}}
>
```

When the wrapper `LayerLegendContainer` was added, the animation classes `reveal-*` were not moved to it form `LayerLegend`, causing the transition to never happen, leading to the `transitionend` event to never fire. Since the transition never finished the legend was not updated.

hanbyul-here: Thank you for handling!

app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -450,7 +468,11 @@ function Scrollytelling(props) {
             className='root'
             mapRef={mapRef}
             containerRef={mapContainer}
-            onLoad={() => setMapLoaded(true)}
+            onLoad={() => {
+              setMapLoaded(true);
+              // Fit the map to the container once  loaded.
+              mapRef.current?.resize();
```
hanbyul-here, line 474: I noticed the scrollymap doesn't fill up the screen height often (as in the screenshot). This bug seems to exist in the production too,  but somehow the change on this PR makes this bug much more obvious. I think it was because the map was not properly resized when the map was loaded - so I put this line.

Screen Shot 2023-06-06 at 12 12 30 PM

parcel-resolver-veda/index.d.ts
```diff
@@ -63,13 +66,38 @@ declare module 'veda' {
...
+    legend?: never;
+    name?: never;
+    description?: never;
+  }
+      
+  interface DatasetLayerCompareWLegend extends DatasetLayerCompareBase {
+    legend: LayerLegendCategorical | LayerLegendGradient | undefined;
+    name: string;
+    description: string;
+  }
```
danielfdsilva, line 86: Seems that these are no longer needed?

app/scripts/components/common/blocks/scrollytelling/index.tsx
danielfdsilva: This was a great catch!

danielfdsilva (changes requested):
If a dataset layer has the compare functionality available, both legends show even if the comparison is disabled.
The second legend should only be shown when the comparison is toggled on.
Example: https://deploy-preview-547--veda-ui.netlify.app/data-catalog/no2/explore?position=0%7C0%7C1.16&datetime=2016-01-01T00%3A00%3A00.000Z

app/scripts/components/common/mapbox/layer-legend.tsx
```diff
@@ -152,78 +168,84 @@ const LegendBody = styled(WidgetItemBodyInner)`
...
+    />
+  );
+}
+
+function LayerLegendContainer(props: LayerLegendContainerProps) {
+  return (
+    <LegendContainer>
+      <AccordionManager>
+        {props.children}
+        {/* {props.layers.map(e=> <LayerLegend key={e.id} {...e} />)} */}
```
danielfdsilva, line 242:
```suggestion
```

app/scripts/components/common/mapbox/layer-legend.tsx
```diff
@@ -152,78 +168,84 @@ const LegendBody = styled(WidgetItemBodyInner)`
...
+      </AccordionManager>
+    </LegendContainer>
   );
 }
 
-export default LayerLegend;
+export default LayerLegendContainer;
```
danielfdsilva, line 248: nitpick: now that we're exporting more than one component, it would be better to only use named exports

app/scripts/components/common/mapbox/layer-legend.tsx
```diff
@@ -76,9 +78,23 @@ const LayerLegendSelf = styled.div`
...
+
+const LayerLegendSelf = styled.div`
+  display: flex;
+  flex-flow: column nowrap;
+  width: 16rem;
+  border-bottom: 1px solid ${themeVal('color.base-100')};
 
   ${WidgetItemHeader} {
-    padding: ${variableGlsp(0.5, 0.75)};
+    padding: ${variableGlsp(0.25, 0.75)};
```
danielfdsilva, line 90:
```suggestion
    padding: ${variableGlsp(0.25, 0.5)};
```

app/scripts/components/common/mapbox/layer-legend.tsx
```diff
@@ -76,9 +78,23 @@ const LayerLegendSelf = styled.div`
...
+  }
+
+  &:only-child {
+    ${WidgetItemHeader} {
+      padding: ${variableGlsp(0.5, 0.75)};
```
danielfdsilva, line 95:
```suggestion
      padding: ${variableGlsp(0.5)};
```

hanbyul-here: Thank you for the review 🙇 I reflected all your commit suggestions in https://github.com/NASA-IMPACT/veda-ui/pull/547/commits/1ca17925b7dd2a641530663024e572ef505c0d30 !

danielfdsilva (approved):
:shipit:

PR #551, opened 2023-06-05
Runtime Discovery Editor

nerik:
Screenshot 2023-06-05 at 14 48 36


nerik:
This is the state of things:


https://github.com/NASA-IMPACT/veda-ui/assets/1583415/8be4d9c8-b3f4-4d54-b6ed-f01b4b76ea87

I'd be interested in hearing your thoughts about the approach @danielfdsilva @hanbyul-here ... Now to tackle error handling :trollface:

hanbyul-here:
It is amazing to see how you approached this problem. Such an exciting prototype with a relatively small amount of code 👯  A few thoughts after playing with it. (Maybe too detailed at this point?)

I tried to make errors in various ways. I have found three ways to make errors so far. 1. MDX syntax level (ex. `<>` in the middle of the paragraph): error message displays below the code mirror editor 2. Component level error (ex. giving a wrong dataset id to `Map` component): this displays hinted error, leaving other parts of the page intact 3. Block composition error (ex. giving the wrong combination to <Block /> such as <Block><Map></Block>): I thought this would be handled in the same way with 2, but it looks like this error breaks the whole page. As you said, a good next step will be error handling, including consolidating how these errors are handled. 

A draggable code editor is a brilliant idea to work around some problems quickly. But while trying to edit, I realized that dragging is also needed for editing (duh) 🤔  maybe some mechanisms to unlock/lock the code editors dragging? 

How do you picture the output of this editing process to be saved? Local files? GitHub?

danielfdsilva:
@nerik This is quite an interesting approach and it's nice to get the immediate preview!

I spent some time thinking about the error handling, but it is not straightforward to do with the current block scoped approach.
I think the error handling for the MDX editor should be different from the rest. The whole error handling should happen at a higher level so it can be cleared once the content changes. (An idea that you had already mentioned) Each block would render their own error, but they would be captured and stored higher up.

I tried to see if we could compare the props of a given block and clear the error if they were different, but since the mdx get evaluated on every render, the components are always different even if the content did not change, so this approach is a no go.

nerik:
@hanbyul-here @danielfdsilva This is ready for review. And merge if we all agree on https://github.com/NASA-IMPACT/veda-ui/issues/559#issuecomment-1598645007

This is new since we spoke last time:
- Content is stored on local storage to persist on page reload
- Added option to reset editor contents
- Block-level errors are properly handled: any error at this level will be caught by an ErrorBoundary, reported at the block site, and reset at each editor content update (and if the content is still faulty, the block will simply throw right away again)
- Global errors making the whole thing crash could in theory happen. 
   - The only manner I managed to get this was when changing from no content to any content (or the opposite), because of a rules of hooks error. I've added a specific hack for this case (no empty string is allowed on the editor), there is a more elegant fix for sure, but this does the trick.
   - Any putative global error will lead to a fallback screens which offers to do a full page refresh with or without the currently saved content

app/scripts/components/sandbox/mdxEditor/MDX.tsx
```diff
@@ -0,0 +1,264 @@
...
+  );
+
+  const errorHumanReadable = useMemo(() => {
+    if (!error) return null;
+    const { line, message } = JSON.parse(JSON.stringify(error));
+    return `At line ${line - 1}: ${message}`;
+  }, [error]);
+
+  return (
+    <ErrorBoundaryWithCRAReset FallbackComponent={GlobalError}>
```
hanbyul-here, line 140: What kind of errors trigger `GlobalError` component to show up? I tried to trigger it, but I couldn't find the way.

app/scripts/components/sandbox/mdxEditor/MDX.tsx
```diff
@@ -0,0 +1,264 @@
...
+    />
+  );
+};
+
+export const MDXBlockWithError = (props) => {
+  const result = useContext(MDXContext);
+  return (
+    <ErrorBoundaryWithCRAReset
+      FallbackComponent={MDXBlockError}
+      resetKeys={[result]}
```
hanbyul-here, line 257: ❗ resetKeys ❗

package.json
```diff
@@ -152,10 +159,13 @@
...
     "react-transition-group": "^4.4.2",
     "recharts": "2.1.12",
+    "rodemirror": "^2.0.0",
     "scrollama": "^3.1.0",
     "shpjs": "^4.0.4",
     "styled-components": "^5.3.3",
-    "use-deep-compare": "^1.1.0"
+    "use-deep-compare": "^1.1.0",
+    "use-local-storage": "^3.0.0",
+    "vfile": "^5.3.7"
```
hanbyul-here, line 168: nitpick: do we need this?

app/scripts/components/sandbox/mdxEditor/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  Image,
+  Map: LazyMap,
+  ScrollytellingBlock: LazyScrollyTelling,
+  Chart: LazyChart,
+  CompareImage: LazyCompareImage,
+  NotebookConnectCallout: NotebookConnectCalloutBlock,
+  Link: SmartLink
+};
+
+export const MDX_LOCAL_STORAGE_KEY = "MDX_EDITOR";
```
hanbyul-here, line 32: nitpick: It might be a good idea to start prefixing these local storage keys with `VEDA` or something?

nerik:
@hanbyul-here 

> We can consolidate the diverged approaches to achieve a similar goal inside the applications. ex. Analysis page and MDX editor use both local storage but in different ways. 

Interesting, we might even be able to have a pseudo "user parameters" page where a user could see saved AoIs, analysis params, discovery draft, etc... With the obvious caveat that this is tied to the current browser, which must be clearky communicated.

 > But I think it is more important to test this idea fast. so if you want to consider the integration with the existing architecture later, I am down with it.

Cool! Let's also see what comes out of https://github.com/NASA-IMPACT/veda-ui/issues/561

app/scripts/components/sandbox/mdxEditor/MDX.tsx
nerik: None _that I am aware of_ ... But considering there's an `eval` happening within this ErrorBoundary, anything could happen (as in [zombo.com](https://zombo.com/)).
You can however manually trigger the error by commenting [this line](https://github.com/NASA-IMPACT/veda-ui/pull/551/files#diff-36308502277c3029ba212de372c8d8cb28d5f03c35c7c20d936c1d45aab9226fR174) out, and clear the editor's content.

app/scripts/components/sandbox/mdxEditor/index.tsx
nerik: Maybe you're right... Maybe it should be part of a broader effort to centralize local storage handling within the app? I created an issue for now: https://github.com/NASA-IMPACT/veda-ui/issues/564

app/scripts/components/sandbox/mdxEditor/MDX.tsx
nerik: Anything wrong here?

danielfdsilva (commented):
nitpick: all files and folders on veda follow the kebab-case syntax. Would be good to maintain consistency

app/scripts/components/sandbox/mdxEditor/MDX.tsx
danielfdsilva: What is this `resetKeys`? I haven't seen it being used anywhere

danielfdsilva (changes requested):
The CRA overlay is still popping up. I'd suggest we change the selector to the following one so it is more generic. It is very unlikely that this would interfere with any other piece
```
body > iframe:last-of-type {
```

app/scripts/components/sandbox/mdxEditor/MDX.tsx
nerik: @hanbyul-here @danielfdsilva, indeed it's part of `react-error-boundary`'s API. There's not much documentation, but this [blog post](https://kentcdodds.com/blog/use-react-error-boundary-to-handle-errors-in-react) explains it, kinda works like a memoize.

nerik:
> The CRA overlay is still popping up. I'd suggest we change the selector to the following one so it is more generic. It is very unlikely that this would interfere with any other piece

@danielfdsilva Fixed, although I was not able to replicate this.

app/scripts/components/sandbox/mdxEditor/MDX.tsx
danielfdsilva: ahh, Thank you! That's an interesting feature.
I had looked at the docs, but didn't find it. It's a bit buried. 🫣

hanbyul-here (approved):
Not a blocker, just a note so we know the quirks.
- the components that use the internal files (chart, image) don't work in the editor. 
- the editor gets 'scrolled away' with scrolly map component. 

but an exciting change! I look forward to getting some feedbacks.

danielfdsilva (approved):

PR #553, opened 2023-06-06
Add ability for users to add custom pages

danielfdsilva:
I tried to repurpose the process we already have in place to override content, to allow the users to add custom pages.

So they'd use the `pageOverrides` option on `veda.config.js` and include their new routes.
They'd be differentiated from the other overrides by starting with a `/`. Then the behavior is the same: point to an MDX file with the content for the page, and frontmatter properties for options (like the `menu` label)

For example:
```js
  pageOverrides: {
    '/disclaimer': './custom-pages/disclaimer.mdx',
    '/partnership': './custom-pages/partnership.mdx'
  }
```

@nerik @hanbyul-here What do you think about this approach?

The main navigation menu is already a bit crowded and adding a couple of pages already causes it to overflow.
Screenshot 2023-06-06 at 16 50 56

On small screens all the menu links are inside a drawer so this is not a problem, but if we use the drawer at a larger breakpoint then the menu will be almost never visible.

I tried to put these additional pages inside a dropdown menu. This works and scales indefinitely, but I'm not sure how happy people will be to have some pages buried.
image

Maybe an alternative could be to have all the pages of the right navigation permanently in a drawer?
@faustoperez What's your take on this?
@j08lue Would having the custom pages hidden be something acceptable?

### TODO
- [x] Add documentation
- [x] Prevent users from overriding existing pages like `/data-catalog`


mock/veda.config.js
```diff
@@ -6,6 +6,10 @@ module.exports = {
 
   pageOverrides: {
     aboutContent: './about.mdx',
-    'sandbox-override': './sandbox-override.mdx'
+    'sandbox-override': './sandbox-override.mdx',
+
+    '/disclaimer': './custom-pages/disclaimer.mdx',
```
hanbyul-here, line 11: I imagined in a way that we have one folder (like `custom-pages`), and whichever mdx file is created in this folder auto-magically gets published, so users don't even need to touch veda.config.js. so developers don't need to be pulled in for support. (but honestly, this will happen rarely, so I think the current approach is fine.)

danielfdsilva:
⚠️ Rebased to include the latest main changes

mock/veda.config.js
nerik: +1 to this ☝️

nerik: I'd also argue that the slash prefix is a hidden behaviour (until you read the docs)

nerik (approved):

faustoperez:
Hi, @j08lue @danielfdsilva sorry this slipped my attention. Here's my 2 cents:

I would say that, for the main site navigation we only need three elements:

1. Data Catalog
2. Analysis
3. Discoveries

We can safely remove the Welcome tab (in the context that we don't have thematic areas homepages anymore), and use the logo to go to the homepage.

For the secondary menu, I see 3 strategies that could work:

1. If we want to keep all the elements there, change the styling to differentiate the group from the main navigation and lower the font size to gain some extra pixels.
2. Potentially, group them all under the three vertical dots at smaller screen sizes
3. Put them all in the footer, as this is secondary information

I would vote for option 3 here. Hope this helps.

PR #554, opened 2023-06-07
Update React to 18

### LINKED ISSUE #545: Migrate app to react 18
Investigate the feasibility of upgrading the react version to 18.

There may be some libraries that are not compatible with the upgrade. The first pass is to understand if this is the case, and log which ones are problematic.
Following this we can gauge the LOE for the upgrade.

hanbyul-here:
Close #545 

React 18 upgrade guide:  https://react.dev/blog/2022/03/08/react-18-upgrade-guide
- React announcement says that  ReactDOM.render has been deprecated. Using it will warn and run your app in React 17 mode. There are a few problems with concurrent client rendering API (`createRoot`) 
  - NavWrapper throws an error `Element #header-wrapper was not found.` . It looks like rendering order got changed with the new API and doesn't guarantee nav wrapper dom element on the page. Maybe this is a good time to rewrite so we don't have to depend on querySelector? 🤔 _ I put a work-around.
  - `use-mapbox-control` depends on the deprecated methods (`render`, `unmountComponentAtNode`)

- They dropped the support for IE. It might be worth checking at least GA to see if there are any IE users.


danielfdsilva:
@hanbyul-here Added a couple of commits for ts typings and updates the mapbox control to use the new api.

I tried to wrap the app with the `StrictMode` but get an error with `UNSAFE_componentWillMount`. It seems to come from recharts. Maybe it is worth double checking.

yarn.lock
```diff
@@ -1213,33 +1213,33 @@
 
 "@devseed-ui/accordion@4.1.0":
   version "4.1.0"
-  resolved "http://3.220.121.64:4873/@devseed-ui%2faccordion/-/accordion-4.1.0.tgz#622483e69e553a4f1e8d4a8c44f61ad4cfa1395b"
+  resolved "http://verdaccio.ds.io:4873/@devseed-ui%2faccordion/-/accordion-4.1.0.tgz#622483e69e553a4f1e8d4a8c44f61ad4cfa1395b"
```
hanbyul-here, line 1216: @danielfdsilva What is Verdaccio? (I think you told me about this when I first started, but obviously, I forgot about it 😓 )

danielfdsilva: Verdaccio is our private npm registry. We were mostly using it when we had beta versions of the ui library, and even though we currently don't have any private libraries, requests still go through it

danielfdsilva (approved):
I tested this a bit more and everything seems to work fine.
Perhaps a final pass by @nerik?

hanbyul-here:
Yeah, I was just thinking about how to proceed with this pr. I think we can proceed. Considering Erik is at the conference this week... Do we still have any dedicated QA person to check everything in? It will be nice to get one more go-through, but this should not be the blocker.

danielfdsilva:
@hanbyul-here ah right. Totally forgot about the conference 😅 . Perhaps @karitotp can have an additional look?

@karitotp nothing should have changed. It is just a matter to double check that nothing is breaking because of the react upgrade - basically going though the pages. Do you have some time for this?

karitotp:
Hey @danielfdsilva, I took a look at all the pages and everything is working fine.

Issue #558, opened 2023-06-19
Respond to user feedback: Legend needs more ticks


j08lue:
@siddharth0248 reports that legends sometimes are hard to read.

My understanding is that the challenge is to see how colors map to values somewhere in the middle of the range, in particular with color maps with many colors, and when it is hard to quickly estimate what the mid-point of the color range is. E.g., what is the mid-point of -17 and 238?

This is [a legend](https://ghg-demo.netlify.app/data-catalog/epa-emissions-annual-tests/explore?position=-95%7C39.6655%7C4.12&datetime=2012-01-01T00%3A00%3A00.000Z&layer=4b-manure-management) I would find hard to read:

![Image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/c149ce8b-d2bd-47c9-b8d9-41497b2bc637)

Maybe a tool tip would help, like we have for categorical color maps? 

![Image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/7b898903-23a3-40ad-9f3f-a550d9586e2e)

Or more labelled ticks?

This ticket is really just to give an initial response to the request and decide what should be done - more research, design suggestions...

## Acceptance criteria

- [ ] Decided how to respond to this request
- [ ] Ticketed next steps

j08lue:
> when the user hovers the bar, a popover shows the interpolated value under the cursor

Sounds great, looking forward to seeing this.

Hope users will discover that this is possible - we'll see. Else we can make that popover flash once, at the midpoint, when the component loads. Or something...

PR #562, opened 2023-06-19
Version 1.0.0

danielfdsilva:
Features present in main, that we're freezing for the production deploy.

From now on, since we have multiple instances depending on veda-ui, we need to start versioning it to make tracking the version easier.
Process detailed in [almanac](https://almanac.io/handbook/veda-product-development-documentation-I3UwKo/deployment-L1nqhydmGejnwG5b5IdVqvNpEb4nigHb).

---

# Changelog v1.0.0

## 🎉 Features
- Expand veda ui configurability by instances
- Ensure map legends show up for both layer when comparing (https://github.com/NASA-IMPACT/veda-ui/pull/547)
- Add page footer and header brand component override
- Add custom pages functionality
- Add Link component for long form content
- Add meta images configurability
- Include documentation for new features

## 🚀 Improvements
- Improve map legend display (https://github.com/NASA-IMPACT/veda-ui/issues/558)

## 🐛 Fixes
- Fix Image Block alignment class
- Fix map layer not loading on srollytelling (https://github.com/NASA-IMPACT/veda-ui/issues/542)

---

## Migration guide

Because of how meta images are now handled, instances must add the following configuration to the `package.json` file:

```
"parcelIgnore": [
  ".*/meta/"
]
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "veda-ui",
   "description": "Dashboard",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "author": {
     "name": "Development Seed",
     "url": "https://developmentseed.org/"


j08lue (approved):
Great to see this project maturing with version 1.0 release. Best practice at work - being so explicit about versions and features, which will help manage different needs across instances / users. 🚀

hanbyul-here (approved):
wah wah v1 🥳 
If we follow Semver rule, once we are on v1, we should be very intentional about the versioning. Considering 'incompatibility' is the reason to bump the major number, what does it mean for Veda UI to be incompatible? Is this when veda-config cannot run veda-ui? (like meta property change?)

danielfdsilva:
@hanbyul-here yes. I'd say that's the idea. If we make changes that require the instances to migrate something, then it will be "incompatible". It is very likely that we'll have a lot of then, so this will reach high number versions 😅, but that shouldn't be a problem

PR #563, opened 2023-06-21
Allow for legend to show more values

danielfdsilva:
image

This PR changes the behavior of the legend to allow the users to understand intermediary values when hovering.

This feature only works when the legend's `min` and `max` can be converted to number. This means that for legends where the value is textual, like `Less/More` or `< 3 / > 3`, this feature won't work, since it is not possible to interpolate values in this case.

This `min/max` display was also improved. These values are now shown as scientific notation whenever the value is over 1000 or below 0.001, and the unit is displayed on the left.

### DIFF
--- a/app/scripts/components/common/mapbox/layer-legend.tsx
+++ b/app/scripts/components/common/mapbox/layer-legend.tsx
@@ -54,8 +54,8 @@ const printLegendVal = (val: string | number) => {
   const number = Number(val);
   if (isNaN(number)) return val;
 
-  if (Math.abs(number) < 1000) {
-    return formatThousands(number);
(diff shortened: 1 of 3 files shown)


j08lue (approved):
Both of these changes - scientific notation and the tooltip - make for a great increase in the value of the legend for users to translate colors into numerical values.

My only concern is whether users will discover the mouseover feature, since it may not be obvious that it exists. Feedback from use in the wild will show - let us get this out asap.

ricardoduplos (approved):

hanbyul-here:
I was checking the datasets on veda-config. are we planning to rewrite min, max values of the datasets? It looks like a lot of the legend min max value is textual, not a pure number value.

Issue #568, opened 2023-06-26
CH4 plumes emission dataset still shows up for the dates without any data


hanbyul-here:
[https://deploy-preview-6--ghg-demo.netlify.app/data-catalog/jpl-ch4-plume-emissions/explore?posi[…]926%7C14.57&datetime=2022-08-11T00%3A00%3A00.000Z](https://deploy-preview-6--ghg-demo.netlify.app/data-catalog/jpl-ch4-plume-emissions/explore?position=41.1468%7C34.8926%7C14.57&datetime=2022-08-11T00%3A00%3A00.000Z) 
CH4 plumes dataset doesn't have any data for 12,13,14 Aug, but the dashboard still shows the data from 8/11 even when the selected date is changed. I am unsure if this is the data problem or the dashboard problem. This issue might be moved to UI repo later after the investigation

danielfdsilva:
@hanbyul-here This is interesting and it is a fix we can definitely make, however I'd argue that if a dataset doesn't have data for a given day, then it shouldn't be "periodic". Having a date available and show no data on the screen is misleading to the user.

I'd say, that this also needs fixing on the data side

PR #569, opened 2023-06-27
Remove layer from map if search returned no data

danielfdsilva:
Addresses the frontend portion of #568 - data should still be verified.


hanbyul-here (approved):
This data is probably not periodic, but this change is good to have. Thanks for working on it!

Issue #570, opened 2023-06-27
Analysis doesn't work for multi-asset stac records (absent `cog_default`)


slesaad:
# Description
The current analysis framework works under the assumption that there's a `cog_default` asset for all collections.

We want to move away from the `cog_default` model, to be able to support multiple assets under the same item. [ADR here](https://nasa-impact.github.io/veda-architecture/adr/20230607-stac-catalog-organization/).
(I mentioned in the ADR that Dashboard supports the change as is, but today, realized that for Analysis, that's not the case. TODO: Update ADR)

For multi-asset items, we'd probably want to either show the assets within the returned collection as checkboxes, maybe with collection level grouping, when the user clicks the submit button OR when a collection checkbox is selected, automatically add all the assets in the collection to the analysis.

Also, while digging around, I noticed that the API request made for analysis (at least, haven't checked with others) uses `layer.id` instead of `layer.stacCol`, which will fail to return the collection if `layer.id` and `layer.stacCol` aren't the same. Is the expectation that they'll always be the same?

# Acceptance Criteria
- [ ] Analysis in the dashboard supports multi-asset items

danielfdsilva:
@slesaad For the time being the dashboard is working under the assumption that each stac item has 1 asset (configurable), and that is used both on the explore and on the analysis (with the fix on #571 )

danielfdsilva:
A current work around for this issue is to create multiple "dataset layers" with a different `assets` value.
Closing for the time being. May need to be revisited with #493

slesaad:
Noticed that only the first layer of a dataset would show up as selectable on the analysis page. Created a PR to fix that https://github.com/NASA-IMPACT/veda-ui/pull/582

@danielfdsilva lemme know

ifsimicoded:
closing as I see a merged PR [#582](https://github.com/NASA-IMPACT/veda-ui/pull/582)

PR #571, opened 2023-06-28
Fix analysis and use configured "assets" value

danielfdsilva:
(from #570 ) The analysis was not working properly. It was requesting datasets from stac using the `layerId` instead of `stacCol`. We haven't noticed it before because most of the datasets have the same value (coincidentally). This PR also addresses this bug while making the `assets` value configurable for the analysis.

@slesaad Great catch! I also noticed that the `/statistics` response returns data under a `b1` key, and we're expecting a `1` key (This is so in VEDA). Is this possible to standardize?

image

### DIFF
--- a/app/scripts/components/analysis/define/use-stac-search.ts
+++ b/app/scripts/components/analysis/define/use-stac-search.ts
@@ -39,8 +39,8 @@ export function useStacSearch({ start, end, aoi }: UseStacSearchProps) {
       try {
         const url = `${process.env.API_STAC_ENDPOINT}/search`;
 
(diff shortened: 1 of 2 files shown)


hanbyul-here (approved):
Ah I guess we missed it when separating layer id from stac id. + the analysis page graph still doesn't show up it sounds like this is from the data problem. (statistics endpoint returning b1, not 1)

danielfdsilva:
Tracking data problem in https://github.com/NASA-IMPACT/veda-ui/issues/572

Issue #574, opened 2023-07-05
Rename Discoveries as Data Stories


j08lue:
Following the decision in #550, we will prepare renaming the "Discoveries" to "Data Stories". 

This change will be done at `veda-ui` level and applied to all instances.

After a experimental period, if stakeholders are ok with it, we'll make code changes to get rid of "discoveries" references (urls, file names, etc.)

## Acceptance criteria

- [x] Change the name superficially for VEDA GHG Center
- [x] Show to stakeholders
- [ ] Implement fully in code and URLs (`/stories`)
- [ ] Deploy to VEDA NASA and VEDA GHG

j08lue:
I got some good feedback from EIS regarding renaming "Discoveries" to "Data Stories" on their instance. @dfelikson argued against it that EIS wants to tell about science with these stories ("Science Stories"), while the GHG Center to our current understanding wants to use this feature of the Dasboard to tell about features in the GHG data ("Data Stories").

Could a solution be that we refer to this feature as `stories` everywhere in code and URLs but make the title / display name configurable, such that EIS can call them "Science Stories" or keep "Discoveries" and GHG Center can go with "Data Stories" or something else they may want later?

@danielfdsilva and @ricardoduplos  - how do you see this?

slesaad:
@danielfdsilva any chance this can be done before the friday ghg demo?

danielfdsilva:
From https://github.com/NASA-IMPACT/veda-ui/pull/577#pullrequestreview-1524007033

> The title "Data Stories" is
> * Right for GHG Center as far as we know
> * Ok, although not ideal for NASA/EIS
> 
> Next steps after merging this (AFAIK):
> 1. Deploy for GHG
> 2. Deploy for NASA/EIS
> 3. Consider making the title configurable (see https://github.com/NASA-IMPACT/veda-ui/issues/574#issuecomment-1627197508)
> 4. Implement renaming in code/urls to use `stories`, if possible.

PR #577, opened 2023-07-05
Change Discovery related strings to Data Stories

danielfdsilva:
@j08lue Could the preview url be used to show stakeholders?

### DIFF
--- a/app/scripts/components/common/featured-slider-section.tsx
+++ b/app/scripts/components/common/featured-slider-section.tsx
@@ -136,7 +136,7 @@ function FeaturedSliderSection(props: FeaturedSliderSectionProps) {
 export function FeaturedDiscoveries() {
   return (
     <FeaturedSliderSection
-      title='Featured Discoveries'
+      title='Featured Data Stories'
       featuredItems={allFeaturedDiscoveries}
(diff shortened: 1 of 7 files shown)


j08lue:
Great, yes, thank you!

Quick question: If we implemented this also in the URLs (and other naming behind the scenes) - could we keep a `/discoveries` alias for a while, so existing links elsewhere do not break?

danielfdsilva:
@j08lue We would implement redirection for those paths, like we did for the thematic areas

j08lue:
@danielfdsilva, we got confirmation that we should move ahead with this change. Can we deploy it to the GHG Center instance before July 14, still - perhaps just in this cosmetic version - and then work on the more thorough renaming?

j08lue (approved):
This is good for merging.

The title "Data Stories" is
* Right for GHG Center as far as we know
* Ok, although not ideal for NASA/EIS

Next steps after merging this:
1. Deploy for GHG
2. Deploy for NASA/EIS
3. Consider making the title configurable (see https://github.com/NASA-IMPACT/veda-ui/issues/574#issuecomment-1627197508)
4. Implement renaming in code/urls to use `stories`, if possible.

PR #582, opened 2023-07-13
Fix: Analysis only shows the first layer for a dataset

slesaad:
# Description
For multi-asset items, the Analysis page would only show the first layer because the available datasets were being filtered by unique stac collection id (`stacCol`).

This removes the filter, so all the assets show up.


danielfdsilva (approved):
great catch

Issue #573, opened 2023-06-30
Allow for customization of data catalog filters


danielfdsilva:
From discussion on: https://github.com/NASA-IMPACT/veda-architecture/issues/273

Each instance should be able to define what filters are available in the data catalog page based on dataset properties.
Filtering will be done on a direct property matching basis, without complex logic.

### Automatically inferred
This option would have the users add values under a specific key in the dataset frontmatter (eg. `taxonomy`). The app would then gather all unique values and build a filtering system.  

```yml
taxonomy:
  Sector:
    - Agriculture, Forestry and Land Use
    - Energy
  Producer:
    - NIST
  Gas Emission:
  -   COx
```

**pros & cons** 
- 💚 No need to have filters defined elsewhere
- 💚 Easily readable
- 💔 Prone to error, given that users have to spell the filter name exactly

### Manual definition
This option requires the users to specify the the filter keys and labels in a separate file, besides adding the values to a dataset frontmatter.

```js
filters = {
  sector: {
    label: 'Sector',
    key: `sector`,
    options : [] // Could be implemented to offer option validation
  },
  gas: {
    label: 'Gas Emission',
    key: `gas-emission`,
  }
}
```
On the dataset
```yml
id: dataset id
name: Some name
sector: Energy
gas-emission: COx
```

**pros & cons** 
- 💚 No need to rely on a specific dataset key to hold all values
- 💚 Allows for values to be validated if array of options is provided
- 💔 Requires maintenance of taxonomy source
- 💔💚 Requires users to check file to know the keys, but it is less error prone

## Design

The current UI does not accommodate a growing list of filters and has a pretty low hard limit.  
We need to find a way to ensure a higher number of filters is supported.

<img width="1350" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/1090606/d7d27ef1-7f4c-4345-9817-960dccb3eabd">

## Tasks
_besides technical implementation_

- [ ] Comment on approach @danielfdsilva @hanbyul-here @nerik (feel free to provide other options)
- [ ] Figure out way to accomodate more filters @ricardoduplos @faustoperez

hanbyul-here:
More related to the original issue: It could be from my ignorance, but can we add this taxonomy metadata to the searchable metadata so we don't need an unexpected number of dropdowns on the page? My preference for searching is always just a plain text search + no limit on the level of taxonomies sounds like it can become chaotic fast. 
[...]

PR #583, opened 2023-07-14
Make content filters dynamic

danielfdsilva:
This PR makes the filters on the data catalog and data stories pages dynamic, using the Automatically inferred approach described in #573.

The topics and source properties were moved to `taxonomy`, to be able to have them as filters. Currently these are also used to display information on the cards.

The behavior of the filters is limited to selecting one option from a list. It is not possible to create boolean filters (eg: dataset has property) or even more complex filters.

Given that the filters need to scale they now stack, although there should be some common sense when creating new taxonomies as to not clutter the interface.

### TODO:
- [ ] Review UI @ricardoduplos 
- [x] Update documentation for taxonomies @danielfdsilva 
- [x] Prepare content migration guide prior to any release ⚠️


mock/datasets/nighttime-lights.data.mdx
```diff
@@ -12,6 +12,16 @@ media:
 thematics:
   - covid-19
   - agriculture
```
j08lue, line 14: `thematics` can probably be removed now?

j08lue (approved):
To me, this solution fulfils the two goals we had:
1. Presenting the complete taxonomy (as drop-downs) to the user for a structured overview of the catalog
2. Enable users to quickly find the data they are looking for, by using one of several possible filters, or a combination.

Just for the record - we discussed previously that a possible improvement would be that only combinations of filters are shown that actually exist, such that users do not end up with no results and instead can use the filters to refine their search step-wise. We can discuss that as a future improvement - or table it until users perhaps request this.

mock/datasets/nighttime-lights.data.mdx
danielfdsilva: yes. this will need to move under `taxonomy.Topics`

mock/datasets/nighttime-lights.data.mdx
```diff
@@ -12,6 +12,16 @@ media:
...
+taxonomy:
+  Topics:
```
hanbyul-here, line 16: This makes me wonder if users will understand what they type here is what they will see. (Since this is the only capitalized attribute in mdx, it looks wrong somehow in mdx context.) But I also don't really see a better solution since there will be various requirements for capitalization 🤷

hanbyul-here (commented):
Thanks for working on it. 

Now I realize that the validation for taxonomies will be difficult since we won't know if it is a typo or just a similar text. But this approach still makes sense under the current Veda condition I think. At least editor can see what they did wrong through the UI quickly and can act on it. 

A small incoherent behavior I noticed is when Topic is selected, the URL shows not-encoded query parameters while showing the encoded URL for other taxonomy options. 

Screen Shot 2023-07-20 at 11 39 41 AM


+(This function will definitely benefit greatly from some autocomplete UI.)

mock/datasets/nighttime-lights.data.mdx
j08lue: Agree capitalized keys (perhaps also with special characters and spaces) could perhaps be avoided, to keep YAML keys clean?

Is this better?

```yaml
taxonomy:
  - title: Topic
    values:
      - Covid 19
      - Agriculture
    - title: Sector
      values:
        - Electricity
```

danielfdsilva: yeah, this is a very valid concern. In this way it perhaps increases clarity with the cost of verbosity (in an already complex frontmatter). Not sure which one is preferable.

@hanbyul-here The filters get encoded as json deliberately. It is not possible to create individual keys for each taxonomy, since we don't know how many there are. Each key is created with a hook, and dynamically creating them would mean to have a loop creating the hooks, which is not ok. Therefore, all taxonomy options are stored as an object under `taxonomy` which get encoded as json. We could write a custom `hydrate|dehydrate` function to make it prettier, but I'm wondering if it is worth it. Thoughts?

hanbyul-here: Oh, my point was more about
When you click 'Ccovid 19' from the card, the URL changes to: http://localhost:9000/discoveries?taxonomy={%22Topics%22:%22ccovid-19%22}
but when you select 'Ccovid 19' from the dropdown, the URL changes to: http://localhost:9000/discoveries?taxonomy=%7B%22Topics%22%3A%22ccovid-19%22%7D

I think it is because JSON.stringify doesn't encode `{}`? I don't think it is a blocker but thought you might have missed it.

danielfdsilva: @hanbyul-here I had completely misunderstood what you meant. 😅 Good catch. Fixed!

danielfdsilva:
Fixed the encoding error, added a metadata section on content pages, and updated documentation.
@ricardoduplos can you have a look to see if it looks good?

Content migration guide to add to the release:

-----

## Content migration

### Taxonomy

With the removal of `thematics` and `sources` in favor of `taxonomy`, the values under it must be migrated to `taxonomy.Topics` and `taxonomy.Source` respectively.  
The values must also be converted from kebab-case to a human readable format (Title Case)

Example:

```yml
name: Dataset Name
thematics:
  - covid-19
  - agriculture
sources:
  - dev-seed
```
becomes
```yml
name: Dataset Name
taxonomy:
  Topics:
    - Covid 19
    - Agriculture
  Source:
    - Development Seed
```

### Taxonomy Index

The taxonomy index file is no longer required and can be removed.  
This file should be deleted and its reference in the `veda.config.js` file should be removed.

mock/datasets/nighttime-lights.data.mdx
hanbyul-here: Yay thanks for handling. I like @j08lue 's idea about making the value explicit. It is a valid concern that it will be verbose, but I feel like that is more coherent with other attributes of metadata too.

danielfdsilva:
@hanbyul-here @j08lue I updated the implementation and documentation to use the more verbose array format.

@karitotp Do you have some time to do a test of this dynamic filter feature and ensure everything is working properly?

karitotp:
@danielfdsilva, the dynamic filters are working well on the Data Catalog and Data Stories.

Just to make sure, the filter by search field on both pages, just works when the input text is in the title or the description of the cards as the placeholder says, right? I saw that in the previous versions, it applied to the pill names too, but currently, it does not, and it looks like this.

Selection_961

hanbyul-here:
As @karitotp mentioned, the recent commits seem to change how text filter works?

danielfdsilva:
@karitotp @hanbyul-here That was a bug and was not supposed to happen. It's fixed now

PR #595, opened 2023-07-26
Rename discoveries to stories

danielfdsilva:
This PR completely renames Discoveries to Stories following the discussion on issue https://github.com/NASA-IMPACT/veda-ui/issues/574

Every single code reference to discovery (or a variation of it) has been changed to story. This includes, variables, strings, component names, files, etc.
The files now go into a folder called `stories` instead of `discoveries`, and their name also changed to `<name>.stories.mdx`. All the documentation was also updated to reflect this change.

**This is a vey big change which will require people to adapt to the new nomenclature.** However, it is a change that will help bring more clarity to the concept of stories and how they are used in the different instances.

The url for the stories is now `/stories/<id>` instead of `/discoveries/<id>`. The old urls will still work, but they will redirect to the new ones.
Even though the url can't be configured, the name given to a story and how it shows up in the interface can be configured through the `veda.config.js` file under a `strings` key.  
Because it is used in different ways, the configuration must contemplate a value for the singular, plural and zero form of the word. For example:
```js
{
  stories: {
    singular: 'Story',
    plural: 'Stories',
    zero: 'Stories'
  }
}
```
@hanbyul-here @nerik @ricardoduplos  what do you think of this approach to the configuration? Is it too much?

@karitotp  Can you do an additional check to see that we didn't miss any renaming and that everything still works as expected?

----

## Next steps

As soon as we decide to move forward with this, we need to update all the files and mentions in the `veda-config` and `veda-config-ghg` repos (which includes some docs files in these repos). We can make this change to get things started, but this change will have to be carefully communicated so that content creators are aware that going forward things wil have different name.

@j08lue  is this communication something you can help facilitate? Do you forsee big issues with this change?

----
(for the release notes)

### Discoveries renamed to Stories

This version includes a content type change from `discoveries` to `stories`. This includes a url change from `/discoveries/<id>` to `/stories/<id>`, but redirects are in place to ensure that the old urls still work.

This change will require each instance to update their content to the new format. The following steps should be taken to migrate the content:

1. In the `veda.config.js` file, make the following change
   ```diff
   +stories: './stories/*.stories.mdx',
   -discoveries: './discoveries/*.discoveries.mdx',
   ```
2. It is now possible to customize the name of the stories in the interface. If you want a value different than "Stories" check the [configuration](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/CONFIGURATION.md) section of the documentation to see how to do it.
3. Rename the `discoveries` folder to `stories`
4. Rename each discovery file from `<name>.discoveries.mdx` to `<name>.stories.mdx`
5. Content that has a [related content frontmatter](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/CONTENT.md) key should be updated to use `story` type instead of `discovery`


j08lue (approved):
Looks great and seems like you thought of everything, also in the migration notes, @danielfdsilva. 👌 Will these notes live somewhere in the release notes or so, where we can reference them? Happy to lead the communication with the instance owners.

I suggest we merge this and then open a ticket in the `veda-config` and `veda-config-ghg` repos for them to migrate and update the `veda-ui` library version and then we can help the owners complete that process.

When we migrate the content for each instance, we may want to update the screen shots on the NASA/EIS VEDA homepage, too:
 
image

danielfdsilva:
@j08lue Yes, the notes will live with the release changelog. Anyone looking at the release notes will immediately know what to do to upgrade to this version.

To speed up the process I propose that we're the ones to do this migration right know. And great catch about the screenshots.

karitotp (approved):

hanbyul-here:
https://deploy-preview-595--veda-ui.netlify.app/discoveries/life-of-water/ This page doesn't seem to get redirected to https://deploy-preview-595--veda-ui.netlify.app/stories/life-of-water/ ?

danielfdsilva:
Great catch @hanbyul-here . Fixed

app/scripts/components/stories/hub/index.tsx
```diff
@@ -295,12 +295,12 @@ function DiscoveriesHub() {
           </CardList>
         ) : (
           <EmptyHub>
-            There are no discoveries to show with the selected filters.
+            There are no stories to show with the selected filters.
```
nerik, line 298:
```suggestion
            There are no {getString('stories').zero} to show with the selected filters.
```

docs/content/CONFIGURATION.md
```diff
@@ -15,18 +15,33 @@ The `.env` file contains a list of all available variables and comments explaini
...
 ```
 
 ### Page overrides
 To adapt the Veda dashboard to the individual needs of you instance, some content/component overrides are provided. These overrides allow you to alter certain parts of the application, or inject code without having to fork the UI part of veda.
 
 See [PAGE_OVERRIDES](./PAGE_OVERRIDES.md) for a full list of elements to customize.
 
+### Strings
+
+The `strings` object allows you to customize the nomenclature used in some parts of the application so that it better reflects your use case. Since these values will need to be used in different contexts, a plural, singular and zero form of each term is required.
```
nerik, line 32: Should we always enforce that? My hunch is that if we end up using the strings feature a lot, enforcing the zero/plural/singular rule is going to be cumbersome and not needed in the majority of cases.

nerik (changes requested):
Great jobs on the `strings` feature. This will come in handy 👍

docs/content/CONFIGURATION.md
danielfdsilva: I agree that this adds some overhead. Would your suggestion be to also allow a single string that is used for all values?

nerik: What would be ideal, possibly, would be to follow the standard used across [i18n projects](https://www.i18next.com/translation-function/plurals), or some form of it:

```
strings: {
  stories_one: 'Story',
  stories_other: 'Stories',
  sheep: 'Sheep' // Sheep's plural is 'Sheep' so don't bother
}
```

So when looking up for a key, first look for a version with suffix (`one`, `other`, `zero`), then fallback on a version without suffix. The zero case fallbacks on `other` by default, but can also be specified as `stories_zero`.

nerik: We could also do some variation on it if it feels more comfortable:

```
strings: {
  stories: {
    one: 'Story',
    other: 'Stories'
  }
  sheep: 'Sheep' // Sheep's plural is 'Sheep' so don't bother
}
```

danielfdsilva: This is very interesting, and I like the automated version. Not sure it would work for multiple words, like in this case where in the ghg case, "stories" becomes "data stories"

It indeed seems that in English zero is always plural (aside from uncountable), so I agree to go with a more standard approach of `one|other`. I've a slight preference for the `object` example. Unless you have other arguments here.

nerik: Sounds good, and let's hope we never have to make an [arabic or welsh](https://www.linkedin.com/pulse/internationalization-localization-tips-plurals-daniel-neumann/) version of VEDA 🫠

danielfdsilva:
@nerik Updated to use discussed format

freitagb:
Looking at the netlify preview, I think we planned for these to be called "Data Stories" rather than "Stories" I think mdx organization is fine to be stories, but in the UI it should be "Data Stories"

j08lue:
> I think we planned for these to be called "Data Stories" rather than "Stories"

Good you stress that, @freitagb - I think what you are seeing in the preview is just the "dummy" content that is used in `veda-ui` PR builds. Once this gets deployed to the real instances, their config will define the name - and it will be "Data Stories" for GHGC and tbd for the NASA/EIS instance.

freitagb:
> > I think we planned for these to be called "Data Stories" rather than "Stories"
> 
> Good you stress that, @freitagb - I think what you are seeing in the preview is just the "dummy" content that is used in `veda-ui` PR builds. Once this gets deployed to the real instances, their config will define the name - and it will be "Data Stories" for both the GHGC and the NASA/EIS instance.

Thanks for the clarification @j08lue!

nerik (approved):

Issue #178, opened 2022-06-21
table component


hanbyul-here:
- the need for table component was raised for ej discovery : {placeholder for ej discovery url}

PR #597, opened 2023-07-27
Table component to display tabular data

hanbyul-here:
partially https://github.com/NASA-IMPACT/veda-config-ghg/issues/16  and close https://github.com/NASA-IMPACT/veda-ui/issues/178

preview is in sandbox: https://deploy-preview-597--veda-ui.netlify.app/sandbox/sandboxtable

This PR only handles the table component for the tabular data. (the original ticket doesn't require a reusable component, but a table that can read Excel file seems to be generic enough so I put it as a component. Also making one off thing for a table with these many rows without UI dev env seemed to be more work.) Once this is merged, I will help a data story configuration to use this component. 

I tried putting in some additional features like sorting, but it seemed too cluttered when the data has lots of columns.


j08lue (approved):
Very useful component - looking forward to seeing that used to present GHG data. 

Should we ask @ricardoduplos design review, perhaps? IMO, the up-down arrows ⇅ might be nicer than the arrow circle 🔄 for indicating that sorting is available on a column.

The choice of enabling the user to specify columns that the data can be sorted by is aligned with what we discussed - that the purpose of Dashboard Stories and Dataset pages is largely predefined (optimized) and contextualized data presentation, rather than free data exploration.

At the same time, we do not want to make the components too rigid, so they also can be used for limited exploration beyond the intended context. We just have to strike the right balance...

Approving this as it is now, but please feel free to continue to iterate, if you want.

hanbyul-here:
Yeah let's start here, and I will iterate it in the future.  I agree that 'columns to sort' s look can be much better. I was hoping there is something in our UI library but that loop icon was the best I could find. Do you think you can provide me with a better icon @ricardoduplos ? (I can work on more than an icon if you have the bandwidth to look at the general design too.)

danielfdsilva (changes requested):
We are using the ui library in the app. Its [styles for tables](http://ui.ds.io/latest/?path=/docs/components-typography--table) are enough as a starting point and could be worth exploring.

As for icons use:
```
CollecticonSortAsc
CollecticonSortDesc
CollecticonSortNone
```
image

💅 Please run prettier on the code files, to ensure style consistency and more readability, and include white space where deemed necessary.

hanbyul-here:
ah thanks for pointing Table style from UI library (it is in the typography package? interesting.) I had a brief discussion with Ricardo about Table and we agreed to work on it more in the near future.

hanbyul-here:
@danielfdsilva it looks much better thanks! 
re: column size. I am not sure if I am blindsight for something obvious, but I couldn't find a good solution to fix (but still accommodate the dynamics of the data) the column size with the virtualized rows. I also put it as 'something to improve for later', but please let me know if you already know how to implement it!

Issue #580, opened 2023-07-11
Implement exploration page with timeline


danielfdsilva:
The new exploration page with a [timeline approach was defined](https://github.com/NASA-IMPACT/veda-ui/issues/524) and designed. Using the [timeline proof of concept](https://github.com/NASA-IMPACT/veda-ui/tree/sandbox/timeline) (` sandbox/timeline` branch) as a starting point we can proceed to the implementation of this page.

This page will replace each dataset's individual explore page.

>  ⚠️ **We need to understand how mindful we need to be about url redirection**

This will be the first stage in the E&A merge and the analysis capabilities will be added later after the [analysis flow was prototyped](https://github.com/NASA-IMPACT/veda-ui/issues/579) and better understood.

<img width="1445" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/1090606/580f79a9-618f-4e1d-9555-7146dd200e27">

danielfdsilva:
General structure is [implemented](https://github.com/NASA-IMPACT/veda-ui/pull/598).

Issues remaining for the timeline:
- https://github.com/NASA-IMPACT/veda-ui/issues/605
- https://github.com/NASA-IMPACT/veda-ui/issues/606
- https://github.com/NASA-IMPACT/veda-ui/issues/607

Issue #579, opened 2023-07-11
[E & A] Prototype and validate analysis workflow


danielfdsilva:
The analysis as part of the E&A merge introduces new ways of [interacting with the areas of interest](https://github.com/NASA-IMPACT/veda-ui/issues/526). How the different methods of adding shapes to the map are used need to be tested to understand the optimal flow. Selecting shapes, running an analysis, updating the parameters and refreshing the analysis is a complex flow that is hard to fully test with static designs.

We need to prototype the part of the analysis related to the aoi selection and how an outdated analysis is displayed (result of changing parameters). This issue is not concerned about how the analysis results are displayed.

<img width="1445" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/1090606/d30d0ab4-60ce-4403-af78-fdc8c424fc4a">

j08lue:
NB: Review this old issue and verify that the unexpected behavior does not occur in the new implementation
- https://github.com/NASA-IMPACT/veda-ui/issues/388

Issue #527, opened 2023-04-28
[E & A] - Page status sharing


danielfdsilva:
> 🙋 This issue is part of the larger [Exploration/Analysis merge component breakdown](https://github.com/NASA-IMPACT/veda-ui/issues/513). The goal is to have a discussion about this specific component and come up with implementation details/next steps.  
Check out the parent issue for a list of all components to consider.

---

## Description

Users should be able to share the current state of the analysis/exploration page with others. Given the amount of available settings for the user to define (selected datasets and their settings, dates, analysis areas, etc), it may not be feasible to use url parameters for this (mainly due to the analysis polygon size https://github.com/NASA-IMPACT/veda-ui/issues/472). While maintaining this functionality is important we need a way that does not compromise the user experience.

Some discussed examples include:
- Downloading/Uploading a json file with all settings
- Using a server to cache the settings and generate an id

## Acceptance Criteria
- [ ] Having a list of pros/cons for each option the best option is decided.

Issue #526, opened 2023-04-28
[E & A] - Analysis worflow


danielfdsilva:
> 🙋 This issue is part of the larger [Exploration/Analysis merge component breakdown](https://github.com/NASA-IMPACT/veda-ui/issues/513). The goal is to have a discussion about this specific component and come up with implementation details/next steps.  
Check out the parent issue for a list of all components to consider.

---

## Description
<img width="933" alt="image" src="https://user-images.githubusercontent.com/1090606/235190447-47dbc466-6664-4d14-84db-43fad2d397eb.png">
<img width="935" alt="image" src="https://user-images.githubusercontent.com/1090606/235190513-ba49f1c1-0aee-4b0c-ace4-a46e8780eb1d.png">

The analysis workflow is to be merged with the exploration page so that the users can have additional context while running an analysis.
There are 3 elements necessary for an analysis:
- Spacial definition
- Temporal range
- Datasets to analyze (already selected)

**Temporal range**  
While the dataset exploration requires a single time step, the analysis requires a date range over which the data is analyzed. This date range selector should take into account how the [dataset exploration date is selected](https://github.com/NASA-IMPACT/veda-ui/issues/522) (e.g. single picker vs one picker per dataset)

**Spatial definition**  
To define a spatial area for the analysis, users should be able to:
- Draw an area of interest (https://github.com/NASA-IMPACT/veda-ui/issues/388)
- Upload their own file
   - How can users select what features they want from an uploaded file? (https://github.com/NASA-IMPACT/veda-ui/issues/426, related: https://github.com/NASA-IMPACT/veda-ui/issues/474)
- Select from a list (https://github.com/NASA-IMPACT/veda-ui/issues/405)
   - [Mapbox boundaries](https://demos.mapbox.com/boundaries-explorer/) could be used for geopolitical borders
   - Does selecting from a list add the polygon to the map, or are the polygons visible and the user has to select one?

## Acceptance Criteria
- [ ] The date range selector is designed taking into account how the dataset exploration date is chosen (https://github.com/NASA-IMPACT/veda-ui/issues/522)
- [ ] The workflow for drawing an AOI is defined including the possible shapes, controls and possible actions over drawn shapes
- [ ] The user upload workflow is designed
- [ ] The list selection workflow is designed

ricardoduplos:
[...]

PR #598, opened 2023-08-02
Exploration page (E&A merge)

danielfdsilva:
Is PR is serving as a base for the development on the new Exploration and Analysis page. Branching in/out of this.

During development, the new page is accessible at `/exploration`.

The sandbox code is still present but will be removed eventually. Disregard during reviews.

## To Do

- https://github.com/NASA-IMPACT/veda-ui/issues/605
- https://github.com/NASA-IMPACT/veda-ui/issues/526
- https://github.com/NASA-IMPACT/veda-ui/issues/524
- https://github.com/NASA-IMPACT/veda-ui/issues/522
- https://github.com/NASA-IMPACT/veda-ui/issues/580
- https://github.com/NASA-IMPACT/veda-ui/issues/521
- https://github.com/NASA-IMPACT/veda-ui/issues/523
- https://github.com/NASA-IMPACT/veda-ui/issues/520
- https://github.com/NASA-IMPACT/veda-ui/issues/527

- https://github.com/NASA-IMPACT/veda-ui/issues/720
- https://github.com/NASA-IMPACT/veda-ui/issues/721
- https://github.com/NASA-IMPACT/veda-ui/issues/722
- https://github.com/NASA-IMPACT/veda-ui/issues/723
- https://github.com/NASA-IMPACT/veda-ui/issues/724
- https://github.com/NASA-IMPACT/veda-ui/issues/579
- https://github.com/NASA-IMPACT/veda-ui/pull/771

## TBD

- https://github.com/NASA-IMPACT/veda-ui/issues/519


app/scripts/components/exploration/timeline.tsx
```diff
@@ -0,0 +1,472 @@
+import React, { useEffect, useMemo, useRef, useState } from 'react';
```
nerik, line 1: This file is getting on the larger side. Maybe consider starting with extracting away `Controls` and `DatasetList`?

app/scripts/components/exploration/atoms.ts
```diff
@@ -0,0 +1,37 @@
...
+export const selectedIntervalAtom = atom<DateRange | null>(null);
+// Zoom transform for the timeline. Values as object instead of d3.ZoomTransform
+export const zoomTransformAtom = atom<ZoomTransformPlain>({
+  x: 0,
+  y: 0,
+  k: 1
+});
+// Width of the whole timeline item. Set via a size observer and then used to
+// compute the different element sizes.
+export const timelineWidthAtom = atom<number | undefined>(undefined);
```
nerik, line 24: I think I could use a high-level explanation as to why this is part of the app state. Isn't that defined solely by the available witdh of the timeline viewport?

app/scripts/components/exploration/timeline.tsx
```diff
@@ -0,0 +1,472 @@
...
+        </TimelineContentInner>
+      </TimelineContent>
+    </TimelineWrapper>
+  );
+}
+
+export default Timeline;
+
+function DatasetList(props: any) {
+  const { datasets, ...rest } = props;
```
nerik, line 452: I would recommend against passing down `datasets`, and directly use the atom value instead:

```suggestion
  const datasets = useAtomValue(timelineDatasetsAtom);
```

IMHO it only makes sense to pass down a prop containing data when the component renders an element of an array, ie `<Rabbit rabbitData={rabbits[i]} />`

app/scripts/components/exploration/timeline.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+  const [orderedDatasets, setOrderDatasets] = useState(datasets);
```
nerik, line 454: Instead of duplicating datasets and create potential syncing issues, why not altering the order of datasets on the atom directly?

app/scripts/components/exploration/timeline-head.tsx
```diff
@@ -0,0 +1,200 @@
...
+        }}
+      />
+      <text fill={theme.color?.base} fontSize='0.75rem' y='0' x='2' dy='1em'>
+        L
+      </text>
+    </TimelineHead>
+  );
+}
+
+export function TimelineHeadR(props: Omit<TimelineHeadProps, 'children'>) {
```
nerik, line 135: You could wrap TimelineHeadP/L/R into a single component passing down the letter and the path object.

app/scripts/components/exploration/atoms.ts
```diff
@@ -0,0 +1,37 @@
...
+import {
+  DateRange,
+  HEADER_COLUMN_WIDTH,
+  RIGHT_AXIS_SPACE,
+  TimelineDataset,
+  ZoomTransformPlain
+} from './constants';
+
+// Datasets to show on the timeline and their settings
+export const timelineDatasetsAtom = atom<TimelineDataset[]>([]);
```
nerik, line 11: Have you tried playing with [atomWithHash](https://jotai.org/docs/integrations/location#atomwithhash) ?

nerik (changes requested):
This is really impressive work 👏 

There are a few technical things that we may want to discuss further, otherwise I have some feedback on useability:

- Disable text selection when dragging datasets (on datasets names, etc);
- Users should be able to change L and R markers at the same time by dragging the inner partM
- TBH I'm not 100% sure what L, R and P markers do. And I think it could be useful for the letters to be explained, I mean I guess that L and R means Right and Left, but _of what_ ? Could we use icons to aid understanding of this?
- The timeline whole area should be draggable (maybe it's already the case but I didn't find how);
- The whole timeline is zoomable, which is pretty awesome. However I found out about this by reading the code and performing an horizontal pinch on my trackpad with the Alt key pressed. I think we could use some improved discoverability of the feature;
- When selecting an L/R range that is displayed outside the current timeline viewport, it should automatically move the timeline viewport there;
- Should the date range picker automatically open by default on the start date of the timeline viewport, rather than _today_?;

app/scripts/components/exploration/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    flex: 0;
+    position: relative;
+    outline: none;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 5rem;
+    margin: 0 auto -1.25rem auto;
+    padding: 0.25rem 0;
+    z-index: 5000;
```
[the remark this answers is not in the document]
danielfdsilva: Actually there's a list defined in the theme. We just haven't been using it. 😢 

https://github.com/NASA-IMPACT/veda-ui/blob/1ced2682b23fcc606f734e5d6245237a3a3fd723/app/scripts/styles/theme.ts#L9-L19

app/scripts/components/exploration/atoms.ts
danielfdsilva: There are several elements that need to know the width (specially the svg components), so I thought of storing the size in an atom and access it via a hook instead of having to pass width to every single component (this is something I still have to refactor) - Thoughts on the approach?

app/scripts/components/exploration/timeline-head.tsx
danielfdsilva: There are several properties that are changing, not only the path (transform, text position, etc)

danielfdsilva:
Thank you for the great feedback items. There are a lot of accessibility improvements to be done to the timeline, and this is a great starting list. Some things will have to addressed via [help information](https://github.com/NASA-IMPACT/veda-ui/issues/519) (like the zoom with `option|alt`).

> The timeline whole area should be draggable (maybe it's already the case but I didn't find how);

You only need to click and drag (once zoomed in, if you're viewing everything there's no dragging). Does this not work? 🤔

nerik:
@danielfdsilva How would you feel about adding some structure to the `exploration` folder?
Maybe:
```
📁 atoms 
    📄 atoms.ts
    📄 hooks.ts (move atom-related hooks here)
📁 components
    - ...
📁 hooks
    📄 hooks.ts
    📄 use-dataset-hover.ts
📄 constants.ts
📄 types.d.ts (extract types form constants)
📄 utils.ts
```

nerik:
In order to test this, you can access the Netlify preview here: 
https://deploy-preview-598--veda-ui.netlify.app/exploration
(that page has not been made reachable yet)

Some options not yet implemented in the final UI are accessible by hitting Alt-Shift-M

cc @j08lue

j08lue:
We received feedback about the old analysis page in the context of the GHG Center. Most of it we addressed in a quick fix of the current interface:

- https://github.com/NASA-IMPACT/veda-ui/issues/654
- https://github.com/NASA-IMPACT/veda-config-ghg/issues/129

Since some of the feedback for the old Analysis flow is also relevant to the new, I would like to make sure we keep track of those and see whether the new design can also accommodate them:

1. Show only specific statistics/metrics by default (configurable per dataset) https://github.com/NASA-IMPACT/veda-ui/pull/673
2. Hide some datasets from selection for analysis (and exploration)
3. Offer date range presets (e.g. last year) - configurable per instance
4. Offer area presets (administrative boundaries etc.) - configurable per instance

Furthermore, these are the top features currently still missing from the Analysis feature:

1. Have a labelled Y axis on time series plots
2. Make time series plots vertically expandable

In my personal opinion, the many cursors and their purpose are confusing. When entering the app, I do not need the L/R cursors, do I? Or is there going to be a "play back" function to step through the dates between L and R?

No need to answer all of these in detail now, just as a basis for our next review/discussion in-person.

danielfdsilva:
We looked to bring feature parity with the previous analysis and exploration, but there are some things that are not yet implemented.
From the previous comment list:

1) Metrics are configurable per dataset basis, respecting default values set in the config file.
2) Datasets are excluded from analysis either via the `analysis.exclude` or the `disableExplore` config options.
   - The `disableExplore` doesn't make sense anymore because analysis and exploration are together now, and it should be removed in the future.
3) Date range presets are not implemented and would merit some discussion.
   - We need to be careful not to clutter the interface with even more controls.
   - The previous analysis required a date range to see what datasets were available for analysis, but that is no longer the case. Now the datasets and their temporal extend are immediately shown, making it easier for the user to select the dates they need. It would be good to verify if this feature still makes sense.
4) We started with the process of planning the AOI drawing/upload/preset feature but that was not finalized. Before jumping into implementation we need some data architecting and design around this.
   - The design flow for presets was started but then efforts moved elsewhere (perhaps @ricardoduplos  can still recall enough to comment on this).
   - Where the preset geometries come from/how they are organized/stored and used it is still an open question.

----

1) For the **labeled y axis**, I just opened a PR: https://github.com/NASA-IMPACT/veda-ui/pull/716

2) Not sure what you mean by making plots vertically expandable. Can you clarify?

j08lue:
Thanks for the walk through, @danielfdsilva!

> 2\. Not sure what you mean by making plots vertically expandable. Can you clarify?

I think that users will sometimes need more vertical space to view the time series graph properly, and the designs on Figma seemed to include vertical expansion of some sort:

image
image

danielfdsilva:
@j08lue oh, I see. that's implemented.
Screenshot 2023-10-30 at 13 33 15

It is not terribly visible, but that that will have to be part of the user help that is yet to be designed.

j08lue:
Layer-specific choice of default map projection - will that be a thing in E& again?

- https://github.com/NASA-IMPACT/veda-ui/issues/755

hanbyul-here:
🤚 Have you considered integrating Jotai Devtool? https://jotai.org/docs/tools/devtools. Is there any specific reason you are not using it?

danielfdsilva:
@hanbyul-here Honestly I hadn't even seen them. 😅 Maybe could have been helpful.
Can we get a final review here to merge?

cc @j08lue

j08lue (changes requested):
## Ghost time series value tooltip

After I "Exit Analysis", the time series values tooltip still seems to be active, even though the time series is not shown anymore in the bottom drawer. Steps to reproduce:

1. Add a layer
2. Run analysis - take note of where on the bottom drawer there are values
3. Exit analysis
4. Hover over bottom drawer over an area where there were time series values
5. Issue: value tooltip is shown even though no time series is visible

Screenshot 2023-12-06 at 13 20 17

danielfdsilva:
@j08lue That was a cute one 😅. Fixed

j08lue (changes requested):
## Datasets vs layers

Sorry - late find: We need to replace "dataset" with "layer" in the layer selection modal for now, please, until we introduce a new concept for datasets as layer groups.

image

danielfdsilva:
@j08lue Addressed

j08lue (approved):
All review comments addressed! We are now asking for key stakeholder reviews on the preview deployment.

app/scripts/components/data-catalog/index.tsx
```diff
@@ -90,6 +89,7 @@ const prepareDatasets = (
         d.name.toLowerCase().includes(searchLower) ||
```
hanbyul-here, line 89: Shouldn't there be `d.id.toLowerCase().includes(searchLower) ||` ?

danielfdsilva: Great catch @hanbyul-here! Funnily enough this was working well, but it was a coincidence. Fixed!

app/scripts/main.tsx
```diff
@@ -56,6 +58,8 @@ const composingComponents = [
   LayoutRootContextProvider
 ];
 
+const useNewExploration = !!process.env.FEATURE_NEW_EXPLORATION;
```
hanbyul-here, line 61: One more suggestion to use an explicit value for this environment variable, but other than that I think this should be good to go!

danielfdsilva: Added a simple function to be used in cases like this `checkEnvFlag(process.env.VAR_NAME)`

faustoperez:
Thank you, Daniel! 🙌

PR #604, opened 2023-08-07
Story authoring tool

nerik:
- Adds an entry in the global nav to access the publication tool (not in production)
- Adds a simple UI for managing editable stories (only in local storage)
- Allows editing stories by individual blocks that can be moved around, deleted, edited individually

TODO:
- Missing the other sections of the story editing:
  - Metadata editor
  - "Export" which really should be a raw MDX editor thing 
  - Possibly a "Preview" tab
- Missing the option to create a story from scratch
- Error handling is not working properly
- Currently the model is a structured object with an array for blocks, rather than the full MDX document. Could be maybe more efficient to be the other way around: full MDX document as the source of truth, with a getter to split the document by block and extract the frontmatter.


j08lue:
Great how this is coming together!

I am aware that this is just a placeholder, but can we already pick a different place for this tool than the top nav? The footer, maybe, or the hamburger / kebab menu on the right?

Just so it is a bit less invasive when we demo this feature. 😄 

Really just a quick fix - we can think more thoroughly about how we will advertise / expose this tool to our content contributors later.

nerik:
@j08lue Yeah ok, maybe that was a bit invasive 😅 Moved that to the homepage footer:

Screenshot 2023-10-12 at 15 15 54

hanbyul-here:
Recording Story Editor preview url : https://deploy-preview-604--veda-ui.netlify.app/story-editor

PR #609, opened 2023-08-09
Align legend unit

### LINKED ISSUE #599: Alignment of label in Units (Legend)
Currently, units in the legend section are left aligned and a lot of times it gets broken into multiple lines.

Screen Shot 2023-08-03 at 10 20 16 AM

We want to center-align the units in the legend section so that it is more clear and doesn't overlap with the numbers.

danielfdsilva:
Fix #599
Aligns legend units to the center.
image


danielfdsilva:
cc @deborahUAH When the legend has numeric values, hovering it will give you the value for the different points

hanbyul-here (approved):
👍

PR #619, opened 2023-08-15
Add dataset layer visibility toggle button to explore page

hanbyul-here:
Requested feature from JPL EMIT. The visibiility toggle in this pr is global, not per layer.


danielfdsilva (approved):
Made a quick style change.
Otherwise looks good - a nice temporary fix

PR #620, opened 2023-08-15
Include center coordinates control

danielfdsilva:
Adds a map control to display the coordinates of the center point and allows the user to copy them on click.

This approach was chosen in favor of "coordinates under the cursor", given that this is more friendly towards touch devices and allows the coordinates to be stable while the user goes to the control and copies them.

coords


hanbyul-here (approved):
This solution is great! 👍

PR #623, opened 2023-08-16
Put tile urls for datasets

### LINKED ISSUE #594: Surface tiler links for dataset layers
We would like to enable users of VEDA to load the same layer they are seeing in the VEDA Dashboard into their GIS software, for example QGIS or ArcGIS.

[Previously](https://github.com/NASA-IMPACT/veda-architecture/issues/293), we have been focusing on including copy-pasteable tiler URLs in STAC.

Actually, this might be much better suited to be a feature implemented in `veda-ui` - the information about the tiler being used and in particular the styling is actually coming together in the Dashboard, rather than STAC.

Could we imagine adding another "bridge" button or something to the dataset Exploration page that basically lets a user copy a tiler WMTS or XYZ link (which ever we are actually using) for the view they are currently looking at?

## Acceptance criteria

- [ ] Documented discussion and decision on whether and how we could address the user need for getting copy-pasteable tiler URLs from Exploration

hanbyul-here:
This PR is based on : https://github.com/NASA-IMPACT/veda-ui/pull/620 
Close #594 

- [x] check vector data

### DIFF
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -373,6 +373,8 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
             // Ignore errors.
           }
 
+          const wmtsBaseUrl = mosaicUrl.replace('tilejson.json', 'WMTSCapabilities.xml');
+
           const mosaicSource: RasterSource = {
             type: 'raster',
             url: tilejsonUrl
@@ -396,7 +398,7 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
               id,
               layerOrderPosition: 'raster',
               xyzTileUrl: tileServerUrl,
-              wmtsTileUrl: `${mosaicUrl}/WMTSCapabilities.xml?${tileParams}`
+              wmtsTileUrl: `${wmtsBaseUrl}?${tileParams}`
             }
           };
 
--- a/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
+++ b/app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
@@ -373,7 +373,9 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
           },
           minzoom: minZoom,
           metadata: {
-            layerOrderPosition: 'raster'
+            layerOrderPosition: 'raster',
+            xyzTileUrl: `${mosaicUrl}?${tileParams}`,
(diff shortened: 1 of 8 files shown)


app/scripts/styles/theme.ts
```diff
@@ -16,7 +16,7 @@ export const VEDA_OVERRIDE_THEME = {
     popover: 1500,
     skipLink: 1600,
     toast: 1700,
-    tooltip: 1800
+    tooltip: 9999
```
hanbyul-here, line 19: I edited to show the tooltip on top of Modal, and this probably is not ideal if there is a tooltip (not from the modal) and the modal at the same time on the screen. Do you know any better way to handle this? @danielfdsilva

danielfdsilva (approved):
This was a great base to start. 🎉 I rebased the work after merging the coordinates PR. (be sure to reset your branch)

- I noticed that the xyz endpoint was actually the tilejson, which may not be exactly what we want, so I fetched and stored the xyz url.
- I did not change anything about the wmts as I'm not sure about the format.
- Added support for the vector data tile endpoint
- Tweaked the widget styles and the different types

I think this is good to go! (maybe double check the wmts)

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -337,92 +337,123 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
...
+              'raster-opacity-transition': {
+                duration: 320
+              }
+            },
+            minzoom: minZoom,
+            metadata: {
+              id,
+              layerOrderPosition: 'raster',
+              xyzTileUrl: tileServerUrl,
+              wmtsTileUrl: `${mosaicUrl}/WMTSCapabilities.xml?${tileParams}`
```
hanbyul-here, line 399: @anayeaye can you confirm this is the right format for wmts tile URL ?

anayeaye: I'm not sure: if `mosaicUrl` = `/mosaic/{searchid}/{TileMatrixSetId}` or `/mosaic/{searchid}` this works. 

If mosaicUrl includes `/tiles/` like `/mosaic/tiles/{searchid}/...`, it won't work (which is generally how the XYZ urls are formatted).

anayeaye: Just looked at the preview `tiles.json` needs to be stripped from `/mosaic/cfa5568f19003ffe39f4f05c9a38b481/tilejson.json/WMTSCapabilities.xml?`

hanbyul-here:
@anayeaye brought up a good point that custom color formula can make it look like it is not working (ex. color_formula=gamma+r+1.05.). I don't think this should be a blocker, but worth noting. Thanks for testing the features 🙇 🙇

PR #625, opened 2023-08-16
Add country coastlines

danielfdsilva:
Mapbox admin boundaries layer does not include the coastlines by default.
I had to add them to every style we use, which involved creating a new group for the coastlines layers (on each style). Because of the lack of group id customization, I had to make some changes to the code to account for this.


aboydnw (approved):
Looks good to me!

hanbyul-here (approved):

PR #629, opened 2023-08-18
Add dataset classification support

danielfdsilva:
A dataset is classified according to 2 values:

- `Uncertainty`: Degree of uncertainty of this dataset.
  - High
  - Medium
  - Low
- `Nature`: The nature of this dataset indicates the status of its data.
  - Research
  - Agency standard/regulatory

These values are added to the dataset like taxonomies, but must have only one value from the list and are then displayed as icons in the dataset card.

| YMAL | Result |
|--|--|
|image|image|

----

We want to make sure that we have the correct labels for names and for the properties. We decided on "Nature" to describe the status of the data, but may not be the best option.

@j08lue Can you help us validate these labels? If Nature is not appropriate, we are looking for a single word to describe this (can't be "Type").


j08lue:
> We want to make sure that we have the correct labels for names and for the properties. We decided on "Nature" to describe the status of the data, but may not be the best option.
> 
> @j08lue Can you help us validate these labels? If Nature is not appropriate, we are looking for a single word to describe this (can't be "Type").

In the [dataset description template](https://docs.google.com/document/d/18qV6dI2Nb2KdHQ-eI83aIcYBHNSHIJskfCHIiCDoA0o/edit#heading=h.g5xc8g7us147) that @Jeanne-le-Roux devised, they call this category "Data Type". I think it describes the grade or standard of the dataset, or what type of use the dataset is suitable or recommended for.

Off the top of my head, I would either go for "Suitability" or "Grade", with a slight preference for the latter.

danielfdsilva:
@j08lue That was a great suggestion! Changed to "Grade"

PR #630, opened 2023-08-22
Fix contextual link for taxonomy pills

danielfdsilva:
Great catch @j08lue


j08lue (approved):
Nice, taxonomy links on stories now point to the right target (stories overview).

PR #632, opened 2023-08-24
Improve shadow scrollbar behavior

danielfdsilva:


j08lue (approved):
Great, I can see the scroll bar now flashes on first reveal or mouse over the component, making it more apparent that there are more controls to scroll down to.

image

PR #633, opened 2023-08-24
Make basemap a configurable dataset property

danielfdsilva:
The basemap property was not being tracked in the url, meaning that if a user changes it and shares the url they'd get the default satellite. It was also not possible to specify the starting basemap id (like you could do with the projection). This PR addresses this.


j08lue (approved):
Quick turnaround on completing this feature, complete with docs. 👌👌👌

app/scripts/components/datasets/s-explore/index.tsx
```diff
@@ -374,6 +385,33 @@ function DatasetsExplore() {
...
+      if (currActiveData.basemapId) {
+        setBasemapId(currActiveData.basemapId);
+      } else {
+        setBasemapId(BASEMAP_ID_DEFAULT);
+      }
+    },
+    [activeLayer]
+  );
+
+  // On layer change, reset the basemapId.
```
hanbyul-here, line 397: Is this comment at the right place? It seems to describe what https://github.com/NASA-IMPACT/veda-ui/pull/633/files#diff-be896ef6bddd109e493aede34acda96085cbbcfa648f310813b3f10842c2de61R388-R392 does. am I reading it wrong?

hanbyul-here (commented):
Two different variables starting with  `base~` don't seem to be ideal, but also this page won't live too long, so it should not be a blocker.
I left a question about the comment, but feel free to merge it if I am reading it wrong!

app/scripts/components/datasets/s-explore/index.tsx
danielfdsilva: good catch. had fixes this, but apparently not saved the file 😅

hanbyul-here (approved):

PR #639, opened 2023-08-30
Implement preliminary dataset selector modal

danielfdsilva:
The modal is very basic and lack all design and dataset search options, but it starts to become possible to select the existing datasets. From this we can stop using the timeline mock data and start to use the repo data.

This PR also removes the timeline tests from the sandbox to fix the linting and TS errors


app/scripts/components/exploration/dataset-options.tsx
```diff
@@ -48,16 +52,22 @@ export default function DatasetOptions(props: DatasetOptionsProps) {
...
-                datasets.filter((d) => d.data.id !== dataset.data.id)
-              );
-            }}
+          <Tip
+            disabled={datasets.length > 1}
+            content="It's not possible to remove the last dataset. Add another before removing this one."
           >
-            <CollecticonTrashBin /> Remove dataset
-          </DropMenuItemButton>
+            <RemoveButton
```
nerik, line 59: Found it a little bit odd that the disabled "Remove dataset" button get's a highlight effect on hover (I guess it's related to the tooltip).

Screenshot 2023-09-04 at 10 30 05

nerik (commented):
Great work!
The only thing I'd like to see before merging: can the modal be closed by default, rather than having it open on mount? It's going to be annoying for development IMHO. 
We could have a dataset selected/reconcilied by default (or several). I also think that maintaining the mock datasets for a while could be useful, maybe directly integrated in the modal?
I'm also curious to see how you would like to handle state URL, let's discuss this at some point.

app/scripts/components/exploration/dataset-options.tsx
danielfdsilva: Good catch! Fixed! :)

PR #641, opened 2023-08-31
Fix projection and basemap switch

danielfdsilva:
You can use the no2 mock dataset to test


j08lue (approved):
Works nicely now - opening Exploration on the NO2 dataset loads the first layer in a different projection than Mercator (nice choice of north-polar stereographic for Portugal, btw!).

PR #645, opened 2023-09-04
Integrates STAC metadata request

danielfdsilva:
When datasets are added to the timeline, metadata from STAC is requested and reconciles with local data

### DIFF
--- a/app/scripts/components/exploration/types.d.ts.ts
+++ b/app/scripts/components/exploration/types.d.ts.ts
@@ -1,3 +1,5 @@
+import { DatasetLayer } from "veda";
+
 export enum TimeDensity {
   YEAR = 'year',
   MONTH = 'month',
@@ -7,12 +9,18 @@ export enum TimeDensity {
 export enum TimelineDatasetStatus {
   IDLE = 'idle',
   LOADING = 'loading',
-  SUCCEEDED = 'succeeded',
-  ERRORED = 'errored'
+  SUCCEEDED = 'success',
+  ERRORED = 'error'
+}
+
+export interface StacDatasetData {
+  isPeriodic: boolean;
+  timeDensity: TimeDensity;
+  domain: string[];
 }
 
 export type AnalysisTimeseriesEntry = Record<string, number | null> & {
-  date: Date;
+  date: Date; 
 };
 
 export interface TimelineDatasetAnalysis {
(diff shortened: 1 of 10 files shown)


app/scripts/components/exploration/types.d.ts.ts
```diff
@@ -27,9 +35,15 @@ export interface TimelineDatasetAnalysis {
...
+  domain: Date[];
+}
+
 export interface TimelineDataset {
   status: TimelineDatasetStatus;
-  data: any;
+  data: TimelineDatasetData;
   error: any;
   settings: {
     // user defined settings like visibility, opacity
```
danielfdsilva, line 49: @nerik How much type narrowing should we be doing?

The `isPeriodic` and so on properties are only available when the status is `success`, otherwise they're `undefined`. We could do this type narrowing or set it to be possibly undefined. On one hand we have a more complicated type, but with `undefined`, we need to check every prop, instead of relying on status. Thoughts?

app/scripts/components/exploration/types.d.ts.ts
```diff
@@ -7,30 +9,109 @@ export enum TimeDensity {
...
+    // user defined settings like visibility, opacity
+    isVisible?: boolean;
+    opacity?: number;
+  };
+  analysis: TimelineDatasetAnalysisIdle;
+}
+export interface TimelineDatasetSuccess {
+  status: TimelineDatasetStatus.SUCCESS;
+  data: TimelineDatasetData;
+  error: null;
```
nerik, line 114: This can be simplified by:
- wrapping settings into a common Settings typr:
```
interface Settings {
  isVisible?: boolean;
  opacity?: number;
}
```
- extend a base type for all the unioned types:
```
interface TimelineDatasetBase {
  data: DatasetLayer | TimelineDatasetData;
  error: unknown | null;
  settings: Settings;
  analysis: TimelineDatasetAnalysisIdle | TimelineDatasetAnalysis;
}
```

nerik (changes requested):
At first glance, not sure what's going on, but do the following:
- Add a dataset. Time extent displays fine on the timeline
- Add another dataset. After the STAC endpoint fetch, no time extent displayed for both datasets.

danielfdsilva:
@nerik whit which datasets did you notice this happening?

danielfdsilva:
@nerik now that we addresses the issues with the dataset domain, I think this is good to go. Can you have a final check?

nerik (approved):

PR #646, opened 2023-09-05
Feature/exploration map

nerik:
Adds a new Map component whose goal is to be reusable and composable for all the dashboard use cases.
This PR is the first step, limiting itself to demo how the maps would be configured:

```
              <Map>
                <Basemap basemapStyleId='satellite' />
                <GeocoderControl />
                {compare && (
                  <Compare>
                    <Basemap basemapStyleId='dark' />
                  </Compare>
                )}
              </Map>
```

For now functionally, this only includes:
- compare map functionality
- the styles/generators managers as it was before 
- the basemap generator as it was before 
- a Geocoder control as an example of how controls could be built

And in terms of functionality, we should be able to merge this PR as is.

One thing that I'm struggling is with the map height vs Panels, I'd like to talk with you @danielfdsilva before doing any further research. 

Next steps:
- Bring back raster|vector timeseries generators
  - ideally figuring out a way to avoid having to deal with the mapInstance inside the generator
- Bring back all controls
- Connect with the datasets atoms + STAC metadata


danielfdsilva (changes requested):
You can easily achieve the resizing by observing the `MapContainer` size and updating the maps.

```jsx
 const { observe } = useDimensions({
    onResize: () => {
      setTimeout(() => {
        mapRef.current?.resize();
        mapCompareRef.current?.resize();
      }, 0);
    }
  });

  const mapRef = useRef();
  const mapCompareRef = useRef();

  return (
    <MapContainer id='comparison-container' ref={observe}>
      <MapProvider>
        <CompareHandler />
        <Map
          ref={mapRef}
          id='main'
          generators={generators}
          controls={controls}
        />
        {compareGenerators.length && (
          <Map
            ref={mapCompareRef}
            id='compared'
            generators={compareGenerators}
            controls={controls}
          />
        )}
      </MapProvider>
    </MapContainer>
  );
```

You need to modify the `Map` component to forward the ref.
The `setTimeout`is needed to avoid a map flickering.

You'll also need to add a `padding-top: 0.8rem;` to `.resize-handle` on `exploration/index` to fix the visual glitch.

app/scripts/components/common/map/map-wrapper.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {compareGenerators.length && (
+          <Styles>
+            {compareGenerators}
+            <MapComponent id='compared' controls={[]} />
+          </Styles>
+        )}
+      </MapWrapperContainer>
+    </MapWrapperContext.Provider>
+  );
+}
```
danielfdsilva, line 105: It would be a good idea to have an `id` property passed to the map to then be used as a prefix for these ids. If we end up reusing this component in the story pages, we'll have id clashes which can cause bugs.

app/scripts/components/common/map/map-wrapper.tsx
```diff
@@ -0,0 +1,116 @@
...
+  return (
+    <MapWrapperContext.Provider
+      value={{ initialViewState, setInitialViewState }}
+    >
+      <MapWrapperContainer id='comparison-container' ref={observe}>
+        <Styles>
+          {generators}
+          <MapComponent id='main' controls={controls} />
+        </Styles>
+        {compareGenerators.length && (
```
danielfdsilva, line 96:
```suggestion
        {!!compareGenerators.length && (
```
Otherwise a `0` gets printed when there are no generators

app/scripts/components/common/map/map-wrapper.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  & > * {
+    position: absolute !important;
+    top: 0;
+    bottom: 0;
+    left: 0;
+  }
+
+  ${MapboxStyleOverride}
```
danielfdsilva, line 34:
```suggestion
  ${MapboxStyleOverride}
  
  height: 100%;
```

The map container itself should not be absolutely positioned. It is better if the parent takes care of positioning it.

The parent, being a flex container will ensure that the map takes up the full size with just `height: 100%`.

Now, the mock controls and the compare button also take up space, but they should be moved elsewhere as they do not belong there.

danielfdsilva (changes requested):
Additionally:
- We're using kebab-case for file names. for consistency sake the `styleGenerator` folder should be renamed.

app/scripts/components/common/map/map-wrapper.tsx
nerik: It was a bit harder than expected: https://github.com/NASA-IMPACT/veda-ui/pull/646/commits/46be4fc26e32f4e9ef27c42e48c0379d84769a99

app/scripts/components/common/map/map-wrapper.tsx
nerik: > The parent, being a flex container will ensure that the map takes up the full size with just height: 100%.

It should, yes, but here `MapsContainer` while being set at 100% height, doesn't occupy the whole vertical space. Not sure what's going on here :/

nerik:
@danielfdsilva Thanks. I fixed the keyboard shortcut as it would not trigger with my keyboard layout. 
Unless I'm missing something, your CSS fix breaks the compare mode.

danielfdsilva (changes requested):
Looking good.
The compare position no longer resets on map move. The only thing left would be to instantiate the controls on the compare map as well and avoid the following:

image

nerik:
@danielfdsilva Thanks. Will address controls on a later PR.

PR #649, opened 2023-09-08
Lump dataset blocks

danielfdsilva:
Merges dataset blocks visually to allow for greater zooming out.

### DIFF
--- a/app/scripts/components/exploration/components/datasets/block-utils.ts
+++ b/app/scripts/components/exploration/components/datasets/block-utils.ts
@@ -1,0 +1,81 @@
+import {
+  endOfDay,
+  endOfMonth,
+  endOfYear,
+  startOfDay,
+  startOfMonth,
+  startOfYear,
+} from 'date-fns';
+import { TimeDensity } from '$components/exploration/types.d.ts';
+
+/**
+ * Calculate the start and end of a block of time, given a date and a time
+ * density
+ * @param date Starting date
+ * @param timeDensity Dataset time density
+ *
+ * @returns Array of two dates, the first being the start of the block and the
+ * second being the end of the block.
+ */
+export function getBlockBoundaries(date: Date, timeDensity: TimeDensity) {
+  switch (timeDensity) {
+    case TimeDensity.MONTH:
+      return [startOfMonth(date), endOfMonth(date)];
+    case TimeDensity.YEAR:
+      return [startOfYear(date), endOfYear(date)];
+  }
+
+  return [startOfDay(date), endOfDay(date)];
+}
+
+/**
+ * Lumps blocks of time together if they are too small to be seen on the chart.
+ *
+ * @param params.domain The dataset domain from which to calculate the blocks
+ * @param params.xScaled The xScaled function from the chart
+ * @param params.timeDensity The dataset time density
+ * @param params.minBlockSize The minimum size of a block
+ *
+ * @returns The blocks and whether or not any lumping happened.
+ */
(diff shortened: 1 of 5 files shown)


app/scripts/components/exploration/components/datasets/dataset-list-item.tsx
```diff
@@ -367,51 +351,64 @@ function DatasetTrack(props: DatasetTrackProps) {
...
+          />
+          {wasLumped && (
+            <DatasetGLabel
+              fill='white'
+              x={Math.max(xScaled(blockStart), 2) + 2}
+              y={12}
+            >
+              G
+            </DatasetGLabel>
+          )}
```
nerik, line 378: Not a fan of the 'G' letter to signal a group. Is there a more visual way to express this? Like for instance:

- a "stacked" look
Screenshot 2023-09-14 at 09 34 39

- or an SVG pattern that reproduces what it already looks like when not lumped:
Screenshot 2023-09-14 at 09 36 00

app/scripts/components/exploration/components/datasets/block-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Lumps blocks of time together if they are too small to be seen on the chart.
+ *
+ * @param params.domain The dataset domain from which to calculate the blocks
+ * @param params.xScaled The xScaled function from the chart
+ * @param params.timeDensity The dataset time density
+ * @param params.minBlockSize The minimum size of a block
+ *
+ * @returns The blocks and whether or not any lumping happened.
+ */
+export function lumpBlocks({ domain, xScaled, timeDensity, minBlockSize = 4 }) {
```
nerik, line 41: Have we considered the "disjointed repetitive block" use case? (ie, every day of 2015 then every day of 2017). 
I think that's not allowed by the STAC API, but just checking.

app/scripts/components/exploration/components/datasets/dataset-list-item.tsx
```diff
@@ -367,51 +351,64 @@ function DatasetTrack(props: DatasetTrackProps) {
...
-  const { xScaled, date, dataset, isVisible } = props;
+  const { xScaled, startDate, endDate, isVisible } = props;
 
   const theme = useTheme();
 
-  const [start, end] = getBlockBoundaries(date, dataset.data.timeDensity);
-  const s = xScaled(start);
-  const e = xScaled(end);
+  const s = xScaled(startDate);
+  const e = xScaled(endDate);
```
nerik, line 398: (nit=
```suggestion
  const xStart = xScaled(startDate);
  const xEnd = xScaled(endDate);
```

nerik (approved):
Works great 👍 
A few non-critical comments/questions, feel free to disregard.

app/scripts/components/exploration/components/datasets/dataset-list-item.tsx
ricardoduplos: I like this "stacked" look suggestion a lot! CC/ @danielfdsilva

app/scripts/components/exploration/components/datasets/block-utils.ts
danielfdsilva: Not sure I understood what you mean here

app/scripts/components/exploration/components/datasets/dataset-list-item.tsx
danielfdsilva: This was a great suggestion!

PR #651, opened 2023-09-11
Exploration map controls

nerik:
Adds coords, scale, zoom, and map options controls, on both compare maps.


danielfdsilva (changes requested):
This is looking good! A couple of things:
- Selecting a different base style does not update the compare map (the projection does)
- We should disable the pitch change (right mouse click)

danielfdsilva (changes requested):
The basemap now updates both maps and the pitch was disabled.
However, the "labels" and "boundaries" are only toggled for the main map

danielfdsilva (approved):
@nerik I went ahead and fixed the labels and boundaries issue to speed up things.

This is good to go! :shipit:

PR #655, opened 2023-09-14
Raster timeseries

nerik:
Adds raster timeseries back to the new map component with some architecture fixes (please review ignoring the first commit which was mostly copying files from the old version around).


app/scripts/components/common/map/hooks/use-maps.ts
```diff
@@ -1,13 +1,19 @@
 import { useContext } from 'react';
 import { useMap } from 'react-map-gl';
 import { MapsContext } from '../maps';
+import { StylesContext } from '../styles';
 
 export default function useMaps() {
   const { mainId, comparedId } = useContext(MapsContext);
+  const { isCompared } = useContext(StylesContext);
```
danielfdsilva, line 8: nit: Do we want yo use context selectors (like the `useMapStyle`) or not worth it?

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -11,7 +11,7 @@ import {
 } from 'mapbox-gl';
```
danielfdsilva, line 11: Please remind me:
I recall us having a discussion about whether or not to make the STAC layer data requests outside the layer component. Did we end up deciding to keep it this way because of how tightly coupled with the rendering these requests are?

danielfdsilva (commented):
This is looking really good. And the layer usage is pretty clean!

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
danielfdsilva: Yes, that is correct

danielfdsilva (approved):

Issue #656, opened 2023-09-19
[Exploration page] Can we limit initial view to -180, 180 on Equirectangular projection


j08lue:
When we load datasets in Equirectangular projection in the Exploration page (like [this](https://www.earthdata.nasa.gov/dashboard/data-catalog/no2/explore?projection=equirectangular%7C%7C&basemapid=satellite&datetime=2020-01-01T00%3A00%3A00.000Z)) on a wide viewport, they get shown such that the Earth is shown several times:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/8f878566-982d-4a65-9dfc-5cb6fce81da0)

It would be great to instead be default zoom in enough such that the view is more like -180 to 180 degrees or so:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/43761ef8-bc92-4138-ab0e-4f3212011b9b)

Better making use of the screen and giving the user a nicer out of the box experience, also in this projection.

## Acceptance criteria

- [ ] Initial zoom level for global datasets on Equirectangular projection has been optimized

hanbyul-here:
I read this request as the initial landing viewport of the layer (which is configured to use equirectangular projection) needs to be limited to -180, 180? (Rather than the viewport should be restricted to -180, 180 whenever equirectangular projection is selected through interaction.) The first one can be configured through mdx (even though it looks like it needs a bit of code change), and the other one needs to be limited through code I believe. Can you confirm which approach is needed @j08lue?

j08lue:
> the initial landing viewport of the layer (which is configured to use equirectangular projection) needs to be limited to -180, 180?

Yes, it is the first case you mentioned (initial viewport).

PR #662, opened 2023-09-19
Remove check map bbox, add bbox validation

hanbyul-here:
As far as I can read, the code requires the new bound either to totally fit into the current bound or totally outside of the current bound. This blocks me from setting up the initial view as [-180, -90, 180, 90] for this issue: https://github.com/NASA-IMPACT/veda-ui/issues/656 (It looks like Mapbox's default viewport for equirectangular projection is somewhere [-190, -59, 190, 59]... pretty sure it changes depending on the screen size.) 

I don't have much background context but I wonder if we really need to check the current bound? (If not, we can just merge this change to `main` instead of `ghg` - and I will remove the function.) This PR skips this check and just validate the bbox against Mapbox's requirement.

### DIFF
--- a/app/scripts/components/common/mapbox/layers/utils.ts
+++ b/app/scripts/components/common/mapbox/layers/utils.ts
@@ -435,6 +435,17 @@ export function useLayerInteraction({
 type OptionalBbox = number[] | undefined | null;
 
 /**
+ * MapboxGL requires maxX value to be 180, minX -180, maxY 90, minY -90
+ * @param bounds Bounding box to fit layer
+ * @returns Boolean
+ */
+
+function isBboxToFitValid(bounds) {
+  const [minX, minY, maxX, maxY] = bounds;
+  return minX >= -180 && maxX <= 180 && minY >= -90 && maxY <= 90;
+}
+
+/**
  * Centers on the given bounds if the current position is not within the bounds,
  * and there's no user defined position (via user initiated map movement). Gives
  * preference to the layer defined bounds over the STAC collection bounds.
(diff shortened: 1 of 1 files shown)


j08lue:
> I don't have much background context but I wonder if we really need to check the current bound?

🤔

danielfdsilva:
If I remember correctly, we needed the check to recenter the viewport when changing layers. If the data changed from one layer (or date) to the other, the user shouldn't be left on an empty map (this would happen on sparse datasets). However most datasets are global and I can't find what prompted the original use case anymore. Perhaps is no longer needed.

danielfdsilva (commented):
@hanbyul-here I'm not sure I understand why this check is needed. The bounds don't ever seem to be greater than 180/90. If you switch between datasets with the equirectangular projection (when no user position is set on the url) it will work as expected - removing the `checkFitBoundsFromLayer` seem to be enough

Do you have some steps I can follow to see what was happening/what this you're trying to fix?

hanbyul-here:
@danielfdsilva STAC returns bbox going beyond -180, 180, -90,90 
ex: This dataset's bbox https://ghg-demo.netlify.app/data-catalog/eccodarwin-co2flux-monthgrid-v5/explore?projection=equirectangular%7C%7C&basemapid=satellite&datetime=2022-12-01T00%3A00%3A00.000Z
`("bbox":[-180.125,-90.12483,179.875,89.87517]})`

If we don't have a clear use case anymore, I'd like to propose merging this into main.

danielfdsilva:
@hanbyul-here oh, that's weird. so perhaps instead of checking if valid or not, we should clamp it to 180/90. There's no reason (afaik) to have bounds over that. what do you think?

hanbyul-here:
ok @anayeaye confirmed that there are some datasets that have invalid bbox, and the bbox is coming from the item in the dataset. @ividito added a validation for the extent https://github.com/NASA-IMPACT/veda-stac-ingestor/blob/main/api/src/schema_helpers.py#L34-L45 but but it looks like some datasets were directly ingested without validation. 🤔 
so I will say let's clamp the bbox for now.

danielfdsilva (approved):

PR #664, opened 2023-09-20
Allow customizing strings for data catalog and stories banners

nerik:
https://github.com/NASA-IMPACT/veda-config-ghg/issues/96


hanbyul-here (approved):
👍

PR #665, opened 2023-09-20
Adds option to hide dataset explore button

nerik:
- Adds a `disableExplore` parameter on the dataset frontmatter
- When set to true, the 'Explore data' button and the 'Overview/Exploration' submenu are not shown on the overview page
- When set to true, the `/[dataset]/exploration' route redirects to `/[dataset]`

Added for https://github.com/NASA-IMPACT/veda-config-ghg/issues/132

### DIFF
--- a/app/scripts/redirects.tsx
+++ b/app/scripts/redirects.tsx
@@ -2,6 +2,9 @@ import React from 'react';
 import { Navigate, Route, useParams } from 'react-router';
 
 import { STORIES_PATH } from '$utils/routes';
+import { useDataset } from '$utils/veda-data';
+import DatasetsOverview from '$components/datasets/s-overview';
+import DatasetsExplore from '$components/datasets/s-explore';
 
 function DiscoveryRedirect() {
   const { discoveryId } = useParams();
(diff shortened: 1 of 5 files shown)


app/scripts/redirects.tsx
```diff
@@ -52,3 +55,8 @@ export const discoveryRoutes = (
     <Route path='discoveries/:discoveryId' element={<DiscoveryRedirect />} />
   </Route>
 );
+
+export function DatasetExploreRedirect() {
+  const dataset = useDataset();
+  return dataset?.data.disableExplore ? <DatasetsOverview /> : <DatasetsExplore />;
```
danielfdsilva, line 61: I think this should be a redirect (`<Navigate replace ... />`) and not a conditional rendering. Otherwise the url `/explore` will render the overview.

---

For my curiosity: why have this check in an additional component, instead of the Explore page? Wouldn't it make more sense to look at the explore page code and see immediately what makes it behave as it does?

danielfdsilva (changes requested):

danielfdsilva (approved):

Issue #658, opened 2023-09-19
[Analysis page] Query for datasets via collection metadata?


j08lue:
Currently, we query for datasets available for the user-defined area and date range of interest by asking STAC for all items and then finding all collections.

This approach has several issues - it is costly and a lot of data gets transferred to the client that is not needed and it currently does not return all collections that should be returned, probably because not all items are loaded due to some limit / no pagination.

We have been discussing adding an aggregation endpoint to the STAC API / pgSTAC that could perform these queries in the database. However, also there, the issue remains that these queries are very costly and pgSTAC (unlike ElasticSearch) is not too fast for them.

An alternative solution is to make use of the total bounding box and date range information on the STAC collection level: STAC collection metadata already contains this information and we would just need to do the intersection in the client. While this approach is less accurate than the item query for edge cases where data coverage is sparse with large gaps, it is a lot faster and could at least limit the number of collections to query.

We will push for developing an aggregation function on our STAC backend, but that will take a while to develop. In the meantime, replacing the current approach by the fast collection metadata method would be great

## Acceptance criteria

- [ ] Tested whether collection metadata can be used to query for collections that cover area/time of interest

hanbyul-here:
I used the collections endpoint in https://github.com/NASA-IMPACT/veda-ui/pull/666. I think the main concern with this approach is that we can filter datasets only through their bbox, therefore spatially sparse datasets can have empty results. Check the preview and let me know what you think / if the filter can be better fine-tuned.

j08lue:
Wow, that turnaround was quick. 

I am sure we will hit the challenge with spatially (or temporally) sparse datasets eventually, but this solution is better than the current situation, at least for the GHG datasets. Rather show a bit too many datasets (and then have empty plots) than too few.

We need to make a few random tests and validate that the results are as expected. All datasets that (possibly) have _any_ data within the query should be listed.

To address the spatial case in the future, maybe we could compute the real coverage upon ingest (`union(existing_geom, new_geom)`) and store that in addition to the max bbox. 🤷

j08lue:
Done!

PR #666, opened 2023-09-20
Use collections endpoint instead of search for analysis 

hanbyul-here:
This PR uses `collections` endpoint to get all the collections, and filters them on the client side based on aoi/date range that user inputs. I followed the guidance provided in this issue: 

- https://github.com/NASA-IMPACT/veda-ui/issues/658

A few things to note
- I am not sure how item search works. Currently, the code catches all the datasets with the bbox that intersects with the AOI && the date domain that overlaps with the selected date range. 
- `collections` endpoint doesn't offer a detailed spatial extent. The bbox is a convex hull that includes all the data points. If a dataset is sparse like nightlight and plume, a user can see an empty chart like the screenshot below. @anayeaye and I talked and it might be helpful to have a flag to signal that this dataset is spatially sparse. (something similar to `is_periodic` but for spatial extent.) 

Screen Shot 2023-09-20 at 2 35 39 PM

@anayeaye thanks for your help 🙇 and let me know if you see anything unexpected! 

## Related issues

Supersedes / temporarily replaces https://github.com/NASA-IMPACT/veda-ui/pull/534


j08lue:
Not to overload this job here, but extra sugar for the user would be if a spatial query already limited the maximum range for the date selection - and possibly vice versa.

We can keep that for next time, though, unless you think it is very simple to add.

j08lue:
The query UI is already a lot more responsive. 🎉

hanbyul-here:
This PR is ready for review @danielfdsilva !

danielfdsilva (approved):

PR #668, opened 2023-09-22
Display selected datasets on the map

danielfdsilva:
Done so far:
- Displays selected datasets on the map
- Dataset order support
- Dataset opacity support

Compare is not possible yet.

### DIFF
--- a/app/scripts/components/common/map/utils.ts
+++ b/app/scripts/components/common/map/utils.ts
@@ -142,12 +142,10 @@ export function resolveConfigFunctions<T>(
   datum: T,
   bag: DatasetDatumFnResolverBag
 ): Res<T>;
-/* eslint-disable-next-line no-redeclare */
 export function resolveConfigFunctions<T extends any[]>(
   datum: T,
   bag: DatasetDatumFnResolverBag
 ): Res<T[number]>[];
-/* eslint-disable-next-line no-redeclare */
 export function resolveConfigFunctions(
   datum: any,
   bag: DatasetDatumFnResolverBag
--- a/app/scripts/components/common/map/utils.ts
+++ b/app/scripts/components/common/map/utils.ts
@@ -1,18 +1,22 @@
 import axios, { Method } from 'axios';
 import { Map as MapboxMap } from 'mapbox-gl';
 import { MapRef } from 'react-map-gl';
-import { endOfDay, startOfDay } from "date-fns";
-import { StacFeature } from "./types";
-import { userTzDate2utcString } from "$utils/date";
-import { validateRangeNum } from "$utils/utils";
+import { endOfDay, startOfDay } from 'date-fns';
+import {
+  DatasetDatumFn,
+  DatasetDatumFnResolverBag,
+  DatasetDatumReturnType
+} from 'veda';
 
+import { StacFeature } from './types';
+
+import { userTzDate2utcString } from '$utils/date';
+import { validateRangeNum } from '$utils/utils';
 
 export const FIT_BOUNDS_PADDING = 32;
 
 export const validateLon = validateRangeNum(-180, 180);
 export const validateLat = validateRangeNum(-90, 90);
-
-
 
 export function getMergedBBox(features: StacFeature[]) {
   const mergedBBox = [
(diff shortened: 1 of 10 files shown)


app/scripts/components/common/map/utils.ts
```diff
@@ -123,3 +122,63 @@ export async function requestQuickCache({
...
+type Res<T> = T extends Fn
+  ? T extends DatasetDatumFn<DatasetDatumReturnType>
+    ? DatasetDatumReturnType
+    : never
+  : T extends any[]
+  ? Res<T[number]>[]
+  : T extends object
+  ? ObjResMap<T>
+  : T;
+
```
nerik, line 140: Move to `types.d.ts` ?

app/scripts/components/common/map/utils.ts
```diff
@@ -123,3 +122,63 @@ export async function requestQuickCache({
...
+export function resolveConfigFunctions<T>(
+  datum: T,
+  bag: DatasetDatumFnResolverBag
+): Res<T>;
+/* eslint-disable-next-line no-redeclare */
```
nerik, line 145: No need for this AFAIK

app/scripts/components/common/map/utils.ts
```diff
@@ -123,3 +122,63 @@ export async function requestQuickCache({
...
+export function resolveConfigFunctions<T extends any[]>(
+  datum: T,
+  bag: DatasetDatumFnResolverBag
+): Res<T[number]>[];
+/* eslint-disable-next-line no-redeclare */
+export function resolveConfigFunctions(
```
nerik, line 151: TBH I'm struggling to understand what this does. `resolveConfigFunctions` can call itself if datum is iterable, that I get. but then in other situations, it calls datum as a function? How is that related to the call in index.tsx line 220? What does `bag` mean exactly? 

Sorry, might just be a case of Friday afternoon.

nerik (changes requested):
It works 🙌 

I guess my main comments are around `resolveConfigFunctions`, and also layer order:
1. the stacking order on the map is the opposite of the one in the timeline. As in, 'up' in the list should translate to 'on top' in the map, at least that's how I've always seen that kind of stuff work.
2. this is due to the sorting algorithm in styles.tsx, hence 100% my legacy, but I think we should either:
  - follow the natural order of generators in their array, and not explicitly declare order in the raster-timeseries signature (in that case the sorting logic should use the generators indices, not the id)
  - keep explicitly declaring order, but in that case use that numerical order directly in styles.tsx, rather than the string id. Also, in that case, no need to declare and use the order prop in raster-timeseries, but put it in the `BaseGeneratorParams` type and let styles.tsx handle this
  
This can be merged without addressing 2. which I can do myself later.

app/scripts/components/common/map/utils.ts
danielfdsilva: This basically resolves the functions in the dataset config and is needed to support this behavior.
The [docs](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md#function-values) show some examples and explain the behavior. If it is not clear, we should improve the docs.

This was used in the past for the customization of the compare map message. Although it is a neat little power user functionality, not sure it is used (known/needed) a lot.

danielfdsilva:
@nerik I think that 2.1 (layer's own index order) is what makes most sense.

app/scripts/components/common/map/utils.ts
danielfdsilva: These types are only ever used for the `resolveConfigFunctions` function declaration and are not needed elsewhere so I think they can live with the declaration. However feel free to move them if you like

danielfdsilva:
@nerik Did some cleanup to and moved the map code to a separate component.
The compare now can be enabled through the mock controls and shows the correct datasets with a date set 1 month from the main date.
If you want to address the layer order issue before merging, feel free to continue to work on this branch.

nerik:
@danielfdsilva Merged this for now. I'd like your eyes on [this](https://github.com/NASA-IMPACT/veda-ui/pull/668/commits/c055024679ba1be231d60cc00a453390a06a6b86) at some point if you have time.

Issue #660, opened 2023-09-19
[Analysis page] Zonal statistics jobs need guardrails for concurrency / number of requests


j08lue:
## Problem description

During demos, we by accident triggered a large zonal statistics computation job - probably by requesting zonal stats for a high time density dataset over a long time.

This took way too long for a user to wait for and also naturally touched a lot of data - as seen from these S3 read logs:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/3bca12c1-90c7-44d2-b7b3-1e33d9391ce0)

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/ff41e3fd-e762-43a9-9e1e-ee6d1bdf3bfd)

## Possible solution - guardrails / limits

As a user, I want to be guided to what size of analyses make sense to compute in the application, so I know what to use the tool for and what not.

Considering the coarse spatial but dense temporal resolution of most of our datasets, I think the issue is mostly to limit jobs in terms of number of time steps to analyze. But also large numbers due to large AOIs could be relevant to limit.

This could either be on the API side - the API throwing an error once it computed the number of items to touch and they exceed some limit - or by limiting the time range in the client. Allowed time range depends on the time density, though.

## Acceptance criteria

- [ ] A user cannot from the UI submit a job that triggers more than, say, 500 statistics computations / file reads.

vincentsarago:
@j08lue I feel we should set a limit on titiler-pgstac side so we don't try to open too many file

j08lue:
Yes, I think so, too. Failing in a way such that the client (frontend in this case) can recognize what is going on and tell the user.

j08lue:
We can still add guardrails in the frontend such that users ideally do not get to this point.

j08lue:
Ah, we actually calculate the number of assets in the frontend and then send all of these requests in parallel. 💡 I thought that was happening in the backend.

Yes, then this is just what we should do, we just need to have a different user flow around it, so it does not show as a failure once the user already sent off the request, but a warning / correction to their query, so they get good guidance on how to proceed from there.

I guess we only know the number of items to process once the user clicked `GENERATE`, right? Or do we already fire off requests to `https://staging-stac.delta-backend.com/search` before?

```json
    "context": {
        "limit": 10000,
        "matched": 745,
        "returned": 745
    },
```

j08lue:
Completed!

PR #671, opened 2023-09-22
Throw an error if too many stats are requested

hanbyul-here:
Close https://github.com/NASA-IMPACT/veda-ui/issues/660


hanbyul-here:
Closing as we decided to take other approaches. (guessing the # of requests based on the info from /collections endpoint)

PR #672, opened 2023-09-23
Fix pointer events on map controls

danielfdsilva:
The fact that this element takes up the full width was causing interaction problems. If you tried to draw on top of it, it was not possible.
image

This PR removes pointer events from the parents, including them only on the children.

@nerik This may need to be applied to the new map approach as well. I'd ask that you please check if it is needed.


nerik (approved):
LGTM, thanks!

PR #675, opened 2023-09-25
Limit the datasets for analysis

hanbyul-here:


hanbyul-here:
Heads-up that there might be some datasets that is not periodic without dates available through `summaries.datetime`. @anayeaye  explained this to me this happens when the dataset didn't go through the ingestion pipeline. (ex. put directly to database because of some issues at that time etc.) I think it is fine considering that not-periodic data is never that dense, but good to know.

PR #676, opened 2023-09-25
Check only day level equality for selected date

hanbyul-here:
The date gets set back when I try to change the date for Plume datasets: https://ghg-demo.netlify.app/data-catalog/emit-ch4plume-v1/explore?projection=mercator%7C%7C&basemapid=satellite&datetime=2023-07-29T10%3A06%3A30.000Z

This is because when a user inputs the date, the code checks if the date is in datetime list of the collection. Since Dashboard only provides daily level selection, selected datetime gets rounded down to the midnight of the selected date. Plume data datetimes don't have 00:00 timestamps, the selected date gets invalidated. I changed the code to check only day-level equality to align with dashboard capacity.

### DIFF
--- a/app/scripts/components/datasets/s-explore/index.tsx
+++ b/app/scripts/components/datasets/s-explore/index.tsx
@@ -8,6 +8,7 @@ import React, {
 import styled from 'styled-components';
 import { useNavigate } from 'react-router';
 import useQsStateCreator from 'qs-state-hook';
+import { isSameDay } from 'date-fns';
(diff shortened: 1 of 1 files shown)


hanbyul-here:
We are disabling the plume dataset on ghg, this fix doesn't seem to be needed :[

nerik (approved):
LGTM, thanks!
(should this be merged into `ghg`?)

hanbyul-here:
Yeah, I will make all the changes to `ghg` now so we can think of what to do altogether later.

PR #677, opened 2023-09-26
Feature/analysis flow

nerik:
Implements https://github.com/NASA-IMPACT/veda-config-ghg/issues/129


app/scripts/components/analysis/results/use-analysis-params.ts
```diff
@@ -25,8 +25,8 @@ type AnyAnalysisParamsKey = keyof AnalysisParams;
 type AnyAnalysisParamsType = Date | DatasetLayer[] | FeatureCollection<Polygon>;
 
 const initialState: AnalysisParamsNull = {
-  start: undefined,
-  end: undefined,
+  start: new Date(2018, 0, 1),
+  end: new Date(2022, 11, 31),
```
j08lue, line 29:
```suggestion
  end: new Date(2022, 12, 31),
```

j08lue (changes requested):
I see we now have a default geometry over North America, as we had discussed. I am not sure about this feature - the area is very large for a default and there could be political issues or so (you would probably need to include Hawaii, too, etc.).

So I would suggest we leave out this feature for now and leave the map blank.

app/scripts/components/analysis/results/use-analysis-params.ts
nerik: @j08lue No, in JS the date constructor takes a 0-indexed month.

j08lue: oops, sorry

j08lue: wait, what?!

nerik: See https://stackoverflow.com/questions/2552483/why-does-the-month-argument-range-from-0-to-11-in-javascripts-date-constructor

PR #673, opened 2023-09-23
Configurable metrics for analysis page

danielfdsilva:
Analysis metrics can now be selected on a chart basis.
image

The default metrics are configurable through the dataset mdx file under the `analysis.metrics` property of every layer.
image

Any invalid id id discarded.

@ricardoduplos Please have a quick design sanity check

PR #679, opened 2023-09-27
GHG -> main

nerik:
### Features

- Analysis define
  - New step-by-step layout and various UI tweaks
  - Added sticky footer, moved submit and cancel buttons there
  - Added an AoI preset for North America
  - Datasets are not filtered from `/collections` metadata (spatial, temporal extent)
  - Dates are now set by default to 2018 - 2022
  - Replaced date presets with "2018 - 2022" and "last 10 years"
  - Date selection not possible anymore before 01-01-1980
  - Prevent users from selecting datasets when time range selected would mean fetching too many STAC items
- Analysis results page: set a hardcoded limit on the number of items that can be requested as a failsafe mechanism for ☝️  

### Bugfixes

- Removed logs in production (https://github.com/NASA-IMPACT/veda-config-ghg/issues/152)
- Temporary fix of critical error encoutered during demos (https://github.com/NASA-IMPACT/veda-ui/pull/680) 
- Remove Shadow Scroll as a preventative method for (https://github.com/NASA-IMPACT/veda-ui/pull/691)
- Check only day level equity to validate dates (https://github.com/NASA-IMPACT/veda-ui/pull/676)
- Remove unneded attributes (https://github.com/NASA-IMPACT/veda-ui/pull/694)
- BlockMap initial position error (https://github.com/NASA-IMPACT/veda-ui/pull/687)

### Copy changes
- [Updated define copy](https://github.com/NASA-IMPACT/veda-ui/pull/679/commits/f1e7171c64e8198772ee2c9848e4cb00f9ab7fb9)
- [Disclaimer on results page](https://github.com/NASA-IMPACT/veda-ui/pull/679/commits/54cf01ed8d523c606e7f5b9f11683dcfc6408bc5)

--- 

### Added in veda UI but made configurable (so not affecting veda-config):
- https://github.com/NASA-IMPACT/veda-ui/pull/673
- https://github.com/NASA-IMPACT/veda-ui/pull/665
- https://github.com/NASA-IMPACT/veda-ui/pull/664

### Bugfixes
- https://github.com/NASA-IMPACT/veda-ui/pull/672
- https://github.com/NASA-IMPACT/veda-ui/pull/670 / https://github.com/NASA-IMPACT/veda-ui/pull/666
- https://github.com/NASA-IMPACT/veda-ui/pull/662
- https://github.com/NASA-IMPACT/veda-ui/pull/691
- https://github.com/NASA-IMPACT/veda-ui/pull/650
- https://github.com/NASA-IMPACT/veda-ui/pull/676
- https://github.com/NASA-IMPACT/veda-ui/pull/694
- https://github.com/NASA-IMPACT/veda-ui/pull/687


danielfdsilva (changes requested):
I pushed a commit with some css fixes. Mostly it rounds the button corners on the analysis page given that it was clashing with the remaining buttons on the same page. If the idea is to move to square buttons, then a overall change is required.

I'd like to us to understand if the ShadowScrollbar really is a source of problems and performance issues, given that it adds a shinier finish to the app.

Without it, the dropdowns look a bit like an afterthought.
image

Besides this I didn't see anything that would be GHG specific.

hanbyul-here:
That is a good point about how the dropdown looks, so I added `overscroll-behaviour: none` to give a better look. I'd like to know the reason why ShadowScroll was making so many commits, too. But in any case, it is ideal not to depend on Javascript for scroll styling since it is prone to over-rendering.

danielfdsilva:
@hanbyul-here My comment was related to styling, rather than behavior.
I'm not sure why it was causing so many commits just on open. That's definitely worth investigating.
On usage, it does some [computation to display the shadows](https://github.com/developmentseed/ui-library-seed/blob/develop/packages/shadow-scrollbar/src/index.js#L112-L154) but that should not happen when it is opened.

Unfortunately there's no other way to style a scrolling element besides javascript. In this case we also want the shadow indicating that there's scrollable content which is not doable with css alone.
I think it is ok to degrade the visual experience a bit in favor of performance and not having errors, but I'd just lie us to be sure that this is actually causing problems.

hanbyul-here:
As far as I can read, ShadowScrollbar attaches extra html to generate the shadow gradient in the background. If we think it is necessary, we can still make that effect with additional markup and styles. It won't be as sleek as the ShadowScroll, but there are some ways that we can style scrolls so they look like more thoughts went into them. (I made the style change, then I thought it was better to apply it more globally so I reverted the style.)

But all in all, you agree that the visual degradation is acceptable, then can you clarify what changes you are requesting?

danielfdsilva:
I just wanted us to understand if ShadowScrollbar really is the culprit behind the errors that we were seeing and why that was happening.
Given that its use gives the product a shinier finish I'd like us to be sure of any incompatibility before removing it completely.
I think that the extra elements that it adds and the calculations it performs are a perfectly acceptable thing to have - unless it really is breaking the app.

However I do understand that there may not be time to track down this error, and this (together with removing the redirects), _seems_ to solve the problem. I'll leave this to your decision.

danielfdsilva (approved):

hanbyul-here:
I will merge this, and still have a discussion about how we can improve style of scroll. nudging @faustoperez

PR #680, opened 2023-09-27
Temporary fix of critical error encoutered during demos

hanbyul-here:
openign this to investigate, do not merge this yet
@amarouane-ABDELHAK  I do think my change here achieves the same effect that you are trying to do. (not using special routting, but treat the tag like normal a tag, so the page gets reloaded everytime.) 
lmk how you think.

### DIFF
--- a/app/scripts/components/common/page-header.tsx
+++ b/app/scripts/components/common/page-header.tsx
@@ -60,6 +60,11 @@ const PageHeaderSelf = styled.header`
   }
 `;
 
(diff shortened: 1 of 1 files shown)


xhagrg (approved):

amarouane-ABDELHAK (commented):
The solution looks good

hanbyul-here:
I can't understand why lint checks fail while there is no error. 🤔 I will merge and see if this still happens on `ghg` branch.

Issue #682, opened 2023-09-28
App sporadically crashes


j08lue:
The issue is a bit mysterious and transient at this point. The app crashes sometimes when a new page is loaded, e.g. via click on another menu item.

Forcing reload fixes it, as we do in 

- https://github.com/NASA-IMPACT/veda-ui/pull/680

But that degraded the user experience because the page reloads completely when switching between menu items.

Further investigation, e.g. via Sentry, is needed to get to the root of the issue

- https://github.com/NASA-IMPACT/veda-ui/pull/681

## Acceptance criteria

- [ ] Diagnosed and fixed the root cause of app crashes on page loads

nerik:
Currently trying to work our way back to what could be causing the problem in the GHG instance as this wasn't happening in `veda-config`.
https://github.com/NASA-IMPACT/veda-ui/issues/682

hanbyul-here:
I used the time after the demo to get to the root of this problem. 

The problem seemed to be introduced after we put the logic for redirections using `<Navigate />` component of React Router. Through React profiler, we could find that `ShadowScrollbar` is making way too many [commits](https://react.dev/learn/render-and-commit). (> 300 commits for scrolling on the page + clicking on the link on the nav.)  We got rid of the redirection logic to resolve this issue. (The redirection logic was to prevent navigating to /explore page from the datasets that disable the dataset exploration/.  which I can be a nice touch, but not really necessary.) 

![Screen Shot 2023-10-04 at 1 03 48 PM](https://github.com/NASA-IMPACT/veda-ui/assets/4583806/6a8e49ab-fcc8-4a1f-bfe2-25bc6f8d45a0)

Noting that I am not sure what part of redirection exactly triggered the unnecessary amount of `ShadowScrollbar`. However, redirection might have amplified the problem, but not be the root of the problem. For example, `ShadowScrollbar` makes way too many commits when it is active on the page still. (just opening nav makes more than 100 commits.)

https://github.com/NASA-IMPACT/veda-ui/assets/4583806/cb813180-5255-441c-9c75-b664e9048b98

I will go ahead and edit or get rid of `ShadowScrollbar` to prevent further breaks.

PR #685, opened 2023-09-28
Hacky patch against crashes

xhagrg:
Add event listener for every anchor tags to refresh the page.

Attempt at fixing https://github.com/NASA-IMPACT/veda-ui/issues/682

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -134,10 +134,12 @@
     }
 
     document.addEventListener('click', function (event) {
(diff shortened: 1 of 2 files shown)


j08lue:
Before we go down this route much deeper, can we try to see whether the other fix does the trick, pls? https://github.com/NASA-IMPACT/veda-ui/pull/684

PR #687, opened 2023-09-28
Do not pass initial position at all if there is none

hanbyul-here:
fix the bug where we pass `initialPosition` prop even when there is none defined by an editor.


nerik:
@hanbyul-here Do you need a review here? Should this be merged into `ghg` ?

hanbyul-here:
yeah thank you @nerik . 
Context: This bug was preventing to use default spatial extent for BlockMap, I worked around it by setting a zoom, center in this pr : https://github.com/NASA-IMPACT/veda-config-ghg/pull/183/files

nerik (approved):

PR #690, opened 2023-10-04
Basic AoIs

nerik:
Prototypes a simple AoI control working with the new map. This includes storing the AoI(s) in the URL.


app/scripts/components/exploration/atoms/atoms.ts
```diff
@@ -1,11 +1,15 @@
 import { atom } from 'jotai';
```
danielfdsilva, line 1: Should the atoms used directly by the map (a common component) be defined together with the exploration atoms?

app/scripts/components/exploration/atoms/atoms.ts
```diff
@@ -43,3 +47,55 @@ export const activeAnalysisMetricsAtom = atom<DataMetric[]>(dataMetrics);
...
+});
+
+// Setter atom to delete AOIs, writing directly to the hash atom.
+export const aoisDeleteAtom = atom(null, (get, set, ids: string[]) => {
+  const features = get(aoisFeaturesAtom);
+  const newFeatures = features.filter(
+    (feature) => !ids.includes(feature.id as string)
+  );
+  set(aoisHashAtom, encodeAois(newFeatures));
+});
```
danielfdsilva, line 101: This is an interesting paradigm. You have atoms that just update the main atom. Is this a normal jotai paradigm?

app/scripts/utils/polygon-url.ts
```diff
@@ -47,10 +60,32 @@ export function polygonUrlEncode(
...
+      return [...acc, decodedFeature];
+    } else {
+      const lastFeature = acc[acc.length - 1];
+      const prop = i % 3 === 1 ? 'id' : 'selected';
+      const newFeature = { ...lastFeature, [prop]: current };
+      acc[acc.length - 1] = newFeature;
+      return acc;
+    }
+  }, []);
+  return features!;
```
danielfdsilva, line 90: I like how you stored the data in the array, however I found this piece of code a bit tricky to parse.
A possible approach would be to chunk the array in groups of 3 and handle it that way.

Just a suggestion, not a blocker.
```tsx
import { chunk } from 'lodash';

export function decodeAois(aois: string): AoIFeature[] {
  const decoded = JSON.parse(aois) as string[];

  const features: AoIFeature[] = chunk(decoded, 3).map((data) => {
    const [polygon, id, selected] = data;
    const decodedFeature = decodeFeature(polygon) as AoIFeature;
    return { ...decodedFeature, id, selected };
  });

  return features;
}
```

danielfdsilva (commented):
A couple of things I'd suggest maybe adding/changing on this PR:
- the cursor over the map when the drawing is active
- disabling the trash can when nothing can be deleted
- jotai location uses hash to store the values. Query parameters are better suited for this purpose - can we use that instead?

app/scripts/components/exploration/atoms/atoms.ts
nerik: Probably not. That's a good point, I'll move them.

app/scripts/components/exploration/atoms/atoms.ts
nerik: Is this normal? 😅 
They are called _derived atoms_ in Jotai. In my understanding, they extend the concept of _selectors_ that exist in Redux or Recoil or many others, except they are not necessarily read-only (they can be write-only like here or read-write). The main idea is to use the URL hash as the single-source-of-truth, then provide a getter to get an array of AoIs, and setters to manipulate it. 
I considered another alternative, which was to have a single read-write derived atom, then do various write operations in React land, but this turned out to be more concise. Happy to discuss if you wanna suggest alternatives.

app/scripts/utils/polygon-url.ts
nerik: @danielfdsilva I had no idea `chunk` existed. You made my day ❤️

app/scripts/components/exploration/atoms/atoms.ts
danielfdsilva: Just wondering about this. It is an interesting approach! :)

danielfdsilva (approved):

PR #691, opened 2023-10-04
Remove shadow scroll

hanbyul-here:
Handles issue in https://github.com/NASA-IMPACT/veda-ui/issues/682#issuecomment-1747334303

While putting this fix, I noticed that `use-media-query` returns document body size, which differs from how CSS media query works. This results in a weird style in between 991 px ~ 1000 px. (screenshot attached, it is on ghg.center now)  👉  I didn't fix this problem, but I patched it with CSS.

(Eventually, we would need to move to Container query instead of media query: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) 

Screen Shot 2023-10-04 at 3 01 10 PM


app/scripts/components/common/shadow-scrollbar-improved.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import { ShadowScrollbar } from '@devseed-ui/shadow-scrollbar';
 
+// Moving away from this component because of the issue
+// https://github.com/NASA-IMPACT/veda-ui/issues/682#issuecomment-1747334303
```
nerik, line 5: I think you can go ahead and remove the file. Juste use this comment as the commit message.

app/scripts/components/common/mapbox/layer-legend.tsx
```diff
@@ -252,17 +251,11 @@ export function LayerLegend(
...
       renderBody={() => (
         <LegendBody>
-          <ShadowScrollbar
-            scrollbarsProps={{
-              autoHeight: true,
-              autoHeightMin: 32,
-              autoHeightMax: 120
-            }}
-          >
+
```
nerik, line 254: Text is cut off here:

Screenshot 2023-10-05 at 14 38 07

On `data-catalog/sedac-popdensity-yeargrid5yr-v4.11`

nerik (changes requested):
Looking good, thanks 👍 
Please review the issue with the map legend.

nerik (approved):

hanbyul-here:
Ah looks like I missed the case when there are many layers. working on the fix now. (for PanelBody)

Screen Shot 2023-10-05 at 2 11 46 PM

PR #698, opened 2023-10-12
Include scrollbar width in media calculation

danielfdsilva:
Turns out that the difference in the returned values was due to the fact that CSS includes the scrollbar width in the media size calculation, and the hook wasn't doing that.

Should not impact the app in any negative way, but it is worth a quick check.


hanbyul-here (approved):
React Cool Deimensions seems to be written specifically for container query, not for media query based on their readme. (https://www.npmjs.com/package/react-cool-dimensions#features) It might be worth looking at other libs/ if we are not using its container query feature in the future. But this is for the future, not a blocker.

PR #699, opened 2023-10-12
Add analysis data request with AOIs

danielfdsilva:
Implements the analysis flow, using the aois drawn on the map, requests data and displays it on the chart.
Also adds support for each dataset to have independent customizable metrics.


j08lue:
Amazing to see this working! 😍 

Probably premature comment - the AOI is very hard to see e.g. on top of GEOGLAM...

image

https://deploy-preview-699--veda-ui.netlify.app/exploration?aois=%5B%22wqk%7D%40uxueIzv%7CMr%7B%7CDauV%60_zG%7BloGddcHsyiKnfpBoydDxqAhw%7DAydzG_cgHob%60A%7BrtD%7BpiDjsZeaoEpydDaw%7CB%7EryKwbM%22%2C%22b8fcdf%22%2Ctrue%5D

app/scripts/components/exploration/hooks/use-analysis-data-request.ts
```diff
@@ -0,0 +1,98 @@
...
+  const selectedInterval = useAtomValue(selectedIntervalAtom);
+  const { features } = useAois();
+  const selectedFeatures = features.filter((f) => f.selected);
+
+  const { analysisRun } = useAnalysisController();
+
+  const dataset = useAtomValue(datasetAtom);
+  const datasetStatus = dataset.status;
+
+  const [, setAnalysis] = useTimelineDatasetAnalysis(datasetAtom);
```
nerik, line 63: You could have `useTimelineDatasetAnalysis` only return a setter (`useSetAtom`) and simplify this:
```suggestion
  const setAnalysis = useTimelineDatasetAnalysis(datasetAtom);
```

nerik (commented):
Looking good so far. The analysis state should probably be reset on deleting polygons. We aslo should migrate at some point the logic implemented [here](https://github.com/NASA-IMPACT/veda-ui/pull/677/commits/35bd7fadd2714cfa34dff03eda75f0f912af7b43) to limit the potentially very high number of stac items requested.

app/scripts/components/exploration/hooks/use-analysis-data-request.ts
danielfdsilva: that's a good point. I was trying to be consistent with the other `useTimelineDataset*`, like `useTimelineDatasetVisibility` or `useTimelineDatasetSettings` where they return getters and setters. You think it is not needed?

nerik: The consistency argument is a reasonable one, but in that case why not just returning the atoms themselves?

```
export function useTimelineDatasetAnalysisAtom() {
  const analysisAtom = ...

  return analysisAtom;
}
```

```suggestion
  const setAnalysis = useSetAtom(useTimelineDatasetAnalysisAtom(datasetAtom));
```

danielfdsilva: I was trying to avoid this function train and make the hook usage more direct. Nevertheless I'm not set on this format and I'm not opposed to changing it. Do you have a preference?

hanbyul-here (changes requested):
+1 on @nerik 's comment that we need to put a guardrail to prevent excessive amount of requests.

When I was checking, Nightlight dataset was returning errors on some of the items. (Not sure if it is something that can be reproduced consistently or a hiccup.) And a dataset list item doesn't reflect the error, gets stuck at the loading state. (As far as I can read, it seems  error state gets washed out from the other request's loading state?)  
Screen Shot 2023-10-16 at 3 04 49 PM

hanbyul-here (approved):

danielfdsilva:
@hanbyul-here Thank you. It is Draft since still need to address the message with @ricardoduplos . But the functionality can be reviewed :)

ividito:
>As for Nightlights: it sometimes errors seemingly randomly. It seems that if there are a lot of requests it errors (although we're throttling at 15 simultaneous). Perhaps it has to do with api load or something. @ividito any debugging we can do here?

I looked into this and it looks like a handful of nightlights "regions" will cause excessive loading times, which can cascade into the service timing out. I think the newest version of the preview handles this a bit better - I still see longer loading times in some areas, but it's no longer causing an actual error for me.

This is probably something we can correct on the backend - I'll make an issue to track this in that repo.

danielfdsilva:
@ricardoduplos Messages implemented according to the design. Please have a look!

ricardoduplos (approved):
Looks good!

PR #700, opened 2023-10-13
Update titler-xarray-endpoint to production

abarciauskas-bgse:
There are now 2 deployments of titiler-xarray: one production endpoint and one dev. Dev will be reserved for new feature development and testing, so we should use the production endpoint in VEDA dashboard services. I will make an equivalent PR to veda-config.

### DIFF
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ APP_CONTACT_EMAIL=email@example.org
 
 API_RASTER_ENDPOINT='https://staging-raster.delta-backend.com'
 API_STAC_ENDPOINT='https://staging-stac.delta-backend.com'
-API_XARRAY_ENDPOINT='https://dev-titiler-xarray.delta-backend.com/tilejson.json'
+API_XARRAY_ENDPOINT='https://prod-titiler-xarray.delta-backend.com/tilejson.json'
 
 # If the app is being served in from a subfolder, the domain url must be set.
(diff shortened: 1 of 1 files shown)


hanbyul-here:
While we are at it, I want to flag that the example data you put (`oco2-geos-l3-daily`) is missing spatial and temporal extent. which leads to the error on the page. (I noticed this issue a while ago, so I believe it is unrelated to changing the endpoint to production.)

- Are spatial and temporal extent intentionally missing?
- If so, is there a better example zarr data to put? since `oco2-geos-l3-daily` is missing the values that dashboard needs?

abarciauskas-bgse:
@hanbyul-here thanks for looking, tagging @slesaad who published that dataset. @slesaad should we fix or remove this dataset? I can look into doing so if it would help.

danielfdsilva:
@abarciauskas-bgse Should we change this in the veda-ui, given that this is never used directly? Wouldn't it make more sense to keep the dev endpoint here? Is it unstable?

abarciauskas-bgse:
🤔 The intention to have dev and prod deployment is that dev might be unstable. I do think it could be useful to leave the configuration here pointing to dev so we can test new features' in titiler-xarray are still working with VEDA UI. So yes I think it makes sense to close this for now.

PR #703, opened 2023-10-17
Custom AoI upload

nerik:
Migrates the previous upload custom AoI modal to the Exploration page.

Screenshot 2023-10-17 at 15 12 35

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
@@ -1,0 +1,52 @@
+import React, { useState } from 'react';
+import { Feature, Polygon } from 'geojson';
+import { CollecticonUpload2 } from '@devseed-ui/collecticons';
+import { Button, createButtonStyles } from '@devseed-ui/button';
+import styled from 'styled-components';
+import { themeVal } from '@devseed-ui/theme-provider';
+import useThemedControl from '../hooks/use-themed-control';
+import CustomAoIModal from '../custom-aoi-modal';
+
+const SelectorButton = styled(Button)`
+  &&& {
+    ${createButtonStyles({ variation: 'primary-fill', fitting: 'skinny' })}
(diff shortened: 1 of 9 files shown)


app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -0,0 +1,52 @@
...
+function CustomAoI({ onConfirm }: CustomAoIProps) {
+  const [aoiModalRevealed, setAoIModalRevealed] = useState(false);
+  return (
+    <>
+      <SelectorButton>
+        <CollecticonUpload2
+          title='Upload geoJSON'
+          meaningful
+          onClick={() => setAoIModalRevealed(true)}
+        />
```
danielfdsilva, line 36: The click listener should be on the button
```suggestion
      <SelectorButton onClick={() => setAoIModalRevealed(true)}>
        <CollecticonUpload2 title='Upload geoJSON' meaningful />
```

danielfdsilva (changes requested):
Maybe not part of this PR, but could be good to allow the user to clear all the AOIs by pressing the trash can with nothing selected.

nerik:
@danielfdsilva Good suggestion. I don't see any other way to implement this (ie adding a custom listener on the trash button) than to rebuild the control from scratch, thoughts?

danielfdsilva:
Yeah... seems to be that way. I've rarely used the default controls. For the current analysis I had [created custom ones](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/analysis/define/aoi-selector.tsx) and then [handled the events](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/aoi/use-aoi-controls.ts) as I needed.

danielfdsilva:
There could be also an additional button to do that, maybe it is also clearer? like a reset?
The current analysis uses the arrows one:
image

PR #704, opened 2023-10-17
Dataset exploration comparison

danielfdsilva:
Implements the UI controls needed for the comparison


nerik:
There was a critical bug that prevented the compare map to show a different set of layers than the base ones. [Fixed here](https://github.com/NASA-IMPACT/veda-ui/pull/704/commits/fc7ffcc7c360fc98f0c80f6b5960fcd38f8d1427), but I'm not entirely convinced by my fix.

PR #706, opened 2023-10-18
[#181] Update default sorting to pubDate if exists

sandrahoang686:
Issue/Ticket [Link](https://github.com/NASA-IMPACT/veda-config-ghg/issues/181)

Description
Updates the default sorting for stories from `name` to `pubDate`

### DIFF
--- a/app/scripts/components/common/browse-controls/use-browse-controls.ts
+++ b/app/scripts/components/common/browse-controls/use-browse-controls.ts
@@ -48,7 +48,7 @@ export function useBrowserControls({ sortOptions }: BrowseControlsHookParams) {
   const [sortField, setSortField] = useQsState.memo(
     {
       key: Actions.SORT_FIELD,
-      default: sortOptions[0].id,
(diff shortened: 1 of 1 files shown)


app/scripts/components/common/browse-controls/use-browse-controls.ts
```diff
@@ -48,7 +48,7 @@ export function useBrowserControls({ sortOptions }: BrowseControlsHookParams) {
   const [sortField, setSortField] = useQsState.memo(
     {
       key: Actions.SORT_FIELD,
-      default: sortOptions[0].id,
+      default: sortOptions[1]?.id || sortOptions[0]?.id, // If pubDate exists, default sorting to this
```
danielfdsilva, line 51: Since the order of values on `sortOptions` may change and we're looking to have a specific default rather than the 1st, this could be made a bit more resilient.

```suggestion
      // If pubDate exists, default sorting to this
      default: sortOptions.find((o) => o.id === 'pubDate')?.id || sortOptions[0]?.id,
```

danielfdsilva (changes requested):

danielfdsilva (approved):

PR #707, opened 2023-10-20
Reset AoIs

nerik:
This adds a reset button as suggested [here](https://github.com/NASA-IMPACT/veda-ui/pull/703)

Unfortunately, I couldn't get this to work. The control needs access to the map ref, which is stored in the MapsContext (reachable via the `useMaps` hook), but `useThemedControls` creates a separated React root, hence without access to any context.

I've been looking for hours for a clean solution but I'm stuck, sorry :(

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/reset-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/reset-aoi-control.tsx
@@ -2,9 +2,11 @@ import React, { useCallback } from 'react';
 import { CollecticonArrowLoop } from '@devseed-ui/collecticons';
 import { Button, createButtonStyles } from '@devseed-ui/button';
 import styled from 'styled-components';
+import { useSetAtom } from 'jotai';
 import { themeVal } from '@devseed-ui/theme-provider';
 import useThemedControl from '../hooks/use-themed-control';
 import useMaps from '../../hooks/use-maps';
+import { aoiDeleteAllAtom } from './atoms';
 
 const SelectorButton = styled(Button)`
   &&& {
@@ -20,11 +22,13 @@ const SelectorButton = styled(Button)`
 `;
 
 function ResetAoI({ map }: { map: any }) {
+  const aoiDeleteAll = useSetAtom(aoiDeleteAllAtom)
   const onReset = useCallback(() => {
-    const mbDraw = map?.instance?._drawControl;
+    const mbDraw = map?._drawControl;
     if (!mbDraw) return;
-    mbDraw.trash();
-  }, [map]);
+    mbDraw.deleteAll();
+    aoiDeleteAll();
+  }, [map, aoiDeleteAll]);
 
   return (
     <>
--- a/app/scripts/components/common/map/controls/aoi/reset-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/reset-aoi-control.tsx
@@ -1,0 +1,51 @@
+import React, { useCallback, useContext, useEffect } from 'react';
+import { CollecticonArrowLoop } from '@devseed-ui/collecticons';
+import { Button, createButtonStyles } from '@devseed-ui/button';
+import styled from 'styled-components';
+import { themeVal } from '@devseed-ui/theme-provider';
+import useThemedControl from '../hooks/use-themed-control';
+import useMaps from '../../hooks/use-maps';
+
+const SelectorButton = styled(Button)`
+  &&& {
+    ${createButtonStyles({ variation: 'primary-fill', fitting: 'skinny' })}
+    background-color: ${themeVal('color.surface')};
+    &:hover {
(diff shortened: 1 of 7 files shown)


app/scripts/components/common/map/controls/aoi/reset-aoi-control.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useThemedControl(
+    () => {
+      return <ResetAoI />;
+    },
+    {
+      position: 'top-left'
+    }
+  );
+  return null;
+}
```
danielfdsilva, line 51: Would it be possible to access the context here and pass it as a prop to `ResetAoI`?

nerik: Great suggestion: https://github.com/NASA-IMPACT/veda-ui/pull/707/commits/d102d801642e909b69ea3f88ea16b04a9559a29b

However `instance` doesn't exist on this map ref for some reason.

I could create a specific context for this map implementation in order for the draw control to expose its internals for other draw-related controls, but this really feels like a hack.

danielfdsilva: Where is `instance` coming from?

danielfdsilva: The previous `ref` implementation had an `instance` because we added it since it served our needs.

https://github.com/NASA-IMPACT/veda-ui/blob/1bb5c5efe94767c0b4e3a55387e8e22c024119db/app/scripts/components/common/mapbox/index.tsx#L168-L175

And then the map stored the draw control so we can act on it, but none of this is default.

https://github.com/NASA-IMPACT/veda-ui/blob/1bb5c5efe94767c0b4e3a55387e8e22c024119db/app/scripts/components/common/mapbox/aoi/mb-aoi-draw.js#L32-L40

danielfdsilva: A solution for this implementation could be to add a reference to the control when it is mounted:

`@app/scripts/components/common/map/controls/aoi/index.tsx`
```diff
useControl<MapboxDraw>(
-    () => {
+    ({map}) => {
      control.current = new MapboxDraw(props);
+      map._drawControl = control.current;
      return control.current;
    },
```

Then you could access it via `map._drawControl` on the reset button

danielfdsilva: However for the clear button I don't think you need to access the `trash` method. Using this without a selection does nothing.
I'd imagine that deleting all the feature via the atom would do the trick.

nerik: I've used `deleteAll`: https://github.com/NASA-IMPACT/veda-ui/pull/707/commits/d102d801642e909b69ea3f88ea16b04a9559a29b, it works!
> I'd imagine that deleting all the feature via the atom would do the trick.
Sadly no, you have to manually keep the atom and the mbdraw state in sync

I have taken the opportunity to clean up that flow with the upload AoI feature: https://github.com/NASA-IMPACT/veda-ui/pull/707/commits/66a9f32ebcee06e41f3bf7ae8da1f16f31c0b024

danielfdsilva:
I had a brief conversation with Erik at the end of last week and this is at a point where it is working.
However the result is that this implementation relies both on Mapbox draw imperative methods and react reactive code.

For example:
https://github.com/NASA-IMPACT/veda-ui/blob/2f1e5f016630574c0771888b7db3a549313557b2/app/scripts/components/common/map/controls/aoi/reset-aoi-control.tsx#L25-L30

Here, the features are being deleted from the draw plugin and at the same time from the atom that stores them.

There no absolutely clean way to do this though. We are working with a mapbox plugin that is imperative by nature.
A way to reduce this problem is to create custom controls for the deletion and then based on those set the features on the plugin. (deleting would become setting and empty feature array).
This was the approach used in the current analysis. The [controls](https://github.com/NASA-IMPACT/veda-ui/blob/2f1e5f016630574c0771888b7db3a549313557b2/app/scripts/components/analysis/define/aoi-selector.tsx) fire AOI events that are [captured and processed](https://github.com/NASA-IMPACT/veda-ui/blob/2f1e5f016630574c0771888b7db3a549313557b2/app/scripts/components/common/aoi/use-aoi-controls.ts#L32-L127).

This same behavior could be migrated to this implementation but it would require some additional work and changes. Do you have any thoughts @hanbyul-here @sandrahoang686? Do you feel that this is worth it?
cc @j08lue

hanbyul-here:
We would eventually want to have one place that syncs these states, but I think it is acceptable to file this as a tech debt and move on for now.

danielfdsilva:
Rebased to fix the conflict and logged the tech debt.
Will merge.

PR #708, opened 2023-10-23
Store selected exploration datasets on url

danielfdsilva:
By storing the selected datasets in the url we allow the user to share the page state, although for it to be truly shareable there are other things that also need to be stored, which is not the main purpose of this PR.

The main goal here it to be able to give the user an entry point to the exploration page, when coming from the dataset catalog - not implemented in this PR

### DIFF
--- a/app/scripts/components/exploration/atoms/atoms.ts
+++ b/app/scripts/components/exploration/atoms/atoms.ts
@@ -1,11 +1,87 @@
 import { atom } from 'jotai';
+import { atomWithLocation } from 'jotai-location';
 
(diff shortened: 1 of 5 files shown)


hanbyul-here:
I am not sure why, but the preview is broken: https://deploy-preview-708--veda-ui.netlify.app/exploration (I don't see this problem on my local.)

danielfdsilva:
@hanbyul-here @sandrahoang686 Rebased this on top of the current exploration so the error is gone. Should be ready to review.

hanbyul-here (approved):

PR #709, opened 2023-10-24
Fix critical error on /exploration

danielfdsilva:
The error identified in https://github.com/NASA-IMPACT/veda-ui/pull/708 was introduced by the commit 6721546004ff38bbac40eb9e44e3a6c57342ba47.

https://github.com/NASA-IMPACT/veda-ui/blob/bd3440cafe950d335e31543eefa91512ae04fa42/app/scripts/components/common/map/maps.tsx#L95-L103

The code here uses the `name` property (which is derived from the function name) to understand what to do with the component being rendered. The problem is that in a `production` environment, the code is minified and function names change making this property unreliable and unpredictable.

The solution I found was to use the `displayName` property that react allows us to attach to components to give them some names. This is a paradigm we already use in the app with the blocks for the scrollytelling ([eg1](https://github.com/NASA-IMPACT/veda-ui/blob/0a84a5e14d2978aca66bb180304a6412c4368e84/app/scripts/components/common/blocks/figure.js), [eg2](https://github.com/NASA-IMPACT/veda-ui/blob/0a84a5e14d2978aca66bb180304a6412c4368e84/app/scripts/components/common/blocks/images/index.js))

**The only thing to keep in mind is that if we create a new map control, we have to give it a `displayName`.**

### DIFF
--- a/app/scripts/components/common/map/index.tsx
+++ b/app/scripts/components/common/map/index.tsx
@@ -2,9 +2,19 @@ import React, { ReactNode } from 'react';
 import { MapProvider } from 'react-map-gl';
 import Maps, { MapsContextWrapperProps } from './maps';
 
+export const COMPARE_CONTAINER_NAME = 'CompareContainer';
 export function Compare({ children }: { children: ReactNode }) {
   return <>{children}</>;
 }
+
+Compare.displayName = COMPARE_CONTAINER_NAME;
+
+export const CONTROLS_CONTAINER_NAME = 'MapControlsContainer';
+export function MapControls({ children }: { children: ReactNode }) {
+  return <>{children}</>;
+}
+
+MapControls.displayName = CONTROLS_CONTAINER_NAME;
 
 export default function MapProviderWrapper(props: MapsContextWrapperProps) {
   return (
--- a/app/scripts/components/common/map/maps.tsx
+++ b/app/scripts/components/common/map/maps.tsx
@@ -3,7 +3,6 @@ import React, {
   Children,
   useMemo,
   ReactElement,
-  JSXElementConstructor,
   useState,
   createContext
 } from 'react';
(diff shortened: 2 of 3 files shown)


hanbyul-here:
I feel like I am missing the bigger context. If we need to sort out the components based on their display names in the background, why are we just passing them as explicit props?

danielfdsilva:
@hanbyul-here all the components are added as [children of Map](https://github.com/NASA-IMPACT/veda-ui/blob/a0a5cad1edf0a2319c1d5071352403d933d8b294/app/scripts/components/exploration/components/map/index.tsx#L67-L130), but then depending on what they do (Layer, Control, Compare config) they need to be processed differently, hence this code. This was Erik's implementation of how to abstract away the complexity of the [different layer types](https://github.com/NASA-IMPACT/veda-ui/tree/a0a5cad1edf0a2319c1d5071352403d933d8b294/app/scripts/components/common/map/style-generators) we have.

hanbyul-here:
It will be ideal if we don't have to name controls manually + have a coherent way of handling the children of Map. This : https://github.com/NASA-IMPACT/veda-ui/pull/711/files is what I thought out of it. (Having one container element that holds all the controls) how do you think?

danielfdsilva:
@hanbyul-here I kind of like the current approach since it reduces wrappers and keeps a lean list of children. However the approach that you propose is exactly the same that is in place for the `Compare` so I think it makes more sense.
Will update the PR 👍

danielfdsilva:
@hanbyul-here Ready to review. Fixed the conflicts with the base branch.

hanbyul-here (approved):
Thanks for the fix. 

Not related to this PR, but do you see any problem making this map component (used by mdx map) : https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/mapbox/map.tsx to follow the same pattern?

danielfdsilva:
I'll open an issue for this

PR #714, opened 2023-10-26
[DEMO NOT FOR MERGE] Adding support for NetCDF datasets in CMR

abarciauskas-bgse:


app/scripts/context/layer-data.tsx
```diff
@@ -51,9 +51,13 @@ const fetchLayerById = async (
       }
     };
   } else {
-    const domain = data.summaries
+    let domain = data.summaries
       ? data.summaries.datetime
       : data.extent.temporal.interval[0];
+    const lastDatetime = domain[domain.length - 1]
+    if (lastDatetime == null) {
+      domain[domain.length - 1] = new Date().toISOString()
```
hanbyul-here, line 59: I don't think I understand what is going on with temporal interval. Can you explain why this step is necessary for netCDF data?

hanbyul-here (commented):
One more question:  Did you use Zarr layer for prototype purpose? Or can netcdf and zarr actually be treated in the same way?

app/scripts/context/layer-data.tsx
```diff
@@ -28,14 +28,14 @@ const fetchLayerById = async (
...
+  const stacApiEndpoint = layer.stacApiEndpoint || process.env.API_STAC_ENDPOINT;
   const { data } = await axios.get(
-    `${process.env.API_STAC_ENDPOINT}/collections/${stacCol}`
+    `${stacApiEndpoint}/collections/${stacCol}`
   );
 
   const commonTimeseriesParams = {
-    isPeriodic: data['dashboard:is_periodic'],
-    timeDensity: data['dashboard:time_density']
+    isPeriodic: data['dashboard:is_periodic'] || true,
```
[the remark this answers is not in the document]
abarciauskas-bgse: we won't be able to get is_periodic or time_density from NASA's CMR STAC so we need some fallback, but I think a better way may be to make this parameter configurable in the dataset configuration page, do you think that would be straight forward to do?

app/scripts/context/layer-data.tsx
abarciauskas-bgse: For each of these colletions, the CMR STAC record's second temporal extent is null (see [GPM_3IMERGDF.v07](https://cmr.earthdata.nasa.gov/stac/GES_DISC/collections/GPM_3IMERGDF.v07) and [TRMM_3B42_Daily.v7](https://cmr.earthdata.nasa.gov/cloudstac/GES_DISC/collections/TRMM_3B42_Daily.v7),

abarciauskas-bgse:
@hanbyul-here [good question](https://github.com/NASA-IMPACT/veda-ui/pull/714#pullrequestreview-1708815717). NetCDF and zarr can be treated in the same way, the difference is that NetCDF is a single file and Zarr URL points to the entrypoint of a zarr store which is many small files representing metadata or chunks of data values or coordinate data. I could definitely imagine we may need better constructs for handling cases of Zarr vs NetCDF but I'm not sure it's necessary yet.

abarciauskas-bgse:
Closing in favor of https://github.com/NASA-IMPACT/veda-ui/pull/782

PR #719, opened 2023-10-31
Display error message when the analysis items is over max

danielfdsilva:
Screenshot 2023-10-31 at 10 16 36

### DIFF
--- a/app/scripts/components/exploration/components/datasets/dataset-list-item-status.tsx
+++ b/app/scripts/components/exploration/components/datasets/dataset-list-item-status.tsx
@@ -117,10 +117,8 @@ export function DatasetTrackError(props: {
         {patternContent}
         <TrackMessage isError>
           <p>
-            Too many data points to analyze ({error.details?.assetCount} / max.{' '}
-            {MAX_QUERY_NUM}).
-            <br />
-            Select a shorter time range.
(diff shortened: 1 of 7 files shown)


app/scripts/components/exploration/components/datasets/dataset-list-item-status.tsx
```diff
@@ -79,24 +92,46 @@ export function DatasetTrackLoading(props: { message?: React.ReactNode }) {
...
+  if (error instanceof ExtendedError && error.code === 'TOO_MANY_ASSETS') {
+    return (
+      <>
+        {patternContent}
+        <TrackMessage isError>
+          <p>
+            Too many data points to analyze ({error.details?.assetCount} / max.{' '}
+            {MAX_QUERY_NUM}).
+            <br />
+            Select a shorter time range.
```
j08lue, line 123:
```suggestion
            Please select a shorter time range.
```

j08lue (changes requested):
I have a suggestion for the error message to be a bit more gentle, perhaps. 😄 Would that work or does it get too long?

Either way - we will see, how users react to this and whether they understand how to change the range and repeat / continue the analysis in that case.

danielfdsilva:
@j08lue Message updated!

j08lue (approved):

PR #726, opened 2023-10-31
Fix error caused by updating url too frequently

danielfdsilva:
If the url is updated too often in a short period of time, the app crashes.
It also fixes the error where the analysis was always shown as outdated


j08lue:
Hmm, still seems to happen. Here is what I did:

1. Open /exploration
2. Pick the NO2 Diff datasets
3. Select the entire time period
4. Draw a 4-5 point polygon somewhere around eastern China / Beijing
5. Hit "Analyze"
6. Timeline is now **blank** and Analyze thing shows the "Outdated" warning
7. Hitting refresh on the Analysis controls does not change anything

image

This happens on Brave on Mac OS, also in private window / after hard reload

On Chrome on Mac OS as well as Safari, I also immediately get the "Outdated" warning after submitting an Analysis, but at least the time series still loaded.

image

image

danielfdsilva:
This stared as a PR to fix the fact that the URL was being updated too frequently, but it turned out that there was another problem related to the URL.

I added a [README](https://github.com/NASA-IMPACT/veda-ui/blob/881fb4886b57a0b91c4b5ddfffeb14689ed67a02/app/scripts/components/exploration/atoms/README.md) with an explanation of both problems and what was done to solve them.

https://github.com/NASA-IMPACT/veda-ui/blob/881fb4886b57a0b91c4b5ddfffeb14689ed67a02/app/scripts/components/exploration/atoms/README.md

I feel that this whole url business is a bit hack-y especially Problem 2. We're basically caching values in between state updates which feels a bit wrong. It would be great to have a second opinion on this. @hanbyul-here @sandrahoang686 . If you have ideas for different approached let's discuss.

hanbyul-here:
I was hoping to try the suggestion on my end first and leave a comment, but it looks like I won't have the luxury of it, so I will just comment 😢 Have you considered using Jotai's custom compare function?  https://jotai.org/docs/recipes/atom-with-compare ? We would need to pass deepcompare function in the end, but can take a bit of advantage of Jotai system.

danielfdsilva:
@hanbyul-here Thank you for the tip. Even though I was not able to use `atomWithCompare` directly (and only), because of all the url updates and reconciliation, it gave me a great idea on how to improve the current approach. I was able to remove that cache file that was annoying me, and came up with a much cleaner approach. Updated the code and the readme with the explanation.

Thank you 😊 . Have a look if you're able

hanbyul-here (approved):
atom with url value stability makes more sense with your change, thanks for working on it @danielfdsilva 

This is not a blocker, but I can't help wonder if we need to update the atom value on drag event. I wonder if we can just update on the release? (Apologies if this was already discussed before!)

danielfdsilva:
We do need to store the value on drag so that the handle can move. We could store it in a state hook, and then update the atom, but since the atom is the state, I fear we'd be duplicating efforts.

Issue #729, opened 2023-11-02
Bring back route redirection from /discoveries to /stories


j08lue:
There are apparently still links to old "Discoveries" of the VEDA NASA/EIS instance out there, which are broken since we renamed `/discoveries` to `/stories`.

Like https://www.earthdata.nasa.gov/dashboard/discoveries/wq-models

Did we have a redirect from these legacy links in place at some point? If so, perhaps we removed it while fixing https://github.com/NASA-IMPACT/veda-ui/issues/682?

If it is not much trouble, we should bring this redirect back.

## Acceptance criteria

- [ ] Assessed the effort / consequences of (re-)introducing redirect from discoveries to stories and implemented, if sensible.

hanbyul-here:
There is a change that reintroduces the discovery redirection. At least, based on the assessment of my local profilers, it seems to be working ok. I would rather deprecate this, especially if we are launching EIC page (where this page is linked.) 

https://github.com/NASA-IMPACT/veda-ui/tree/fix/back-redirection

PR #731, opened 2023-11-02
Add discoveries redirection back

hanbyul-here:
This PR reintroduces the redirection of the discoveries -> stories. 

Close https://github.com/NASA-IMPACT/veda-ui/issues/729

Since we could not figure out the root cause of the app crash other than there was a problem with a combination of redirecting and shadow scroll, I generally want to be careful about this change. I tested it a bit and it seems working as expected without overheads.


danielfdsilva (commented):
Wouldn't it be a good idea to also add the thematic redirects back?
Wondering if like with discoveries, some thematic area links are out there.

j08lue:
> Wondering if like with discoveries, some thematic area links are out there.

Thanks for asking. I think discoveries is enough.

PR #733, opened 2023-11-03
Improvements to the analysis

danielfdsilva:
Includes:
- Default selected date to newest of the data domain (on load)
- Removes "A" letter from playhead. Letters "A" and "B" only show up when in compare mode.
- Remove letters from date range selector
- Change range date picker label to "From ... to ..."
- Range datepicker and range playheads only show up when there are features on the map


j08lue:
I can see almost all the changes you made in the preview:

- Default selected date to newest of the data domain (on load)
- Removes "A" letter from playhead. Letters "A" and "B" only show up when in compare mode.
- Remove letters from date range selector
- Change range date picker label to "From ... to ..."

But, on a clean browser, picking NO2 diff and GEOGLAM to begin with, I do get the range sliders and range date pickers even before adding a geometry:

image

danielfdsilva:
@j08lue Was able to replicate the problem and fixed it!

j08lue:
Thanks for the update!

Now I see only one date picker, but initially **no playhead** when I load `/exploration` with NO2 and GEOGLAM.

image

With other combinations (e.g. NO2 PT and NO2 Diff), the playhead does show.

j08lue:
Btw, the "reset analysis"(?) button seems to have the wrong tooltip:

image

And it is also the same as the "update analysis" button:

image

j08lue (changes requested):
UI state changes with reset etc now works nicely. 👌 

However, I still very frequently get the "Outdated! Refresh to analyze an area...":

1. Right after every analysis
2. When I drew an AOI, ran an analysis, removed all AOIs, and then drew a new AOI - in this case I would expect a fresh "Analyze" option, since no analysis is active.

image

Are you able to reproduce this issue (in either/both of the two cases)?

danielfdsilva:
@j08lue Added a couple more commits for some other improvements that I discussed with @ricardoduplos 

- Analysis charts are expanded by default, and the button gets hidden when there are no features (like the datepicker)
- The B playhead appears within a distance from A instead of on top of it

danielfdsilva:
@j08lue Addressed.

1. Fixed with zoom to feature!
2. See comment above
3. There were some situations where the message was not present. I changed the whole logic to ensure there's always user guidance if the analysis is "active" and/or there are features on the map

j08lue (approved):
Thank you so much for these fixes, @danielfdsilva, it is coming together nicely! I noted your pointer to your message / the other PR with the outdated state issue.

I see how you handled the reset / exit case now - it feels more safe that I as a user will not be left with a UI state that is puzzling. Nice solution.

image

I'll review the banner message texts tomorrow, if that is okay - I think it might be too ambiguous what "Outdated" means...

image

If you want to merge this PR already, though, and tackle any new feedback in new PRs, please feel free to go ahead.

hanbyul-here (approved):
I wish the conditional syntax for all the statuses were in a more readable form (it is a lot of conditions!!) but I don't think it should be a blocker

j08lue:
This is a job for pro designers. 🙇  If any are available, please chime in or take over. If not, please consider my proposal below.

## Proposal

1. When user made a selection:
   
   

2. When the selection changed while there is an active analysis:

   

3. When the user reset the selection but there is still a time series on the bottom drawer:

   


## Changes to current implementation

1. No "warning"-level alert to users - changed analysis parameters that are not yet applied is not a problem they need to be alerted to, IMO.
2. Use the same "timeseries" icon on the message head to indicate at that the message relates to the zonal statistics / analysis
3. Spell out button functions not only for initial "Analyze" but also the other options ("Apply Changes" and perhaps also "Exit") for more consistency and clarity
4. Updated message texts for the two cases of (1) changed selection and (2) no active selection

danielfdsilva:
@hanbyul-here I too felt that there were a lot of conditionals, but there are so many states...
I tried to split them into separate components. It helps a bit but not great. If you have other ideas let me know.

For now I'm merging the PR and will tackle the messages in a different one with some input from designers. cc @j08lue

Issue #732, opened 2023-11-02
AOI coordinates outside the -180,180 range cause the /search api to return no features.


danielfdsilva:
FROM: https://github.com/NASA-IMPACT/veda-ui/pull/726#issuecomment-1788111343

On the Exploration page, when the map is zoomed to the point where there are multiple world copies, it is possible to draw an AOI on a map that falls outside the -180,180 range.

In the image below, we can see 3 "full maps", each with its own coordinate range. Since I drew the AOI in the leftmost one, the API will return 0 features.

<img width="1241" alt="Screenshot 2023-11-02 at 19 27 03" src="https://github.com/NASA-IMPACT/veda-ui/assets/1090606/d641d3c0-bd7e-4309-bc09-7ac60f723161">

The following payload will cause 0 features to be returned on a `POST https://test-stac.delta-backend.com/search`

<details><summary>View JSON Payload</summary>
<p>

```
{
    "filter-lang": "cql2-json",
    "limit": 10000,
    "fields": {
        "include": [
            "assets.cog_default.href",
            "properties.start_datetime",
            "properties.datetime"
        ],
        "exclude": [
            "collection",
            "links"
        ]
    },
    "filter": {
        "op": "and",
        "args": [
            {
                "op": "t_intersects",
                "args": [
                    {
                        "property": "datetime"
                    },
                    {
                        "interval": [
                            "2020-01-03T00:00:00.000Z",
                            "2021-01-24T23:59:59.999Z"
                        ]
                    }
                ]
            },
            {
                "op": "s_intersects",
                "args": [
                    {
                        "property": "geometry"
                    },
                    {
                        "type": "MultiPolygon",
                        "coordinates": [
                            [
                                [
                                    [
                                        -200,
                                        50
                                    ],
                                    [
                                        -200,
                                        10
                                    ],
                                    [
                                        -240,
                                        10
                                    ],
                                    [
                                        -240,
                                        50
                                    ],
                                    [
                                        -200,
                                        50
                                    ]
                                ]
                            ]
                        ]
                    }
                ]
            },
            {
                "op": "in",
                "args": [
                    {
                        "property": "collection"
                    },
                    [
                        "no2-monthly"
                    ]
                ]
            }
        ]
    }
}
```

</p>
</details>

This becomes all the more problematic when the AOI crosses into another map, which is very possible if the users' AIO is over the pacific ocean.

[...]

PR #734, opened 2023-11-06
Ensure polygons are inside -180/180 range before making stac request

danielfdsilva:
See #732

I created a script to cut the polygons that cross the antimeridian and them shift them inside the correct range.
Includes some unit tests for the function.

### DIFF
--- a/app/scripts/utils/antimeridian.test.ts
+++ b/app/scripts/utils/antimeridian.test.ts
@@ -2,7 +2,7 @@ import { Feature, MultiPolygon, Polygon } from 'geojson';
 import { fixAntimeridian } from './antimeridian';
 
 describe('Antimeridian', () => {
-  it('Feature in 1st earth east', () => {
+  it('Should move a feature in the first map east (180/540) to the -180/180 range', () => {
     const feature: Feature<Polygon> = {
       type: 'Feature',
       properties: {},
@@ -43,7 +43,7 @@ describe('Antimeridian', () => {
     expect(result).toEqual(expected);
   });
 
-  it('Feature crossing 540 antimeridian earth east', () => {
+  it('Should split a feature crossing the 540 antimeridian into 2 that fall within the -180/180 range', () => {
     const feature: Feature<Polygon> = {
       type: 'Feature',
       properties: {},
@@ -93,7 +93,7 @@ describe('Antimeridian', () => {
     expect(result).toEqual(expected);
   });
 
-  it('Feature spanning several maps', () => {
+  it('Should calculate the resulting feature in the -180/180 range when the input spans several maps', () => {
     const feature: Feature<Polygon> = {
       type: 'Feature',
       properties: {},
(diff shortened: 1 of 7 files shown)


j08lue:
When I draw on the left-most or right-most out of four cycles of the Earth across the antimeridian, part of the geometry seems to be there but not visible:

Drawn on left-most cycle:

image

Drawn on right-most cycle:

image

j08lue:
We earlier discussed whether we could just show only one cycle of the Earth. My concern then was, that this would make the map look weird. However, if this causes too much headache, could we perhaps show the whole map (with repetitions), but guide the user only draw within those cycles where we can do the normalization safely?

danielfdsilva:
3 maps on a very wide and narrow window.
image

Full screen on my high-res monitor, with a smaller timeline allows to view the whole map
image

j08lue (approved):
Awesome, great work handling all these over-the-edge cases... The guardrails keep me from running into the original issue now.

app/scripts/utils/antimeridian.ts
```diff
@@ -0,0 +1,213 @@
...
+    // So we need to find out by how many times we need to shift 360. This will
+    // be defined by the smallest lng value. Any coordinates that are still
+    // outside after the shift will be shifted again in a later run.
+    const largestLng = Math.max.apply(
+      null,
+      ring.map(([lng]) => lng)
+    );
+
+    // Or 1 in case the largestLng is exactly -180.
+    const shift = Math.ceil(Math.abs((largestLng + 180) / 360)) || 1;
```
hanbyul-here, line 147: Is it okay to round up like this? 1 degree in Mercator doesn't seem to be that small number?

app/scripts/utils/antimeridian.test.ts
```diff
@@ -0,0 +1,243 @@
...
+    };
+
+    const expected: Feature<MultiPolygon> = {
+      type: 'Feature',
+      properties: {},
+      geometry: {
+        coordinates: [
+          [
+            [
+              [-170, -20],
```
hanbyul-here, line 30: Can you point me where the order of the coordinates changes? I can't wrap my head around it.

hanbyul-here (approved):
Yay test. (Test description can be more helpful! ) I left a concern about an edge case and a question about the order of the coordinates. 🤔 I wonder if there are better names than `east` and `west` for coordinates to go over >180, <-180 . Nothing seems to be a blocker

app/scripts/utils/antimeridian.ts
danielfdsilva: This does not refer to 1º but to how many times to ship 360º. The result is that the coordinates move to the "repeated map" they are to the correct one (between -180/180)

app/scripts/utils/antimeridian.test.ts
danielfdsilva: That was also strange to me, but for some reason the turf.js function (intersect and difference), reorder the coordinates when cutting the polygons.

danielfdsilva:
@hanbyul-here You're right. The messages were absolutely atrocious 😅. I made them more descriptive and will merge. Feel free to suggest further improvements.
I could not thing of better names than west/east, but happy to hear a suggestion.

PR #735, opened 2023-11-06
Give an option to be an 'external link' for story

### LINKED ISSUE #728: [EIC] Give an option to story to be external link
The EIC wants to use the "Stories" feature to present different "Themes" of Earth information. Some of these themes would be actual content in VEDA UI, but some would link to external content instead.

This would be equivalent to their [existing website](https://gis.earthdata.nasa.gov/portal/apps/sites/#/earth-information-center/pages/connected-earth):

image

Note the subtle difference between arrows pointing right (local content) and right-upwards (external content). If we just modify our cards, we would probably need some kind of indication for this difference, too.

## Acceptance criteria

- [ ] Considered designs for story cards that point to external content
- [ ] Implemented in the story config / UI

hanbyul-here:
Close #728 

I will call it a story in the development context. (even though it will be a theme for EIC).

The external link does not open in a new tab 😏

### DIFF
--- a/app/scripts/components/common/card.tsx
+++ b/app/scripts/components/common/card.tsx
@@ -308,7 +308,6 @@ interface CardComponentProps {
   title: ReactNode;
   linkLabel: string;
   linkTo: string;
-  isExternalLink?: boolean;
   className?: string;
   cardType?: CardType;
   description?: ReactNode;
@@ -330,7 +329,6 @@ function CardComponent(props: CardComponentProps) {
     description,
     linkLabel,
     linkTo,
-    isExternalLink,
     date,
     overline,
     imgSrc,
@@ -341,6 +339,7 @@ function CardComponent(props: CardComponentProps) {
     onCardClickCapture
   } = props;
 
+  const isExternalLink = linkTo.match(/^https?:\/\//);
   const linkProps = isExternalLink
     ? { href: linkTo }
     : { as: Link, to: linkTo };
--- a/app/scripts/components/common/card.tsx
+++ b/app/scripts/components/common/card.tsx
@@ -2,6 +2,7 @@ import React, { MouseEventHandler, ReactNode } from 'react';
 import styled, { css } from 'styled-components';
 import { Link } from 'react-router-dom';
 import { format } from 'date-fns';
+import { CollecticonExpandTopRight } from '@devseed-ui/collecticons';
 import { VerticalDivider } from '@devseed-ui/toolbar';
 
 import {
@@ -198,11 +199,11 @@ export const CardMeta = styled.div`
     }
   }
 
-  > ${/* sc-selector */VerticalDivider}:last-child {
+  > ${/* sc-selector */ VerticalDivider}:last-child {
     display: none;
   }
 
-  > ${/* sc-selector */VerticalDivider}:first-child {
+  > ${/* sc-selector */ VerticalDivider}:first-child {
     display: none;
   }
 `;
@@ -284,11 +285,30 @@ const CardFigure = styled(Figure)`
   }
 `;
 
+const ExternalLinkMark = styled.div`
+  position: absolute;
+  top: 0;
+  right: 0;
+  padding: ${variableGlsp(0.25)};
+  background-color: ${themeVal('color.primary')};
+  color: ${themeVal('color.surface')};
+  z-index: 100000;
+`;
+
+export function ExternalLinkFlag() {
+  return (
+    <ExternalLinkMark>
+      <div>External Link</div>
+      <CollecticonExpandTopRight />
+    </ExternalLinkMark>
+  );
+}
+
 interface CardComponentProps {
   title: ReactNode;
   linkLabel: string;
   linkTo: string;
-  href?: string | undefined;
+  isExternalLink?: boolean;
   className?: string;
   cardType?: CardType;
   description?: ReactNode;
@@ -310,7 +330,7 @@ function CardComponent(props: CardComponentProps) {
     description,
     linkLabel,
     linkTo,
-    href,
+    isExternalLink,
     date,
     overline,
     imgSrc,
@@ -321,10 +341,10 @@ function CardComponent(props: CardComponentProps) {
     onCardClickCapture
   } = props;
 
-  const linkProps = href? { href } : {
-    as: Link,
-    to: linkTo
-  };
+  const linkProps = isExternalLink
+    ? { href: linkTo }
+    : { as: Link, to: linkTo };
+
   return (
     <ElementInteractive
       as={CardSelf}
@@ -338,7 +358,8 @@ function CardComponent(props: CardComponentProps) {
         <CardHeadline>
           <CardTitle>{title}</CardTitle>
           <CardOverline as='div'>
-            {parentName && parentTo && (
+            {isExternalLink && <ExternalLinkFlag />}
+            {!isExternalLink && parentName && parentTo && (
               <CardLabel as={Link} to={parentTo}>
                 {parentName}
               </CardLabel>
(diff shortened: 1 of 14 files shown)


j08lue (commented):
Works nicely! 


Leaving it to @danielfdsilva to give the green light. 

We'll still need to figure out a design that indicates that these are (external) links, plus at least rudimentary docs, I guess? 😬

parcel-resolver-veda/index.d.ts
```diff
@@ -255,7 +264,7 @@ declare module 'veda' {
   export const getOverride: (key: PageOverrides) => VedaDatum<any> | undefined;
 
   export const getString: (variable: string) => {
-    one: string;
```
danielfdsilva, line 258: I think the strings should always have a default. Is there a use case where this would be different?
If this is needed then `other` should also be `string | undefined`

app/scripts/components/common/card.tsx
```diff
@@ -319,16 +321,17 @@ function CardComponent(props: CardComponentProps) {
     onCardClickCapture
   } = props;
 
+  const linkProps = href? { href } : {
+    as: Link,
+    to: linkTo
+  };
```
danielfdsilva, line 327: I'd like to suggest a different implementation for this.

Instead of adding a `href` property to cards that can possible lead to some confusion, I'd propose to use the `linkTo` property even for external links.
Then, here, you'd check if the links starts with `http`. If so it is external.

```js
const linkProps = linkTo.match(/^https?:\/\//)
  ? { href: linkTo }
  : {
      as: Link,
      to: linkTo
    };
```

danielfdsilva (changes requested):

parcel-resolver-veda/index.d.ts
hanbyul-here: This is not a change of how the data is used, (also out of scope for this pr) more of an attempt to reflect what the current data structure is like to types. I read that anything that required singular/plurals needs to have `one`  + `others`, but anything that doesn't require singular/plurals should have others == every string should have `others` but only some require `one`. (ex. `storiesBanner`)

app/scripts/components/data-catalog/index.tsx
```diff
@@ -156,15 +156,13 @@ function DataCatalog() {
       <LayoutProps
         title='Data Catalog'
         description={
-          getString('dataCatalogBanner')?.other ||
```
hanbyul-here, line 159: Wait, this line tripped me into thinking that all the strings are object with `one` and `other`. But it is actually not based on ghg configuration? https://github.com/NASA-IMPACT/veda-config-ghg/blob/develop/veda.config.js#L41

danielfdsilva: That is correct. All strings are an object, but to simplify configuration, the user can provide just a string. If that happens, the veda module will the use the same for `one` and `other`.
See here:

https://github.com/NASA-IMPACT/veda-ui/blob/6105ce04cf964e5df2eb3c9a70d111fc84d323dd/parcel-resolver-veda/defaults.js#L10-L20

danielfdsilva: Documented at: https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/CONFIGURATION.md#strings

parcel-resolver-veda/index.d.ts
danielfdsilva: Both properties will be the there as explained in the other comment.

app/scripts/components/data-catalog/index.tsx
hanbyul-here: ok then it might be a bigger change if we want the type to correctly reflect the current data structure. So getString type is 
```
const getString: (variable: string) => {
    one: string;
    other: string;
  } | string;
```

And then anywhere that uses getString will need to be changed into

```
const storiesStringToUse = (typeof  getString('stories') ===string)?  getString('stories'): getString('stories').one
```

(Separating numeric string from plain string will be ideal, but I don't really want this PR to handle that issue. )

app/scripts/components/common/card.tsx
danielfdsilva: Apologies if my suggestion was not clear. That check should go in the `Card` component. This component, as you said, can be used for internal links as well.

The result is that you can then use the `linkTo` property for both internal and external.

So, instead of having:
```
<Card
  ...
   href={d.external?.url}
   linkTo={getStoryPath(d)}
/>
```

you can have
```
<Card
  ...
   linkTo={d.external?.url ?? getStoryPath(d)}
/>
```
And you leave it to Card to figure out if it is internal or not.

hanbyul-here: I meant any regex check does not seem necessary because the original design only allows the link as an external link (which always should start with https). There are some moving parts right now but I do agree that this should go as a prop for Card level.

app/scripts/components/data-catalog/index.tsx
hanbyul-here: How does `storiesBanner` string get translated in Veda module? If it is a string with a value 'aaa', it becomes `{one: 'aaa', other: 'aaa'}`? 

Also, it might be beneficial to mention I started this whole type thing because my automatic typechecker was giving an error for: `getString('storeisBanner)?.other`
Because `?` should not be necessary.  But when I get rid of `?` The app breaks because `getString('stroiesBanner')` was not guaranteed. So I wanted our type to reflect the data usage correctly, so the typechecker works automatically. Return type definition of `getString` got me think that the data should be defined as `{one: , other: }` form. (But if my assumption in the first paragraph is right, I guess it could be just a string value.).

Based on my confusion, wouldn't it be a better use of type if we actually separate numeric strings from non-numeric strings? I just pushed the change that treats numeric strings differently from the non-numeric ones.

hanbyul-here:
I am still working on style, just pushed the changes to share the progress.

hanbyul-here:
Okay now the change is ready, including the design 😌

app/scripts/components/common/card.tsx
```diff
@@ -284,6 +285,38 @@ const CardFigure = styled(Figure)`
...
+  align-items: center;
+  position: absolute;
+  top: ${variableGlsp(0.25)};
+  right: ${variableGlsp(0.25)};
+  padding: ${variableGlsp(0.125)} ${variableGlsp(0.25)};
+  background-color: ${themeVal('color.primary')};
+  color: ${themeVal('color.surface')};
+  text-transform: none;
+  border-radius: calc(
+    ${multiply(themeVal('shape.rounded'), 2)} - ${variableGlsp(0.125)}
```
hanbyul-here, line 299: So @faustoperez shared this cool trick on how to derive the value for nested border-radius, which should be (outside border-radius - gap ): https://www.30secondsofcode.org/css/s/nested-border-radius/ But if we take it as it is, the value becomes 0, so I halved the value for the gap here. :[

j08lue (approved):
LGTM in terms of functionality / look and feel.

image

😍

danielfdsilva (changes requested):
This looks great! Sorry about all the hoops we had to go through.
The only remaining thing is to document the `asLink` property for stories and we're done!

hanbyul-here:
One thing about this pr, is that we are not blocking the entry of this story. So, if somebody types out the story id in single route, that person will land on the page. It is a very unlikely behavior to do something about it, but something to be aware. 

I created a ticket for documentation: https://github.com/NASA-IMPACT/veda-ui/issues/742

danielfdsilva:
@hanbyul-here That's a great catch, and very easily fixable.

on `/app/scripts/components/stories/single/index.tsx`
https://github.com/NASA-IMPACT/veda-ui/blob/bdcd56e6aac75c97458cbc95fc8d02b88f7168a7/app/scripts/components/stories/single/index.tsx#L18

replace with
```
if (!story || story.asLink) throw resourceNotFound(); 
```

danielfdsilva (approved):

PR #736, opened 2023-11-06
Add Content Override option to stories hub

hanbyul-here:
Add Content Override for Stories Hub page for EIC.
I also got rid of page local nav.

Close https://github.com/NASA-IMPACT/veda-config-eic/issues/6


app/scripts/components/stories/hub/index.tsx
```diff
@@ -161,150 +163,153 @@ function StoriesHub() {
...
-        description={getString('storiesBanner')?.other || 'Explore the guided narratives below to discover how NASA satellites and other Earth observing resources reveal a changing planet.'}
+        description={getString('storiesBanner').other}
       />
-
+      <ContentOverride with='storiesHubHero'>
+        <PageHero
+          title={getString('stories').other}
+          description={getString('storiesBanner').other}
+        />
+      </ContentOverride>
```
danielfdsilva, line 173: Seems that this should be a `ComponentOverride` which doesn't use frontmatter variables or markdown. Seems the [same that is being done on the homepage](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/home/index.tsx#L137-L164), but not sure about the endgoal

danielfdsilva (changes requested):

app/scripts/components/stories/hub/index.tsx
hanbyul-here: Yeah I think you are right, changed it to comonentOverride

danielfdsilva: My suggestion was to just change this one to ComponentOverride. The other one should stay as ContentOverride as it replaces the whole page content.

danielfdsilva:
One additional thing to add is documentation for the new overrides: https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/PAGE_OVERRIDES.md

hanbyul-here:
@danielfdsilva I will make a ticket for documentation for later so it is not a blocker for the pr

app/scripts/components/stories/hub/index.tsx
```diff
@@ -161,150 +163,153 @@ function StoriesHub() {
...
-      <PageHero
-        title={getString('stories').other}
-        description={getString('storiesBanner')?.other || 'Explore the guided narratives below to discover how NASA satellites and other Earth observing resources reveal a changing planet.'}
+        description={getString('storiesBanner').other}
       />
-
+      <ComponentOverride with='storiesHubHero'>
+        <PageHero
+          title={getString('stories').other}
+          description={getString('storiesBanner').other}
```
sandrahoang686, line 171: 💭 : Future case - not a blocker. This makes me think we should probably validate the config if this isn't happening already somewhere? To make sure required keys in the config exist otherwise optional chaining can help here

parcel-resolver-veda/defaults.js
```diff
@@ -4,7 +4,8 @@ const defaultStrings = {
   stories: {
     one: 'Story',
     other: 'Stories'
-  }
+  },
+  storiesBanner: 'Explore the guided narratives below to discover how NASA satellites and other Earth observing resources reveal a changing planet.'
```
sandrahoang686, line 8: ❓ : Below you do `getString('storiesBanner').other` but the default has no `other` key? Would this be alright?

Also is here a config template? Possibly as documentation and as well as usage for a possible validator that I talked about below..

danielfdsilva: @sandrahoang686 For any string that gets added, there should be a default in the `parcel-resolver-veda/defaults.js`. This will ensure that the optional chaining is not need.

Then, to simplify configuration, the user can define either an object with `one` and `other` or a single string. The veda module will take care of normalizing the format, ensuring that there is always `one` and `other` for the implementer to use, with both values being the same if the user just provided a string. In this way the function return is always the same.
There was a bug in the code that performed this operation, but you can see it fixed here: https://github.com/NASA-IMPACT/veda-ui/blob/da4527b6d59110576b13ae12fdc9f15b97ad39f5/parcel-resolver-veda/defaults.js

danielfdsilva (approved):
Looks good. Fixed the conflicts with `eic`

app/scripts/components/stories/hub/index.tsx
hanbyul-here: @danielfdsilva 's comment probably explained the current state of the return value of `getString.` I just want to mention that I also thought our type was wrong and tried to fix it in this pr: (reverted it back.) https://github.com/NASA-IMPACT/veda-ui/pull/735

So my 💭 for the future is `getString` is trying to do too many things, and makes the use of it confusing. It might be worth separating the data that needs `one`, `other` from the one that doesn't need it. But I also think this is out of the scope of this PR.

PR #739, opened 2023-11-08
Cleaning page-header global link items to map over defaults

sandrahoang686:
This is just a PR to map over default menu link items instead of hard coding it. In the future we should look at how we handle custom nav items and think about whether or not the consumer should just provide this entirely


hanbyul-here:
I am sorry that we never got to this PR! Since this navigation PR addresses the concerns, https://github.com/NASA-IMPACT/veda-ui/pull/1072 I think we can close this PR. What do you think? @sandrahoang686

PR #743, opened 2023-11-09
Lazy iframe

hanbyul-here:
It seems EIC will require plenty of Embedded content. This PR wraps `Embed` component (which is essentially iframe) into `<Lazyload >` element, so embedded contents can load only when it is in the view. 

Lazy embed example: https://deploy-preview-743--veda-ui.netlify.app/stories/life-of-water


danielfdsilva (approved):
🥳

PR #744, opened 2023-11-09
Bring your own stac, tile endpoint

hanbyul-here:
Rather than dealing with 'multi catalog' issue, this PR narrows down the scope and makes the layer to be responsible to bring stac and tile endpoints when needed. (`stacApiEndpoint` and `tileApiEndpoint` to data mdx.)

~~We are using `stacCol` as an identifier in layer fetching logic, especially in `context/layer-data`. This will make a problem if there exist the same collection ids in different stac instances. So, we need to figure out a unique identifier for the layer fetching. But with the new A&E page coming and everything, I feel more biased towards documenting this as a caveat (no duplicate stac collection IDs are allowed) and moving forward for now. What do you think @danielfdsilva ? @sandrahoang686 ?~~  : Welp 💦 It was less amount of effort than I thought so I just went ahead and added 'stacEndpoint' as an additional identifier.
 
 
ex.  Previous attempt to implement multi-catalog (from @slesaad ) : https://github.com/NASA-IMPACT/veda-config/pull/301/files


app/scripts/components/analysis/define/use-stac-collection-search.ts
```diff
@@ -37,10 +33,31 @@ export function useStacCollectionSearch({
...
     queryFn: async ({ signal }) => {
-      const collectionResponse = await axios.get(collectionUrl, {
-        signal
-      });
-      return collectionResponse.data.collections;
+      const collectionUrlsFromDataSets = allAvailableDatasetsLayers
+        .filter((dataset) => dataset.stacApiEndpoint)
+        .map((dataset) => `${dataset.stacApiEndpoint}/collections`)
+        .filter((value, index, array) => array.indexOf(value) === index);
+
```
sandrahoang686, line 40: We can probably use the `reduce` method instead to simplify the first two iterations, not sure what the point of the last one is

```
const collectionUrlsFromDataSets = allAvailableDatasetsLayers.reduce((filtered, dataset) => {
   if (dataset.stacApiEndpoint) {
      filtered.push(`${dataset.stacApiEndpoint}/collections`)
   }
return filtered;
}, [])
```

app/scripts/components/analysis/define/use-stac-collection-search.ts
```diff
@@ -37,10 +33,31 @@ export function useStacCollectionSearch({
...
+      const collectionUrls = [
+        ...collectionUrlsFromDataSets,
+        `${process.env.API_STAC_ENDPOINT}/collections`
+      ];
+
+      const colloectionRequests = collectionUrls.map((url: string) =>
```
sandrahoang686, line 46: ⛏️ small spelling mistake

app/scripts/components/analysis/define/use-stac-collection-search.ts
```diff
@@ -37,10 +33,31 @@ export function useStacCollectionSearch({
...
+          }));
+        })
+      );
+      return axios.all(colloectionRequests).then(
+        axios.spread((...responses) => {
+          // Merge all responses into one array
+          const mergedData = [].concat(...responses);
+          return mergedData;
+        })
+      );
```
sandrahoang686, line 60: ⛏️  : can we possibly simplify here with 
`axios.spread((...responses) => [].concat(...responses))`
?

app/scripts/components/analysis/define/use-stac-collection-search.ts
hanbyul-here: I like `filter` + `map` when there is no shared condition between them. (I think it makes the code more straightforward to read. lmk if you think differently though!) But hm it seems the last condition is not clear. It is there to pick only unique values. We need this condition because multiple layers can point the same stac api endpoint. I tried a different approach with `Set` which I am not really sure is easy to read, but probably at least make it exclusive about what it is trying to do? lmk if you have a better idea!

danielfdsilva: I also have a preference for `reduce`. IMO it is not harder to read and avoids array cycles. Fair to say that in this case it has no impact, but the principle stands.
However I'd avoid mutating methods:
```ts
const collectionUrlsFromDataSets = allAvailableDatasetsLayers.reduce((filtered, { stacApiEndpoint }) => {
  returnstacApiEndpoint ? [...filtered, `${stacApiEndpoint}/collections`] : filtered;
}, [])
```

danielfdsilva (approved):
Left a response to a comment, but otherwise this looks good!

app/scripts/components/analysis/define/use-stac-collection-search.ts
hanbyul-here: oki, changed this prt to use `reduce`. Thanks for your inputs!

PR #748, opened 2023-11-14
Add user page items to main nav

hanbyul-here:
I made this change because of the strong request from EIC. 

I am very hesitant about suggesting this change because there is no way to prevent the navigation bar from breaking with this change when a user passes too many 'navMainItem' (or even items with long names). We have to clearly communicate that veda-ui is not in charge of the breaks that can be introduced from`navMainItem`. 

Look at how it can look like : https://github.com/NASA-IMPACT/veda-config-eic/pull/24

### DIFF
--- a/app/scripts/components/common/page-header.tsx
+++ b/app/scripts/components/common/page-header.tsx
@@ -354,12 +354,12 @@ function PageHeader() {
 
   const userPagesMainNavItem = userPages.map((id) => {
     const page = getOverride(id as any);
-    if (!page?.data.navMainItem) return false;
+    if (!(page?.data.mainNavItem && page.data.mainNavItem.navTitle)) return false;
 
     return (
       <li key={id}>
         <GlobalMenuLink to={id} onClick={closeNavOnClick}>
-          {page.data.navTitle}
+          {page.data.mainNavItem.navTitle }
         </GlobalMenuLink>
       </li>
     );
@@ -511,7 +511,7 @@ function UserPagesDotMenu(props: {
     return menuItems;
   }, []);
 
-  if (!dotMenuItems.length) return <>{false}</>;
+  if (!dotMenuItems.length) return false;
 
   if (isMediumDown) {
     return (
--- a/app/scripts/components/common/page-header.tsx
+++ b/app/scripts/components/common/page-header.tsx
@@ -511,7 +511,7 @@ function UserPagesDotMenu(props: {
     return menuItems;
   }, []);
 
-  if (!dotMenuItems.length) return false;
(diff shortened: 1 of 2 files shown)


j08lue:
> We have to clearly communicate that veda-ui is not in charge of the breaks that can be introduced from `navMainItem`.

I'll focus on communicating this risk to configurators / site owners. 

We should also add a big warning sign where we document this feature.

For the future, we should add a backlog ticket to follow up on this implementation and consider designs / responsive features for narrower viewports.

app/scripts/components/common/page-header.tsx
```diff
@@ -342,16 +342,29 @@ function PageHeader() {
...
     if (!isMediumDown) setGlobalNavRevealed(false);
   }, [isMediumDown]);
 
   const closeNavOnClick = useCallback(() => {
     setGlobalNavRevealed(false);
   }, []);
 
+  const userPagesMainNavItem = userPages.map((id) => {
+    const page = getOverride(id as any);
+    if (!page?.data.navMainItem) return false;
```
sandrahoang686, line 357: Interesting! didn't know `return false` is different from `return null`

app/scripts/components/common/page-header.tsx
```diff
@@ -478,25 +489,41 @@ function PageHeader() {
...
+
+function UserPagesDotMenu(props: {
   isMediumDown: boolean;
   onItemClick: () => void;
 }) {
   const { isMediumDown, onItemClick } = props;
 
-  if (!userPages.length) return <>{false}</>;
+  const dotMenuItems = userPages.reduce((menuItems: DotMenuItem[], id: any) => {
+    const page = getOverride(id as any);
```
sandrahoang686, line 504: Does it make sense to cast this id as `PageOverrides` ? Hmm... or do we have to do this because it no longer makes sense too manage the extra pages needed for eic or whatever instance in the future as part of `PageOverrides` because the needs are getting too specific?

Whatever the case, at some point in the future.. if we can tackle debt, it would be great to get rid of this component/page overrides structure and this type

app/scripts/components/common/page-header.tsx
```diff
@@ -478,25 +489,41 @@ function PageHeader() {
...
+    if (page?.data.menu)
+      // eslint-disable-next-line fp/no-mutating-methods
+      return menuItems.concat({
+        id,
+        menu: page.data.menu
+      });
+    return menuItems;
+  }, []);
+
+  if (!dotMenuItems.length) return <>{false}</>;
```
sandrahoang686, line 514: just `return false` like above? Dont think we need the empty tags

sandrahoang686 (approved):

PR #749, opened 2023-11-14
Pass other props for iframe

hanbyul-here:
I realized that there are multiple props that iframe can accept to make the communication safe/ more compatible with standard (such as https://github.com/NASA-IMPACT/veda-config-eic/pull/23/files#diff-9a900b0095d3eecf847dc69b3300d909fb529d47ce3ffe375330081d0bcdee76R145) 

This PR makes all those props be passed to iframe element.

### DIFF
--- a/app/scripts/components/common/blocks/embed.tsx
+++ b/app/scripts/components/common/blocks/embed.tsx
@@ -22,7 +22,7 @@ interface EmbedProps {
   height: number;
 }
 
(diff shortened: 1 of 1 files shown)


sandrahoang686 (approved):

hanbyul-here:
🤚 I realized that this change makes it impossible to override the styles of Embed components (becuase props are directly being attached to iframe, not the wrapper). I will file this as a ff issue.

PR #750, opened 2023-11-14
Remove trailing Overview from page title

### LINKED ISSUE #738: Remove "Overview" from title on Dataset pages
image

minus "Overview", since it weirdly becomes part of the dataset title

This has been requested for GHG Center but is relevant feedback for all instances.

## Acceptance criteria

- [ ] Remove "Overview" from title

j08lue:
This has been requested for a while now: dataset overview page titles are currently the dataset title plus the word "Overview", which cannot be visually distinguished from the dataset name and may be falsely taken to be part of it.

The suggestion is therefore to just remove the trailing "Overview". Since it has long been requested, I recommend that we accept this change and revisit any needs for highlighting that the page presents a "dataset overview" later, if it ever comes up.

Closes #738

### DIFF
--- a/app/scripts/components/datasets/s-explore/index.tsx
+++ b/app/scripts/components/datasets/s-explore/index.tsx
@@ -585,7 +585,7 @@ function DatasetsExplore() {
         hideFooter
       />
       <PageMainContent>
-        <PageHero title={dataset.data.name} />
+        <PageHero title={dataset.data.name} isHidden />
         <Explorer>
           <Panel revealed={panelRevealed} onClick={onPanelClick}>
             <PanelInner ref={panelBodyRef}>
--- a/app/scripts/components/datasets/s-explore/index.tsx
+++ b/app/scripts/components/datasets/s-explore/index.tsx
@@ -571,7 +571,7 @@ function DatasetsExplore() {
   return (
     <>
       <LayoutProps
-        title={`${dataset.data.name} Exploration`}
+        title={`${dataset.data.name} - Exploration`}
         description={dataset.data.description}
         thumbnail={dataset.data.media?.src}
         localNavProps={{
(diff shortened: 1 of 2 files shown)


app/scripts/components/datasets/s-overview/index.tsx
```diff
@@ -35,7 +35,7 @@ function DatasetsOverview() {
   return (
     <>
       <LayoutProps
-        title={`${dataset.data.name} Overview`}
+        title={dataset.data.name}
```
hanbyul-here, line 38: oh do we not even want it on the document title?

hanbyul-here (approved):
I left one question for clarification, but thanks for picking this up!

app/scripts/components/datasets/s-overview/index.tsx
j08lue: Good point - I actually don't know where the document title comes up. We can keep the suffix there, I guess, I'd just separate it from the dataset title, if that's okay.

app/scripts/components/datasets/s-explore/index.tsx
```diff
@@ -585,7 +585,7 @@ function DatasetsExplore() {
         hideFooter
       />
       <PageMainContent>
-        <PageHero title={`${dataset.data.name} Exploration`} isHidden />
+        <PageHero title={dataset.data.name} isHidden />
```
hanbyul-here, line 588:
```suggestion
        <PageHero title={dataset.data.name} />
```

hanbyul-here (changes requested):
Did you add `isHidden` intentionally? If not, can you get rid of it?

j08lue:
> Did you add `isHidden` intentionally? If not, can you get rid of it?

I did not add it - it was there before. I removed it now, though.

hanbyul-here (approved):
First of all, I am not sure why this is happening, but the title is not updating. (Def not introduced by this pr. - I feel like it is react-helmet's quirky behavior but this would be out of scope of this pr.) 

Second, the document title is already formatted in `${pagetitle} - ${apptitle}`. NO2 exploration page title will be `NO2 - Exploration - VEDA Dashboard`. If this is the format we want, go ahead.

Issue #747, opened 2023-11-14
[E & A] Add info icon to dataset card


faustoperez:
Quick action after E&A user feedback, ideally to be completed before Earth.gov deadline @j08lue @ricardoduplos @danielfdsilva

We are missing the info button on the dataset card that was present in the previous iteration of the Exploration page. Users tell us they are missing context for the datasets on the new designs. Transcript of the user sessions on this Miro board: https://miro.com/app/board/uXjVNPNOr2w=/?share_link_id=224183859074

<img width="922" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/3e4c6133-fc47-421b-ae0f-fe4ea8ceafcb">

The proposed fix is to add a third icon to the dataset card to link the dataset to the Overview page to meet the Earth.gov release (for now) until we have time to review the relationship between the two pages more thoroughly.

<img width="353" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/36dcdb6d-d8df-4a65-a52b-4652853463ca">

Link to Figma file: https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=528%3A250&mode=design&t=zOPJ2xvx5njzr9qC-1

PR #751, opened 2023-11-15
Improvements to analysis page from the feedback

danielfdsilva:
Contributes to NASA-IMPACT/veda-config-ghg#221 to get us close to a delivery state.

Includes the work done in #744 plus changes so that custom api endpoints are supported in the exploration page.
Besides the former:
- Connects the data-catalog to exploration by filtering the dataset selector modal by the veda dataset name
- Removes connections to previous analysis and dataset explore pages
- Add info icon to dataset https://github.com/NASA-IMPACT/veda-ui/issues/747
- Ensure that the analysis range is within the current domain (bug identified in the feedback session)
- Add support for vector datasets - I noticed that this was no implemented in the A&E so I added it. The zarr layer support is still not implemented, but not sure we need it immediately.


danielfdsilva:
One thing missing: Port modal to view tile endpoints @danielfdsilva

j08lue:
> * The zarr layer support is still not implemented, but not sure we need it immediately.

If we release this first on VEDA, I guess we will need to have this support, since we have a dataset there that relies on TiTiler-Xarray: https://github.com/NASA-IMPACT/veda-config/pull/323

j08lue (approved):
The three most visible improvements

> Connects the data-catalog to exploration by filtering the dataset selector modal by the veda dataset name
> Removes connections to previous analysis and dataset explore pages
> Add info icon to dataset https://github.com/NASA-IMPACT/veda-ui/issues/747

Are very nicely done and seem to work well.

Not sure how to reproduce the bug with analysis range out of bounds - I think this was caused by running an analysis on a different dataset/layer, then switching while it was still active...? I trust that the fix is helpful either way and we will see whether this issue occurs again in any subsequent testing.

We should revisit the list of showstoppers for a release in https://github.com/NASA-IMPACT/veda-config-ghg/issues/221#issue-1992910002 and add any new ones that have up, but this closes some of the main ones.

danielfdsilva:
@j08lue Alright, I'll port the zarr viz as well. Should be pretty easy since it is only a layer component.

For completion, to replicate the range error on the non fixed version:
- Add NPP and CO2 datasets
- Draw AOI and set range to somewhere in 2020 (only co2 has data)
- Remove CO2 dataset

The bug is that the range will remain in 2020, even though the remaining dataset (NPP) only has data up to 2018. This no longer happens in this PR.

hanbyul-here (approved):
I got some questions/ noticed some items that are not necessarily related to this pr (nothing blocker)

1. What is the expected behavior for vector, zarr. layer for 'analysis' ? I don't actually know if analysis endpoint works the same for these different layers. Probably @j08lue can answer this question?

2. I noticed that people are just repeating the same layer (with the same ids) to populate the layers in different datasets ex. mtbs-burn-severity .  It seems we rely layer id to component key here and there, ends up warning about the duplicate keys.  

3. Since we introduced 'stacApiEndpoint' for layers, we probably can get rid of XARRRAY_ENDPOINT: https://github.com/NASA-IMPACT/veda-config/blob/develop/.env#L15 from env, and move this info to zarr layer mdx file.

j08lue:
Great review comments, @hanbyul-here.

> Probably @j08lue can answer this question?

Regarding Analysis on TiTiler-Xarray datasets: no this functionality (`/statistics`) does currently not exist on TiTiler-Xarray. Can we as a first-off catch the failure when the UI tries to make that call and show an error message to the users?

We need to look into implementing that functionality, but it is also ok that it does not exist - the TiTiler-Xarray dataset is advertised as experimental anyways.

danielfdsilva:
1) I think now just shows empty, but I'll check

2) If they have the same id they'll be considered the same layer, which may not be wrong?! But not sure why you need the same layer on different datasets. Might be good to understand the use case.

danielfdsilva:
Will merge this and make the remaining tweaks on a separate PR

PR #753, opened 2023-11-16
Fix iframe styled component props

hanbyul-here:
Follow up of https://github.com/NASA-IMPACT/veda-ui/pull/749

Handles this issue: https://github.com/NASA-IMPACT/veda-ui/pull/749#issuecomment-1814647907

### DIFF
--- a/app/scripts/components/common/blocks/embed.tsx
+++ b/app/scripts/components/common/blocks/embed.tsx
@@ -18,17 +18,23 @@ const IframeWrapper = styled.iframe`
 `;
 
 interface EmbedProps {
+  className?: string;
   src: string;
   height: number;
 }
 
-export default function Embed({ src, height = 800, ...props}: EmbedProps) {
+export default function Embed({
+  className,
+  src,
+  height = 800,
+  ...props
+}: EmbedProps) {
   if (!src) {
     throw new HintedError('Embed block requires a URL');
   }
 
   return (
-    <EmbedWrapper>
+    <EmbedWrapper className={className}>
       <BrowserFrame link={src}>
         <IframeWrapper loading='lazy' src={src} height={height} {...props} />
       </BrowserFrame>


app/scripts/components/common/blocks/embed.tsx
```diff
@@ -18,17 +18,23 @@ const IframeWrapper = styled.iframe`
...
+  height = 800,
+  ...props
+}: EmbedProps) {
   if (!src) {
     throw new HintedError('Embed block requires a URL');
   }
 
   return (
-    <EmbedWrapper>
+    <EmbedWrapper className={className}>
```
hanbyul-here, line 37: Is there a 'better' way of making a component overridable through styled components?

sandrahoang686: I would like to know this too

sandrahoang686 (approved):

app/scripts/components/common/blocks/embed.tsx
danielfdsilva: Passing the classname is one of the ways to do it. They increase in complexity depending on what you're trying to achieve.

Le me go through some examples with a simple component.

```js
function MyExample(props) {
  const { className } = props;
  return (
    <div className={className}>
      <h1>Simple Component</h1>
      <p>Some text</p>
    </div>
  );
}
```

In this way you can create a styled component from it and style it.

```js
const StyledMyExample = styled(MyExample)`
  background-color: red;
`;
```

This works and for most situations is enough. However it won't let you use the component as a selector.
I.e. the following **won't work**:

```js
const ParentComponent = styled.main`
  ${MyExample} {
    bakground-color: blue;
  }
`
```

Nor will you be able to change the component that get's rendered.
I.e. the following **won't work**:

```js
  <MyExample forwardedAs='section' />
```

To be able to use everything styled components have to offer on your custom component, it needs to return a styled component, and it itself needs to be a styled component.

```js
const MyExampleWrapper = styled.div`
  /* Styled wrapper. No styled needed here */ 
`;

function MyExampleComponent(props) {
  // Extract the props you need for your component and pass everything else to
  // the wrapper
  const { someProp, ...rest } = props;
  return (
    <MyExampleWrapper {...rest}>
      <h1>Simple Component</h1>
      <p>Some text</p>
    </MyExampleWrapper>
  );
}

const MyExample = styled(MyExampleComponent)`
  /* Styled component */
  background-color: teal;
`;
```

Now you can use it as a selector and change the component that get's rendered.
This now works:

```js
const ParentComponent = styled.main`
  ${MyExample} {
    bakground-color: blue;
  }
`;
```

```js
<MyExample forwardedAs='section' />
```

The last thing you may want to do is support refs for react's sake. So the last improvement would be to add `forwardRef`;

```js
const MyExampleWrapper = styled.div`
  /* Styled wrapper. No styled needed here */ 
`;

function MyExampleComponent(props, ref) {
  // Extract the props you need for your component and pass everything else to
  // the wrapper
  const { someProp, ...rest } = props;
  return (
    <MyExampleWrapper ref={ref} {...rest}>
      <h1>Simple Component</h1>
      <p>Some text</p>
    </MyExampleWrapper>
  );
}

const MyExample = styled(React.forwardRef(MyExampleComponent))`
  /* Styled component */
  background-color: teal;
`;
```

If you want the typed, and shorter version:

```tsx
const MyExampleWrapper = styled.div`
  /* Styled wrapper. No styled needed here */
`;

interface MyExampleProps {
  someProp: any;
}

const MyExample = styled(React.forwardRef<HTMLDivElement, MyExampleProps>(
  function MyExampleComponent(props, ref) {
    // Extract the props you need for your component and pass everything else to
    // the wrapper
    const { someProp, ...rest } = props;
    return (
      <MyExampleWrapper ref={ref} {...rest}>
        <h1>Simple Component</h1>
        <p>Some text</p>
      </MyExampleWrapper>
    );
  })
)`
  /* Styled component */
  background-color: teal;
`;
```

Here's an example of this being used (without the `forwardRef` though):

https://github.com/NASA-IMPACT/veda-ui/blob/44dbd82f0d62828bf1585cce64dfb4a3cf537a5c/app/scripts/components/datasets/s-explore/tile-link.tsx#L104-L143

I hope this helps! :)

danielfdsilva (approved):

danielfdsilva:
@hanbyul-here Is this good to merge?

hanbyul-here:
Yup sorry, got distracted by your insightful comment! :) will merge now.

PR #757, opened 2023-11-17
Add params to analysis attribute

### LINKED ISSUE #756: Add option for Analysis to use reprojection
To make use of the TiTiler `/statistics` capability to reproject data to an equal-area projection, such that statistics are accurate for non-equal-area source projections, we need to be able to pass the `dst_crs` URL parameter to `/statistics` calls, plus maybe the `reproject` parameter. (CRS = coordinate reference system)

- https://github.com/NASA-IMPACT/veda-architecture/issues/334

Here are the [API docs for that endpoint](https://ghg.center/api/raster/docs#/Cloud%20Optimized%20GeoTIFF/geojson_statistics_cog_statistics_post) (we are making `POST` requests).

Maybe this could be a front matter parameter in the [Analysis](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md#analysis) section?

image

## Acceptance criteria

- [ ] Layers can be configured to use a specific working CRS (`dst_crs`) in statistics calculations

hanbyul-here:
Close #756  Add `analysis.params` as a layer configuration to pass it to `/statistics` endpoint. 

🤚  I do not know how to test this functionality, I just checked the parameters get passed to the request. 
🤚  This also leaves me with a question if we need to merge this change to `main`, and bring all the changes that have been stacked to main to GHG, or make another feature branch for GHG. Considering there are not many changes stacked on main branch, I will say we just bring all the changes on main to GHG.

- [ ] Add documentation

### DIFF
--- a/app/scripts/components/analysis/results/timeseries-data.ts
+++ b/app/scripts/components/analysis/results/timeseries-data.ts
@@ -267,7 +267,7 @@ async function requestTimeseries({
       }
     });
 
-    const analysisParams = layersBase.layer.analysis?.params;
+    const analysisParams = layersBase.layer.analysis?.sourceParams;
 
     const layerStatistics = await Promise.all(
       assets.map(async ({ date, url }) => {
--- a/app/scripts/components/analysis/results/timeseries-data.ts
+++ b/app/scripts/components/analysis/results/timeseries-data.ts
@@ -267,6 +267,8 @@ async function requestTimeseries({
       }
     });
 
+    const analysisParams = layersBase.layer.analysis?.params;
+
     const layerStatistics = await Promise.all(
       assets.map(async ({ date, url }) => {
         const statistics = await queryClient.fetchQuery(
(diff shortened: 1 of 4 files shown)


j08lue:
Wow, that was quick!

Since this is backwards-compatible (if you do not specify `params`, nothing changes), this might as well get merged.

> I just checked the parameters get passed to the request

That's great, I can take care of testing more, once we have a preview against the GHG data, ideally.

app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -267,17 +267,19 @@ async function requestTimeseries({
       }
     });
 
+    const analysisParams = layersBase.layer.analysis?.params;
```
danielfdsilva, line 270: I'd propose to change this to `sourceParams` so it is the same pattern the layer uses - since it is the same type of content.

https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md

danielfdsilva (approved):
Left a suggestion. looks good otherwise

danielfdsilva (commented):
Looking good. Are you still adding documentation, or leaving it for later?

hanbyul-here:
Added documentation! @danielfdsilva

Issue #724, opened 2023-10-31
[E & A] Add option to retrieve WMTS and XYZ URLs for a layer


j08lue:
In the old Exploration UI, we added an option for users to retrieve the WMTS and XYZ URLs for the layer they were seeing on the map: 

Clicking on this button:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/56f2617a-5f27-4ae6-980f-2804b331bd9d)

would open this modal:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/c6636d6e-0b9e-4247-93a1-895c16c9fdeb)

In the new UI, this would make sense to have on the individual dataset "tracks" somehow - we need a design for this.

## Product context

This is a power feature that serves the goal of making data presented in VEDA interoperable: users should be able to continue an exploration or analysis they start in the VEDA Dashboard in other platforms they use, such as GIS or coding environments.

What the "layer URL" is for visual map exploration (panning and zooming around raster and vector layers), could be exporting time series generating URLs and/or data (as CSV or so) and/or a code snippet (Python/R) for a time series analysis they started in the UI. - This is just to say that the possibility of users jumping from the VEDA UI into other tools is something we ideally want to promote widely across the app.

## Acceptance criteria

- [x] Designed
- [x] Implemented
- [ ] Merged (https://github.com/NASA-IMPACT/veda-ui/pull/758)

Issue #755, opened 2023-11-17
[E & A] Respect dataset default projection and basemap


j08lue:
In the old Exploration view, we added the option that datasets configure a default map production to get shown in, like this one in our mock data:

https://github.com/NASA-IMPACT/veda-ui/blob/44dbd82f0d62828bf1585cce64dfb4a3cf537a5c/mock/datasets/no2.data.mdx#L54-L55

It does not seem like this property is used in the new E&A interface. Is that something to fix? It was a valuable feature for the GHG Center interface, where the data providers wanted to have a more accurate default depiction of the data, e.g. with Equal Earth.

## Acceptance criteria

- [x] Dataset layer projection and basemap can be configured in E&A

danielfdsilva:
It is still possible to change the projection, but it won't do it by default.
Now that multiple datasets can be added it is tricky to know which projection to use. Which dataset defines the projection? The user flow here can result in unexpected changes for the user, so we need to be careful

j08lue:
> Now that multiple datasets can be added it is tricky to know which projection to use

That makes sense. Let us think some more about that. 

GHG scientists were really not happy about Web Mercator, so perhaps a configurable global default could help. Or some logic which one of the per-dataset config we pick...

We can discuss that soon.

j08lue:
Latest idea: 

1. Projection and basemap are initialized to that of the **first** layer added to the map, to optimize for the most common (?) single-layer explore case, where we want users to view the data in the best possible way.
2. Projection and basemap do **not** change when user adds/removes/re-orders layers, to avoid the unexpected changes for the user @danielfdsilva warned about.

danielfdsilva:
@j08lue That's an interesting approach.
What happens if, when starting, the user adds 2 datasets that have different projections? Default to mercator, or pick one "at random" from those 2?

j08lue:
> Default to mercator, or pick one "at random" from those 2?

In which order are the layers currently added to the interface, actually, when a user selected multiple?

I'd use the properties from what ever ends up being the top-most layer. I am really optimizing for the case where the user adds a single layer / looks at a single layer (first) and make sure they get the intended / optimized view in that case. We can leave any more automagic solutions for maybe later. 😄

PR #758, opened 2023-11-17
Exploration features

danielfdsilva:
This PR adds the features to A&E recently discussed:

- [x] Zarr layer type support
- [x] Modal to get dataset layer tiles (through dataset options menu)
- [x] Show error on non supported datasets for analysis (non raster)
- [x] Fix a couple of nasty bugs
- [x] Add a feature flag to enable the new exploration
   If the site is built with any truthy value set on `FEATURE_NEW_EXPLORATION`, the app shows the page, otherwise all normal business
- [x] Fix the analysis state upon selecting / deselecting AOIs or changing parameters + adding datasets
- [x] Resolve race condition with URL parameters
- [ ] https://github.com/NASA-IMPACT/veda-ui/issues/761
- [ ] https://github.com/NASA-IMPACT/veda-ui/issues/755

### DIFF
--- a/app/scripts/components/exploration/components/datasets/dataset-options.tsx
+++ b/app/scripts/components/exploration/components/datasets/dataset-options.tsx
@@ -76,7 +76,7 @@ export default function DatasetOptions(props: DatasetOptionsProps) {
           <li>
             <Tip
               disabled={!!dataset.meta?.tileUrls}
-              content='Current dataset does not offer tile urls'
+              content='This dataset does not offer tile URLs'
               placement='left'
             >
               <TileModalButton
@@ -84,7 +84,7 @@ export default function DatasetOptions(props: DatasetOptionsProps) {
                 type='button'
                 onClick={() => setTileModalRevealed(true)}
               >
-                <CollecticonMap title='Open tile url options' /> View tile urls
+                <CollecticonMap title='Open tile URL options' /> Load into GIS
               </TileModalButton>
             </Tip>
           </li>
--- a/app/scripts/components/exploration/components/datasets/dataset-options.tsx
+++ b/app/scripts/components/exploration/components/datasets/dataset-options.tsx
@@ -48,6 +48,7 @@ export default function DatasetOptions(props: DatasetOptionsProps) {
     <>
       <Dropdown
         alignment='right'
+        direction='up'
         triggerElement={(props) => (
           <Button
             variation='base-text'
--- a/app/scripts/components/exploration/components/datasets/dataset-options.tsx
+++ b/app/scripts/components/exploration/components/datasets/dataset-options.tsx
@@ -1,14 +1,19 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import { PrimitiveAtom, useAtomValue, useAtom } from 'jotai';
 import 'react-range-slider-input/dist/style.css';
 import styled from 'styled-components';
 import { glsp, themeVal } from '@devseed-ui/theme-provider';
 import { Dropdown, DropMenu, DropTitle } from '@devseed-ui/dropdown';
 import { Button } from '@devseed-ui/button';
-import { CollecticonCog, CollecticonTrashBin } from '@devseed-ui/collecticons';
+import {
+  CollecticonCog,
+  CollecticonMap,
+  CollecticonTrashBin
+} from '@devseed-ui/collecticons';
 import { Overline } from '@devseed-ui/typography';
 
 import AnalysisMetrics from './analysis-metrics';
+import { TileUrlModal } from './tile-link-modal';
 
 import DropMenuItemButton from '$styles/drop-menu-item-button';
 import { SliderInput, SliderInputProps } from '$styles/range-slider';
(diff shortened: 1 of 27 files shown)


danielfdsilva:
For now I reenabled the exploration path so you can test it, but you'll need to go there directly (like done so far). We can then completely lock it off.

danielfdsilva:
Fix #724

app/scripts/components/exploration/components/datasets/dataset-options.tsx
```diff
@@ -30,53 +37,87 @@ export default function DatasetOptions(props: DatasetOptionsProps) {
...
-          </Tip>
-        </li>
-      </DropMenu>
-    </Dropdown>
+              <TileModalButton
+                visuallyDisabled={!dataset.meta?.tileUrls}
+                type='button'
+                onClick={() => setTileModalRevealed(true)}
+              >
+                <CollecticonMap title='Open tile url options' /> View tile urls
```
j08lue, line 87:
```suggestion
                <CollecticonMap title='Open tile URL options' /> Load into GIS
```

app/scripts/components/exploration/components/datasets/dataset-options.tsx
```diff
@@ -30,53 +37,87 @@ export default function DatasetOptions(props: DatasetOptionsProps) {
...
+        </DropMenu>
+        <AnalysisMetrics
+          activeMetrics={activeMetrics}
+          onMetricsChange={(m) => setSetting('analysisMetrics', m)}
+        />
+        <DropMenu>
+          <li>
+            <Tip
+              disabled={!!dataset.meta?.tileUrls}
+              content='Current dataset does not offer tile urls'
```
j08lue, line 79:
```suggestion
              content='This dataset does not offer tile URLs'
```

app/scripts/components/exploration/components/datasets/tile-link-modal.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <ModalHeadline>
+          <Overline>Dataset: {datasetName}</Overline>
+          <Heading size='small'>Use with GIS software</Heading>
+        </ModalHeadline>
+      )}
+      content={
+        <Form>
+          {[
+            { label: 'XYZ Tile Url', value: tileUrls?.xyzTileUrl },
+            { label: 'WMTS Tile Url', value: tileUrls?.wmtsTileUrl }
```
j08lue, line 53:
```suggestion
            { label: 'XYZ Tile URL', value: tileUrls?.xyzTileUrl },
            { label: 'WMTS Tile URL', value: tileUrls?.wmtsTileUrl }
```

j08lue (changes requested):
I tested `/exploration` but just in the preview, so without enabling that feature with the feature flag.

This is what I tested:

1. Open `/exploration`, add CMIP6 (Zarr/Kerchunk/TiTiler-Xarray) dataset -> "Oh no, something went wrong" ❌ 
2. Look up layer tile URLs -> worked nicely, wordsmithy changes suggested below. ✅ (Bonus: A more GIS-like icon could be something like [this](https://www.vectorstock.com/royalty-free-vector/global-export-icon-outline-style-vector-25200438))
3. Tried to run analysis on a vector dataset (EIS `Fire` from mock layers) -> ❌ No error/warning is shown, analysis status looks like success, time series graph area is blank (neither graph nor timeline grey nuggets)

Please let me know whether my testing was wrong or so or you could use more details.

danielfdsilva:
1) This is a data problem.
    - Locally I was using the `https://test-stac.delta-backend.com` which returns `summaries: null`. When this happens we default to `extent.temporal.interval` and so it was working.
    - The preview, however was using `https://staging-stac.delta-backend.com` which returns `summaries: { datetime: [null, null] }`, which causes it to break.
    - I added a check for this case, but **we need data guardrails in place**. The app needs the summaries (and other properties) to be able to function, so they need to be provided. If this is not possible, we need to define another way to get the required data - but it has to be consistently available.  

---

3) There was a problem with the error catching which is now fixed. I also added an additional message to inform the user when an analysis did not return data.
image

j08lue (changes requested):
## Previous review addressed

Vector datasets and the TiTiler-Xarray-served CMIP6 dataset now work fine!


## New issue

Another observation during testing, though: The "Outdated" state of the temporal and spatial selection for analysis does not behave as expected:

A) Moving geometry does not seem to cause "outdated"

1. Add dataset
2. Draw AOI
3. Keep date range as-is
4. Start analysis and wait for completion
5. State:  ✅ up-to-date - correct
6. **Move or change geometry**
7. State: ✅ up-to-date - wrong  (should be outdated)
8. Deselect and select the geometry again
9. State: ❌ outdated - correct

B) Changing date range causes permanent "outdated"

1. Add dataset
2. Draw AOI
3. **Change date range with sliders on timeline**
4. Start analysis and wait for completion
5. State: ❌ outdated - wrong (should be up-to-date)

These are reproducible for me on a clean browser - could still be me somehow...

danielfdsilva:
@j08lue You reported the bugs very clearly with perfect reproducibility steps! 🎉 

Just pushed a fix that addressed both things, among other small problems! Let me know if you find anything else!

j08lue (changes requested):
## Previous review addressed

Awesome, the up-to-date/outdated state now behaves as expected in the situations I tested before.

## New find

When adding datasets to an existing analysis, users do not get an obvious option to run the analysis on those:

1. Run analysis on some dataset(s)
2. State: ✅ up-to-date - correct
3. Add more datasets
4. State: ✅ up-to-date - ?? how would a user now extend the same analysis to the added datasets? Should we use "Outdated" when they added new datasets? Or somehow give them another possibility to add these datasets to the analysis (e.g. on the timeline or something)?

image

danielfdsilva:
@j08lue yet another great find :p
The correct behavior should be that the analysis kicks off automatically. I applied the needed fix.

j08lue (approved):
All review comments addressed.

Remaining from the PR check list are the new requests
- https://github.com/NASA-IMPACT/veda-ui/issues/761
- https://github.com/NASA-IMPACT/veda-ui/issues/755

Those could also be for a new PR, if you like.

PR #768, opened 2023-11-29
Add line breaks to mock disclaimer text

j08lue:
so much for culture...


danielfdsilva (changes requested):
ahahah.

With markdown you use double space at the end of a sentence to trigger a line break

PR #769, opened 2023-12-04
Catch missing summaries on api data

danielfdsilva:
Fix https://github.com/NASA-IMPACT/veda-config/issues/353

### DIFF
--- a/app/scripts/context/layer-data.tsx
+++ b/app/scripts/context/layer-data.tsx
@@ -55,6 +55,10 @@ const fetchLayerById = async (
       ? data.summaries.datetime
       : data.extent.temporal.interval[0];
 
+    if (domain.some((d) => !d)) {
+      throw new Error('Invalid datetime domain');
+    }
+
     return {
       timeseries: {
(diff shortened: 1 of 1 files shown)


j08lue (approved):
Not sure how I can test this (using the Netlify preview), since the dataset in question does not seem to be in the mock data.

I get a critical error on the [OCO2 dataset](https://github.com/NASA-IMPACT/veda-ui/blob/fix/missing-summaries/mock/datasets/oco2-geos-l3-daily.data.mdx), though - is that related?

https://deploy-preview-769--veda-ui.netlify.app/data-catalog/oco2-geos-l3-daily/explore

I'll approve here anyways, since the change is so small.

danielfdsilva:
@j08lue It is related.

We look for the data domain under `summaries.datetime`, and when not available, we default to `extent.temporal.interval`. In that dataset's case they're both null, so there's nothing we can do. I did however handle the error better, so only the layer fails, and it is not a critical one.

image

image

Issue #741, opened 2023-11-09
[E&A] Review analysis messages


danielfdsilva:
This is a job for pro designers. 🙇  If any are available, please chime in or take over. If not, please consider my proposal below.

## Proposal

1. When user made a selection:
   
   <img height="25px" src="https://github.com/NASA-IMPACT/veda-ui/assets/3404817/2375a068-8104-4edb-8567-4c7092135f85">

2. When the selection changed while there is an active analysis:

   <img height="25px" src="https://github.com/NASA-IMPACT/veda-ui/assets/3404817/f8b56025-a8e1-492c-af04-30b5a639ac83">

3. When the user reset the selection but there is still a time series on the bottom drawer:

   <img height="25px" src="https://github.com/NASA-IMPACT/veda-ui/assets/3404817/18c4b91f-d43c-42d4-8d76-d94a41671561">

## Changes with respect to current implementation

1. No "warning"-level alert to users - changed analysis parameters that are not yet applied is not a problem they need to be alerted to, IMO.
2. Use the same "timeseries" icon on the message head to indicate at that the message relates to the zonal statistics / analysis
3. Spell out button functions not only for initial "Analyze" but also the other options ("Apply Changes" and perhaps also "Exit") for more consistency and clarity
4. Updated message texts for the two cases of (1) changed selection and (2) no active selection

_Originally posted by @j08lue in https://github.com/NASA-IMPACT/veda-ui/issues/733#issuecomment-1803482867_

danielfdsilva:
1. If a user selects a different feature and gets distracted, they may think that what's on the chart refers to that area. Perhaps we should err on the side of caution.
3. Using full text buttons makes the bar pretty long, which is already long to start with. Personally I also don't think we should be spelling out everything. Could be a matter of choosing better icons, but will defer to designers.

cc @faustoperez and @ricardoduplos for some input

j08lue:
Ok, please ignore my concrete suggestions regarding buttons and icons

The important thing for me was the wording ("Selection changed" instead of "Outdated"). I just spelled out the call to action ("Apply Changes" and "Exit") for our own clarity. How you design this is totally up to you of course.

j08lue:
Closing this so nobody gets angry at me for playing with Figma. 😈

faustoperez:
These are the updated analysis messages, without changing the current analysis flow functionality. This is the general approach:

- Make actions explicit by having a primary button attached to them
- Review the labels on the analysis banners
[...]

PR #772, opened 2023-12-04
Update analysis messages

danielfdsilva:
Closes https://github.com/NASA-IMPACT/veda-ui/issues/741

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
@@ -9,7 +9,7 @@ import {
   CollecticonTrashBin,
   CollecticonUpload2
 } from '@devseed-ui/collecticons';
-import { Toolbar, ToolbarLabel } from '@devseed-ui/toolbar';
+import { Toolbar, ToolbarLabel, VerticalDivider } from '@devseed-ui/toolbar';
 import { Button } from '@devseed-ui/button';
 import { themeVal, glsp, disabled } from '@devseed-ui/theme-provider';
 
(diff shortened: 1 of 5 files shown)


j08lue (changes requested):
A few differences from the [designs](https://github.com/NASA-IMPACT/veda-ui/issues/741#issuecomment-1833607303):

## 1. The cases "one unselected dataset" and "multiple unselected datasets" 

Should be:

image

Is (in both cases):

image


## 2. Analysis tools label should be capitalized (is all caps) and separated by spacer

Should be:

image

Is:

image

## 3. "Selection changed" message is not vertically centered

Should be:

image

Is (slightly lower - you see it best when the message changes in the app):

image

danielfdsilva:
@j08lue Addressed!

faustoperez (changes requested):
Looking good Daniel!

Just one small thing: when an analysis is active, I think that deleting the shape should have the same effect as exiting the analysis. I.e. not getting the "Selection changed" message when deleting the shape, but making the two buttons behave the same.

I know that this changes a bit the current behaviour (not just a styling thing), what do y'all think?

image

j08lue (approved):
Thanks for addressing my review so thoroughly! The messaging and layout looks great and works as expected through different scenarios.

@faustoperez you suggested that we adjust the behavior towards deleting last shape = exit analysis. In the interest of getting this deployed for demo and further testing as soon as possible, I recommend that we keep that for later and observe more users throughout the refined flow and see how they deal with these steps of the flow.

So, if you are okay with that, @faustoperez, I'd say let us keep the flow as-is for now.

faustoperez:
> So, if you are okay with that, @faustoperez, I'd say let us keep the flow as-is for now.

Sounds great Jonas!

Issue #742, opened 2023-11-09
Document features added for EIC that are missing docs


hanbyul-here:
We are adding new functionality to VEDA UI to accommodate EIC needs. In the interest of the very short deadline, we are postponing to document some of these new features. We should catch up on that as soon as possible after the release. This ticket is to keep track of what we are missing to document:

- [ ] Component, Content override for hub index page
- [ ] External link (`asLink`)
- [ ] stacApiEndpoint, tileEndpoint for layer
- [ ] mainNavItem, navTitle for userpage

## Acceptance criteria

- [ ] Documented all features added during the EIC sprint that were missing docs

PR #773, opened 2023-12-05
Add documentations for the features introduced by eic

hanbyul-here:
Close https://github.com/NASA-IMPACT/veda-ui/issues/742

+ I get the intention of separating contents of the dataset layer (https://github.com/NASA-IMPACT/veda-ui/blob/213d9c25a469121d2c420d0e952042bb596b855f/docs/content/CONTENT.md#datasets) from the data(STAC) related properties https://github.com/NASA-IMPACT/veda-ui/blob/213d9c25a469121d2c420d0e952042bb596b855f/docs/content/frontmatter/layer.md but this ends up being confusing. Would it be better just to merge the dataset layer properties into one file?

### DIFF
--- a/docs/content/CONTENT.md
+++ b/docs/content/CONTENT.md
@@ -194,6 +194,9 @@ related: Related[]
   - type: string
     id: string
     thematic: string
+
+asLink:
+  url: string;
 ---
 
 <Block>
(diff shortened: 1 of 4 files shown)


docs/content/CONTENT.md
```diff
@@ -263,3 +266,13 @@ Each content should be formatted like below
   - type: story
     id: story-id
 ```
+
+**asLink**
+
+If the story doesn't hold any contents but the link to outside, use `asLink` property to indicate it. Note that `id` 
+of the story that has `asLink` won't generate the url for the story.
```
sandrahoang686, line 273: might be better to say `but links externally` instead

hanbyul-here: I wonder that gives a wrong impression that `id` of the story works as a link?  `id` doesn't really do anything (at least for users, but still needed internally) when the story has `asLink`

PR #781, opened 2023-12-18
update styles based on hideBrowserFrame prop

### LINKED ISSUE #778: Add option to remove browser frame from Embed Block
I think this should just require a prop being passed down to: 

https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/styles/browser-frame.tsx

Or having an alternative styled component to use when a bare embed is requested. 

"it's confusing, it looks like my safari window has a baby " - David Rager

:\

stephenkilbourn:
closes #778 

This adds a prop to the existing `BrowserFrameComponent` that will hide the controls and change the padding which gives the illusion of the gray bar at the bottom of a browser.  By default, instances of the component already used will be unchanged, so only passing that prop as true will change things

Here's a screenshot with the `<BrowserFrame hideBrowserFrame={true}>` setting on the first occurrence of the component, but the second one is untouched.

Screenshot 2023-12-18 at 12 51 48 PM

### DIFF
--- a/app/scripts/styles/browser-frame.tsx
+++ b/app/scripts/styles/browser-frame.tsx
@@ -3,11 +3,14 @@ import styled from 'styled-components';
 import { themeVal } from '@devseed-ui/theme-provider';
 import { CollecticonExpandTopRight } from '@devseed-ui/collecticons';
 
-function BrowserFrameComponent(props: { children: ReactNode; link?: string }) {
-  const { children, link, ...rest } = props;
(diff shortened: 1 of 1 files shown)


hanbyul-here:
Thanks for your contribution. I was wondering if we need to rethink the design for the embed browser itself, it is confusing to users. I would like an input from @faustoperez  for this issue.

faustoperez:
> I would like an input from @faustoperez for this issue.

Thanks @hanbyul-here , I self-assigned issue #785 , will have a look!

Issue #712, opened 2023-10-25
Replace duplicated map component


danielfdsilva:
When starting to work on the new Analysis+Exploration page, the team decided to build the map component from scratch to improve the layer management process (given that now the map needed to support multiple simultaneous layers), and the overall code. 

The different blocks (and other app sections) that use the [old map](https://github.com/NASA-IMPACT/veda-ui/tree/1c5bf02cd5ff6246844b9d43a7f6dce05fc8a28c/app/scripts/components/common/mapbox) should be refactored to use the [new map](https://github.com/NASA-IMPACT/veda-ui/tree/main/app/scripts/components/common/map) and the old code should be removed from the app.

When doing this some consideration should go into the [map block](https://github.com/NASA-IMPACT/veda-ui/blob/1c5bf02cd5ff6246844b9d43a7f6dce05fc8a28c/app/scripts/components/common/blocks/block-map.tsx) and consider whether its behavior should change a little, since the new map:
- supports multiple layers rather than just one
    - should this be in the block?
    - how many layers are reasonable given the need for legends (which btw are not a feature of the new map)
    - should layer configuration (like opacity) be allowed?
- Layer comparison is now manually defined instead of through the dataset configuration
    - Compare date was a property, but compare layer wasn't

These are just some topics to consider when doing this, but some more discussion should go into this.

sandrahoang686:
**Notes:**
* Maps embedded in stores use old map component
* New map component -> app/scripts/components/common/map
* Old map component -> app/scripts/components/common/mapbox

**Plan:**
* Identify where the old map components and where in the codebase replace
* Replace old with new and just get it works with no new functionality
* And then be able to support multiple layers

sandrahoang686:
1/23/24 Checkin: Focus on story blocks only

PR #784, opened 2024-01-02
Fix user defined position setup for block maps (when compare slider is used)

hanbyul-here:
Close https://github.com/NASA-IMPACT/veda-config/issues/352

+ While looking at this issue, I realized that we are retiring the way of using the layers for the exploration page, but not for the stories. I will make a ticket with more details, but we should consolidate how the layers/maps are used throughout the application.

### DIFF
--- a/app/scripts/components/common/mapbox/index.tsx
+++ b/app/scripts/components/common/mapbox/index.tsx
@@ -264,7 +264,6 @@ function MapboxMapComponent(
     baseTimeDensity,
     compareTimeDensity
   ]);
-
   return (
     <>
       {/*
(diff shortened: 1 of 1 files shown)


hanbyul-here:
It turned out that the ticket is already filed: https://github.com/NASA-IMPACT/veda-ui/issues/712 😅

j08lue (approved):
I can confirm that this fixes the issue that compare slider embeds do not keep the initial position.

I compared the embedded maps between 
1. Before: https://veda-ui.netlify.app/stories/air-quality-and-covid-19
2. After: https://veda-ui.netlify.app/stories/air-quality-and-covid-19

Issue #746, opened 2023-11-13
Prepare analysis page for the datasets without summaries


hanbyul-here:
All the collections are supposed to have `summaries`, but sometimes they are missing this breaks the analysis page totally. We need some basic error handling. 

@danielfdsilva In case this is related to the new a&e page.

j08lue:
Oh, awesome you caught that, @hanbyul-here!

I will create a ticket in https://github.com/US-GHG-Center/ghgc-data to fix those collections, too.

Can you name one collection where this issue exists?

sandrahoang686:
Thoughts
* How do we want to handle missing summaries?
* Identify through out in the application where we handle summaries

PR #786, opened 2024-01-03
Handle analysis/define datasets w/o collection summaries

sandrahoang686:
This aims to solve ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/746)

Exploration and Analysis page seems like it already handles summaries are not available [here](https://github.com/NASA-IMPACT/veda-ui/pull/769/files) and [here](https://github.com/NASA-IMPACT/veda-ui/pull/758/files#diff-cf038be4360f19e9ab0d1ea21567952a3c816545a0c23e57906e0f7223a5e769R61)

For Analysis/Results: We dont have to worry because users would not be able to select the dataset without summaries any longer with this fix

Loom Local Validation: 
https://www.loom.com/share/d0b132ca88b242dc8014e277495d6746

**UPDATE TO FLAG COMMENT:**
Screenshot 2024-01-04 at 12 34 47 PM

### DIFF
--- a/app/scripts/components/analysis/define/use-stac-collection-search.ts
+++ b/app/scripts/components/analysis/define/use-stac-collection-search.ts
@@ -20,6 +20,13 @@ interface UseStacSearchProps {
   aoi?: FeatureCollection<Polygon> | null;
 }
 
+interface DatasetWithCollections extends DatasetLayer {
+  isPeriodic: boolean,
+  timeDensity: string,
+  domain: string[],
+  timeseries?: string[];
+}
+
 export type DatasetWithTimeseriesData = TimeseriesDataResult &
   DatasetLayer & { numberOfItems: number };
 
@@ -64,14 +71,15 @@ export function useStacCollectionSearch({
     enabled: readyToLoadDatasets
   });
 
-  const datasetLayersInRange = useMemo(() => {
+  const [datasetLayersInRange, invalidDatasets] = useMemo(() => {
     try {
-      return getInTemporalAndSpatialExtent(result.data, aoi, {
+      const [datasetWithSummaries, datasetWithoutSummaries] = getInTemporalAndSpatialExtent(result.data, aoi, {
         start,
         end
       });
+      return [datasetWithSummaries, datasetWithoutSummaries];
     } catch (e) {
-      return [];
+      return [[], []];
     }
   }, [result.data, aoi, start, end]);
 
@@ -89,11 +97,13 @@ export function useStacCollectionSearch({
     );
   }, [datasetLayersInRangeWithNumberOfItems]);
 
-  const unselectableDatasetLayers = useMemo(() => {
+  let unselectableDatasetLayers: DatasetWithTimeseriesData[] | DatasetWithCollections[] = useMemo(() => {
     return datasetLayersInRangeWithNumberOfItems.filter(
       (l) => l.numberOfItems > MAX_QUERY_NUM
     );
   }, [datasetLayersInRangeWithNumberOfItems]);
+  
+  if (invalidDatasets) unselectableDatasetLayers = unselectableDatasetLayers.concat(invalidDatasets);
 
   return {
     selectableDatasetLayers,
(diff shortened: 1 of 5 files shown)


app/scripts/components/analysis/results/chart-card.tsx
```diff
@@ -101,7 +101,7 @@ const getNoDownloadReason = ({ status, data }: TimeseriesData) => {
   if (status === 'loading') {
     return 'Download will be available once the data finishes loading.';
   }
-  if (!data.timeseries.length) {
+  if (!data.timeseries?.length) {
```
sandrahoang686, line 104: 📓 : Optionally chaining for type safety and ts complaints but not rendering the datasets in the `define` step protects against this already. Users should not be able to render results of datasets without summaries through ui flow

app/scripts/components/analysis/results/index.tsx
```diff
@@ -130,7 +130,7 @@ export default function AnalysisResults() {
   const availableDomain: [Date, Date] | null = useMemo(() => {
     if (!start || !end) return null;
     const onlySingleValues = requestStatus.every(
-      (rs) => rs.data?.timeseries.length === 1
+      (rs) => rs.data?.timeseries?.length === 1
```
sandrahoang686, line 133: 📓 : Same as above

app/scripts/components/analysis/define/use-stac-collection-search.ts
```diff
@@ -89,11 +91,13 @@ export function useStacCollectionSearch({
...
   }, [datasetLayersInRangeWithNumberOfItems]);
 
-  const unselectableDatasetLayers = useMemo(() => {
+  let unselectableDatasetLayers: DatasetWithTimeseriesData[] = useMemo(() => {
     return datasetLayersInRangeWithNumberOfItems.filter(
       (l) => l.numberOfItems > MAX_QUERY_NUM
     );
   }, [datasetLayersInRangeWithNumberOfItems]);
+  
+  if (invalidDatasets.length) unselectableDatasetLayers = unselectableDatasetLayers.concat((invalidDatasets as unknown) as DatasetWithTimeseriesData[]);
```
hanbyul-here, line 100: wouldn't it be more concise to 

```suggestion
  return [...unselectabledatasetLayers, ...invalidDatasets]
```

app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -35,7 +35,7 @@ export interface TimeseriesDataResult {
   isPeriodic: boolean;
   timeDensity: TimeDensity;
   domain: string[];
-  timeseries: TimeseriesDataUnit[];
+  timeseries?: TimeseriesDataUnit[]; // NOTE: Summaries on collections will not always be available
```
hanbyul-here, line 38: To make the type useful, we seem to need a new type. The components like `chart-card` has to have timeseries data, so it doesn't seem to reflect is actual usage to type `timeseires` as an optional. (Also the data with summaries and wo summaries having the same type is confusing, like this line : https://github.com/NASA-IMPACT/veda-ui/pull/786/files#diff-73d60515e81c1007b36e13ea2886f3b4e1457746381934d40af7802e39841022R182) 

How about leaving the `TimeseriesDataResult` as it is, but making `TimeseriesDataWithMissingSummaries` (I just made the variable name up) and use this type explicitly for the data without summaries? 

More food for thought:  when the data is missing summaries, we don't seem to need any attributes from the collection?

hanbyul-here (changes requested):
I left some thoughts about type!

app/scripts/components/analysis/define/use-stac-collection-search.ts
sandrahoang686: lol yes

app/scripts/components/analysis/results/timeseries-data.ts
sandrahoang686: I like the idea of being explicit but if we dont seem to need any attributes from the collection when summaries is missing from the data, then we can probably just use type `DatasetLayer`. Will make this change

app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -35,7 +35,7 @@ export interface TimeseriesDataResult {
   isPeriodic: boolean;
   timeDensity: TimeDensity;
   domain: string[];
-  timeseries: TimeseriesDataUnit[];
+  timeseries: TimeseriesDataUnit[]; // NOTE: Summaries on collections will not always be available
```
hanbyul-here, line 38: Is this note still necessary?

app/scripts/components/analysis/define/use-stac-collection-search.ts
```diff
@@ -156,14 +160,25 @@ function getInTemporalAndSpatialExtent(collectionData, aoi, timeRange) {
...
       timeDensity: collection['dashboard:time_density'],
       domain: collection.extent.temporal.interval[0],
-      timeseries: collection.summaries.datetime
+      timeseries: collection.summaries.datetime,
     };
   });
-
-  return filteredDatasetsWithCollections;
+  
+  const [collectionsWithSummaries, collectionsWithoutSummaries]: [DatasetWithCollections[], DatasetLayer[]] = filteredDatasetsWithCollections.reduce((result: [DatasetWithCollections[], DatasetLayer[]], d: DatasetWithCollections) => {
```
hanbyul-here, line 178: I think we have some opportunities to make the code simple/coherent here.
So currently, the logic flows in a way 
1. get the collection data 2. filter the collection with too many requests in the hook. (near line90)

How do you think about following the current pattern? so `getInTemporalAndSpatialExtent` will return the collections, and we will filter out the collections' missing summaries in the hook, near line 90. 

Then the code can be 

```
const datasetWithTooManyRequests = ....
const datasetMissingSummaries = ....
const unselectableDatasetLayers = [...datasetWitlTooManyRequests, ...datasetMissingSummaries]
```
How do you think?

hanbyul-here:
I made a pr against this branch with some ts change suggestions: https://github.com/NASA-IMPACT/veda-ui/pull/797 Please take a look!

hanbyul-here (approved):
:shipit: !

PR #788, opened 2024-01-06
set up lint-staged and husky

### LINKED ISSUE #787: Set up a precommit hook for eslint/ts error

stephenkilbourn:
closes #787 

This adds husky and lint-staged to run the lint and typecheck on only staged files. I had to pin lint-staged at an older version because they dropped support of node 16 with their last major release. Testing it out things work for me.  I think it should be noted that anyone modifying an existing file that already has errors will need to either fix them as part of their MR or bypass the hook.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,3 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+yarn lint-staged
--- a/docs/development/SETUP.md
+++ b/docs/development/SETUP.md
@@ -20,6 +20,7 @@ Install Node modules:
 ```
 yarn install
 ```
+This command  will install all the dependencies and devtools needed for local development. 
 
 ## Usage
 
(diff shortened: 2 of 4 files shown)


hanbyul-here:
Thank you so much for your contribution!
I just gave it a try. I wonder 1. if we can configure it to show where errors/warnings are so it helps me to fix them. 2. if we can be selective about whether to ignore the warnings (only care about errors - I can't commit with the current configuration now which I am assuming because of lint warnings - but then it is difficult to spot where those warnings are from the terminal.).

stephenkilbourn:
@hanbyul-here - ggod catch.  lint-staged was actually failing and swallowing the error. I should have tested the happy path better 🤦 

I added the missing `yarn` in the command.  I also explicitly call `eslint` and `stylelint ` on separate lines.  The existing `lint` script with them command overrides the lint-staged feature by passing the apps/scripts path.  If I call that, it runs the lint against unstaged files, too

sandrahoang686 (approved):
Pulled and ran through a test commit, lgtm

hanbyul-here (approved):
I made a small documentation pr as a follow-up: https://github.com/NASA-IMPACT/veda-ui/pull/798 
Thanks for your contribution again 🙇

hanbyul-here:
I noticed that ts-check in the pre-commit stage gives an error incorrectly. 
ex. When I made a misc change to `common/layout-root`, it throws an error that 'app/scripts/components/common/layout-root.tsx(19,37): error TS2307: Cannot find module '$utils/use-google-tag-manager' or its corresponding type declarations.' 
It seems that precommit check only have access to the staged files? Do you know how we can address this problem? @stephenkilbourn ?

PR #794, opened 2024-01-08
Docs about zoom extent

### LINKED ISSUE #754: zoomExtent
There are 2 separate issues for `zoomExtent`.

1. Document actual behavior of `zoomExtent` : https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md
As far as I can check, maxZoom that user specifies through `zoomExtent` doesn't do anything in the code. 

2. `minzoom`, `maxzoom` as sourceParams:
tilejson will specify the minzoom and maxzoom once they are passed as source params. Encourage users to use minzoom and maxzoom so VEDA doesn't request tiles over maxzoom.

3. Investigate raster rendering behavior in maboxgl
🔍 I tried passing. minzoom, maxzoom through source params. I can see that thetiles are not being fetched, but MapboxGL does render some kinds of 'blobs' when the raster layer is overzoomed. (video attached) I wonder if this is from mapboxgl's default interpolation problem? 

https://github.com/NASA-IMPACT/veda-ui/assets/4583806/535c3d27-cbc4-47e9-b223-8e4ee266e6d1

Thanks @slesaad for reporting 🙇

hanbyul-here:
Close #754 

<strike> @slesaad This PR changes the default way of resampling the raster when it is upsampled. Please check if this configuration works with the gridded dataset, and also the other datasets and if it would be ok to change the default method to this one. </strike>

As it turned out it was data problem, I will hold the code change and merge only doc change.

### DIFF
--- a/docs/content/frontmatter/layer.md
+++ b/docs/content/frontmatter/layer.md
@@ -68,7 +68,8 @@ Define the initial date that is shown when enabling a timeseries layer. This val
 
 **zoomExtent**  
 `[int, int] | fn(bag)`  
-Minimum and maximum zoom values for the layer. Below the minimum zoom level the layer will not be shown, but markers will be displayed to indicate where data is available.
+Minimum and maximum zoom values for the layer. Below the minimum zoom level the layer will not be shown, but markers will be displayed to indicate where data is available. 
+For raster and zarr layer, only minimum zoom will be respected and the maximum zoom will be automatically fetched from STAC endpoint.
 
 **sourceParams**  
 `object`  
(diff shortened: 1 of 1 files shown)


docs/content/frontmatter/layer.md
```diff
@@ -80,6 +81,11 @@ These values may vary greatly depending on the layer being added but some may be
 - **colormap_name**  
   `string`  
   The colormap to use for the layer. One of https://cogeotiff.github.io/rio-tiler/colormap/#default-rio-tilers-colormaps
+- **minzoom**
+The minimum zoom of the layer. The tiles below the minzoom won't be fetched from titiler.
+- **maxzoom**
+The maximum zoom of the layer. The tiles over the maxzoom won't be fetched from titiler.
```
j08lue, line 87:
```suggestion
- **minzoom**
The minimum zoom of the layer. Map tiles below the minzoom will not be loaded/shown.
- **maxzoom**
The maximum zoom of the layer. Map tiles over the maxzoom will not be loaded/shown.
```

j08lue (approved):
Good addition to clarify the minzoom/maxzoom behavior.

PR #805, opened 2024-01-18
Cmr layer 

hanbyul-here:
- I ended up leaving the main logic as it is (CMR layer doesn't call CMR endpoint directly - There seem to be some cases that stac collection id doesn't match with cmr's short_name? ex. https://cmr.earthdata.nasa.gov/search/collections.umm_json?short_name=TRMM_3B42_Daily.v7&version=07 - I get an empty response _ all in all, I thought it is better to keep the source of truth in one place for now.) 

- Make CMR layer (that shares zarr paint layer with Zarr layer)
- We currently have two map components (which we should consolidate in the near future related issue: #712 ) This is why there are two places for the layers. 
- CMR layer won't show up for Analysis. (The timeline will fail for new E&A. User can still explore the map.- you can check 
new E&A page by putting a feature flagging variable in `.env`: `FEATURE_NEW_EXPLORATION = 'TRUE'`. I also attached a screenshot below.)

Screen Shot 2024-01-19 at 9 27 09 AM


abarciauskas-bgse (approved):
Thank you @hanbyul-here ! It looks good to me. I have 2 questions which are non-blocking:
1. can you remind me the purpose of style-generators?

hanbyul-here:
1. style-generator styles for Mapbox, that eventually gets passed into mapinstance to update styles for new parameters etc.
2. Yes, they are very similar to each other. It is a part of tech debt that we have to solve sooner, even though it will be more like getting rid of https://github.com/NASA-IMPACT/veda-ui/blob/cmr-layer-sketch/app/scripts/components/common/mapbox/layers/zarr-timeseries.tsx.  Related issue: https://github.com/NASA-IMPACT/veda-ui/pull/805 the code in `mapbox` is OG one, used when only one layer is allowed. As new A&E page requires multiple layers, we moved to the new one in `map`. So the new A&E page uses `map,` while the current explore (and mdx maps) uses `Mapbox.`. 
We are putting more effort towards making everything use the new way of using maps, rather than doing something with the old way of using maps. (because they don't need to exist together.)

abarciauskas-bgse:
@hanbyul-here Thanks for the explanation! I still don't fully understand "style-generator styles for Mapbox, that eventually gets passed into mapinstance to update styles for new parameters etc." but that explanation is good for now and if I need to understand it in the future I will dig in more.

PR #806, opened 2024-01-19
777 critical error link

hanbyul-here:
Added a more explicit link to the homepage from the critical error page.

Screen Shot 2024-01-19 at 11 16 34 AM

### DIFF
--- a/app/scripts/components/uhoh/fatal-error.js
+++ b/app/scripts/components/uhoh/fatal-error.js
@@ -119,6 +119,12 @@ function Child(props) {
                 <p>{nl2br(error.stack)}</p>
               </details>
             )}
+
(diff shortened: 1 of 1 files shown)


j08lue:
Generally looks good to me. Is the email address `email@example.com` replaced by a real one in instances?

hanbyul-here:
The example email will be replaced by whichever email that configuration set up. ex. EIC https://github.com/NASA-IMPACT/veda-config-eic/blob/develop/.env#L8 

I got rid of the email section: 

Screen Shot 2024-01-19 at 11 41 46 AM

j08lue (approved):
I think this offers the way out for users we were looking for.

Issue #807, opened 2024-01-22
Accessibility items that can be handled on UI level


hanbyul-here:
Follow up of : https://github.com/NASA-IMPACT/veda-ui/issues/789
- [x] Create skip navigation link for keyboard browsing
- [x] In all pages, implement tabindex order coordinated with the structure of the webpage
- [ ] Add icon to external links
- [ ] Update header text link styles (font-weight to semibold for better contrast)
- [ ] Update footer text link styles (font-weight to semibold for better contrast)

aboydnw:
I think we've addressed these, but let me know if that's not true

PR #808, opened 2024-01-22
Add 'skip to main' to navigation

hanbyul-here:
Add keyboard focusable 'Skip to main content' Button.
part of https://github.com/NASA-IMPACT/veda-ui/issues/807
Since we don't have an anchor link, We have to rely on some workaround (Manually focusing page body).

You should be able to see it by trying to keyboard navigate the page, but this is the screenshot.

Screen Shot 2024-01-22 at 12 21 03 PM

### DIFF
--- a/app/scripts/components/common/layout-root.tsx
+++ b/app/scripts/components/common/layout-root.tsx
@@ -58,7 +58,7 @@ function LayoutRoot(props: { children?: ReactNode }) {
         thumbnail={thumbnail}
       />
       <NavWrapper />
-      <PageBody id={PAGE_BODY_ID}>
+      <PageBody id={PAGE_BODY_ID} tabIndex={-1}>
         <Outlet />
         {children}
       </PageBody>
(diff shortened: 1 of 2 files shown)


faustoperez:
@hanbyul-here looking good!

In the preview it only works on the homepage, is that the intended behaviour? Thanks!

hanbyul-here:
@faustoperez oh no it should work on every page. But, when the page changes, your focus stays on the element you interacted with. ex. I click on 'data catalog' from the home page, my next focusable element will be data analysis. If I navigate back (shift+tab), then I can get to the navigation. (This focusing behavior can be also confusing for screen reader users... but let's keep this pr in its scope for now.) Lmk if you could get it working in other pages! 

Screen Shot 2024-01-24 at 11 28 58 AM

faustoperez:
gotcha, thanks for the clarification @hanbyul-here ! 🙌

faustoperez (approved):
Looking good, thanks for taking care of this!

PR #809, opened 2024-01-22
Separate out precommit configuration

hanbyul-here:
Fixed the problem (ts check does not scope the type check to the staged files. ex. throwing module not found error if the module is imported in a staged file, but not added to the stage) of ts-check in precommit hook by following:  https://github.com/lint-staged/lint-staged/issues/825#issuecomment-620018284


stephenkilbourn:
great catch! 🧠

stephenkilbourn (approved):

Issue #799, opened 2024-01-17
[VEDA2 Refactor] Exploration - Registry options and build pipeline to creating new core logic library


sandrahoang686:
**Description:**
As part of the refactor work for veda-ui v2, we need to figure out how to access components from the instance level to consume as a library. This will likely require us to wrap the current ui logic with a layer to easily access components by importing through npm registry (to be determined). Current veda instances will still use the veda-ui logic as is (git submodule) but we also need to support consuming this veda-ui v2 as a library.

**Acceptance Criteria:**

- [x] Research best registry option (Github, npm, etc). There might be restrictions from our stakeholders so we should determine what those are to make this choice. Likely npm
- [ ] Determine how much overhead is needed to create access layer around veda-ui to consume react components at the instance level through chosen registry (npm for example)
- [ ] Support current and new way of consumption
   - [ ] Current `veda-ui` structure is intact and does not introduce any breaking changes to current build pipeline and current instances can still use `veda-ui` exactly the same
   - [ ] Can also consume `veda-ui` as a library for the new test instance with same build pipeline
- [ ] Document findings in this ticket so work can be started on #800 

**Resources:**
https://miro.com/app/board/uXjVN6lkBnc=/

sandrahoang686:
Another thought, i guess depending on whether or not we use npm for v2.. to develop this locally with the new test instance, we will probably have to have both running at the same time and may need to do an npm link of veda-ui in the new instance code

hanbyul-here:
We will find out more in the research phase, but I believe where to publish/host our packages should not block us from using npm-related tools (such as npm link). But it is a good point that it might be an extra burden to run both veda-ui and test instance at the same time. We can consider something like sandbox https://visex.netlify.app/sandbox that can be run inside of veda-ui too. (I will detail it in #803 )

hanbyul-here:
## Research for the best registry option

### NPM registry
 https://docs.npmjs.com/cli/v9/using-npm/registry
An NPM registry is just a collection of Node.js packages, metadata, and several API endpoints that are primarily accessed via the npm (Node Package Manager) command line tool. (ex. npm install) - So basically, where the package is hosted. To resolve packages by name and version, npm talks to a registry website that implements the CommonJS Package Registry specification for reading package info.

#### NPM & Yarn
[...]

PR #810, opened 2024-01-22
Get rid of prepare script

hanbyul-here:
I forgot to include this change in the follow pr -_-; I don't see why we need prepare script for husky, unless if you had any specific intention with it? @stephenkilbourn

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,7 @@
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
     "ts-check": "yarn tsc --noEmit --skipLibCheck",
-    "test": "jest",
-    "prepare": "husky install"
+    "test": "jest"
   },
   "engines": {
     "node": "16.x"


stephenkilbourn:
@hanbyul-here - I don't see anything specific from Husky about removing their prepare hook.  It is generated automatically as part of the install process.  My understanding is that it makes sure husky is installed and running even if the developer has not pulled down the latest `package.json` and re-run `yarn install`.  In other words, I think it is a safety to make sure the husky hook gets run for all users even if they don't always re-run the dependency install for their PR.

hanbyul-here:
@stephenkilbourn 
This is related to our future plan for making veda-ui as a library. (Not happening very soon, but you can see the ticket: https://github.com/NASA-IMPACT/veda-ui/issues/799) As far as I can check from [NPM docs](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts), prepare script runs 

- Runs BEFORE the package is packed, i.e. during npm publish and npm pack : which is fine, since the package publisher will have husky installed anyway
- Runs AFTER prepublish, but BEFORE prepublishOnly: also fine.
- Runs on local npm install without any arguments: This is the case I am concerned. People who install veda-ui does not need husky installed together. 

Let me know if there is something missing, or other ways to work around!

stephenkilbourn (approved):

PR #815, opened 2024-01-26
Increase Exploration Run Analysis Performance

sandrahoang686:


hanbyul-here (commented):
This is great! I am learning a lot about generators and how it gives a better control of async requests from this PR 🙇  I've noticed that even when I reduce the asset size to 1 for testing state synchronization, there's no 'bouncing-back' issue, which I believe is a testament to the improved control afforded by the generator's execution process.

As we previously discussed, we need to address the global limit on the number of outgoing requests - currently capped at 15 by our concurrency manager - especially considering scenarios with numerous layers added on the page. Do you have any suggestions on how we could tackle this challenge while leveraging the approach from this PR?

j08lue:
> we need to address the global limit on the number of outgoing requests

Hmm, I am wondering what solution we chose earlier for limiting the number of concurrent `/statistics` requests. I think we implemented that analysis processes with too many time steps / requests will throw an error in the UI, asking the user to reduce their selection.

- https://github.com/NASA-IMPACT/veda-ui/issues/660

There are two separate concerns:

1. Limit the total number of requests for a single analysis job, so costs do not explode - this can be a large number, let's say 10k or 2k (TBD)
2. Limit the number of concurrent requests, so we do not overload the backend / executor (AWS Lambda + database).

It would be great if these could be decoupled, i.e. that we can allow users to submit requests up to, say, 2000 time steps total, but batch the requests towards the backend such that no more than 100 or so are running simultaneously.

Where are we on this goal, with this PR?

hanbyul-here:
@j08lue Since this PR's scope is more to show the different approach of making requests, I handled your concern in #817
To answer your questions, 
1. Limit the total number of requests for a single analysis job: This means the number of assets, I believe? which we set it as 300. (We still have the same limit for a single analysis job.) - and yes, UI prevents the user from proceeding if the expected # of requests is > 300.
2. Concurrency - we set the number of concurrency as 15, the new PR still sets it as <= 15.

j08lue:
Thanks for clarifying!

PR #816, opened 2024-01-30
Convert strings to js string

hanbyul-here:
I blanked that React cannot render HTML string unless it is specifically treated HTML.


j08lue (approved):
my bad - I had just copy-pasted this. thanks for catching.

PR #817, opened 2024-01-30
Change the approach of making requests for statistics

### LINKED ISSUE #790: [E & A] Statistics results are loaded but counter does not complete
We found a bug that analysis loading gets stuck at the last item or sometimes last few items. (It succeeds if the same request is made again with a different TOI or you repeat the same analysis). 

## Steps to reproduce (maybe):

1. Open exploration (e.g. this [preview](https://deploy-preview-217--ghg-demo.netlify.app/exploration))
2. Add a dataset
3. Draw or upload an area of interest
4. Refine the time range, perhaps a year or so
5. Hit "Run Analysis"
6. Result (sometimes): Out of, say, 20 time steps to fetch statistics for, the first 16 or so complete pretty fast (count of loaded items increases fast), but then the number of loaded items does not increase anymore, waiting forever.
   
   Image

8. You can see that the requests actually (did) work: click on the map outside of the area of interest you drew, then click on it again and hit "Apply Changes": The time series graph shows immediately

## Acceptance criteria

- [ ] Stats results loading does not hang anymore

hanbyul-here:
Close #790 

Upon reviewing the issue with Jotai Devtool, I found that the problem wasn't limited to just the last few requests. During the request process, there was instability in how metadata was set. This led to erratic behavior in the tracking of loaded assets. For instance, if there were 5 assets to load, the counter for the number of loaded assets would increase sequentially from 0 to 2, but then unexpectedly revert to 1, before progressing to 3, then 4, and finally 5, indicating a successful call. However, this would be followed by a regression back to 4, erroneously signaling a return to a loading state.

Even though I couldn't pinpoint why this happens (I have a hunch that it could be because of setState race condition) , I took @sandrahoang686 's approach from https://github.com/NASA-IMPACT/veda-ui/pull/815, which involves using generators to create batches of requests. This approach seems to give better control of the sequence, I cannot observe the erratic behavior anymore with the approach. In this PR, these generators were transformed into promises, allowing them to be managed by a concurrency manager effectively. (Since this method is relatively new to me, I'm not entirely sure if there are specific aspects I should be wary of. @sandrahoang686  if you have any insights or notice any red flags with this strategy, your feedback would be greatly appreciated. )

I've followed the pattern of the handling of requests into batches using a generator-based queue system in Sandra's PR. Then each promise in this setup corresponds to a generator that processes a batch of requests upon resolution. (then passed to concurrency manager) I've configured these batches to handle up to five requests simultaneously—a limit I've set via the batchLimit parameter. As a result of this batching strategy, I've significantly reduced the capacity of the concurrency manager. Previously, it could manage 15 jobs simultaneously, but now it's down to managing only 3. Gotcha is that the other requests than statistics requests are also pooled into this concurrency manager ex. The STAC endpoint call to get the metadata of the assets.) 

However, an important consideration is the inherent limitation of generators that only allow ${batchSize} requests to be sent out concurrently and the next batch has to wait. This limitation might make the process feel slower, or indeed be slower, than the current setup - which allows 15 requests simultaneously. I'm contemplating whether it would be beneficial to adjust the batch size dynamically based on the number of layers involved or other factors, to optimize performance. Or.. can we just allow one job at a time and make a batch size 15 ? 🤔 

I think I need to take more time to think about if there is any error state that I am missing, but opening the PR for discussion (and testing)


hanbyul-here:
Ah, after several tests, I realized that the state update inside of the generator function works as expected, but still, the state update outside of the generators' controls (success, error) is not stable. I put back a workaround, which gets the result of the requests, saves it as a local state, and then uses the value to update the Jotai value.

j08lue (commented):
## Fixes the "hanging" issue

In my tests, all requests always complete now.

On the [previous version](https://deploy-preview-217--ghg-demo.netlify.app/exploration), I can reproduce the issue with seemingly hanging progress very frequently (second try or so). 

On this version here, the "hanging" does not occur even within 15 trials (and counting).

I also tried to block a single request and saw that that gets caught properly:

image


## Degrades progress indication UX

You mentioned this already, @hanbyul-here:

> User-facing: the counter number will go up only by batch size.

In effect, this means that the counter sits at 0 for quite a while and then jumps, often to something like 50% or even 100% at once.

If this batch completion is how we best can implement this feature, then I think we should change the loading indicator somehow, so the user can sense that something is / will be happening even before the first batch completed.

https://uxdesign.cc/loading-progress-indicators-ui-components-series-f4b1fc35339a

In principle, we do not need to be that explicit about what exactly is happening and how many items need to be processed, I think.

image

A fancy progress indication might be that data points appear on the chart as they are available, but somehow faint, until all are complete and the lines / target styles are applied. Just a quick idea - not necessarily a good one, I'll defer to designers.

## Request optimization

Let us sync up with the backend team to align how these requests are best submitted / answered. Limiting the overall number of concurrent requests sounds like a good measure, but let's find out.

j08lue (approved):
I'll actually approve this change. Based on my tests, it fixes the "hanging" bug and for that, we can accept the slight UX degradation for now.

PR #818, opened 2024-01-31
Add no layer status

### LINKED ISSUE #791: [E & A] Design a way out from E&A layer selector
In the new Exploration & Analysis interface, users cannot exit (e.g. access a link to a different page) as long as the dataset selector modal is up, since it is the only active item on the page.

This is not nice - if a user enters Exploration but decides they do not actually want to proceed, they have to take a step into the direction they did not want to go (actually add a layer to the map etc), before being able to go back.

We can solve this in different ways:
1. Adding a nav to the layer selection modal that offers a way back to home or something
2. Add an empty state for the Exploration & Analysis interface, such that users can click away the layer selection modal without having to select a layer. For example an empty map with a big call to action to add a layer.
3. Decide that there is always one layer on the map by default

## Acceptance criteria

- [ ] Decided what to do
- [ ] Shared design with team

hanbyul-here:
Close #791 

@faustoperez : I took a liberty and removed all the controls - including zoom control - for no layer status. I couldn't think of any way to make a zoom control meaningful when there is no dataset. lmk if you have a different idea!

### DIFF
--- a/app/scripts/components/exploration/components/timeline/timeline-controls.tsx
+++ b/app/scripts/components/exploration/components/timeline/timeline-controls.tsx
@@ -104,9 +104,7 @@ export function TimelineControls(props: TimelineControlsProps) {
   const { features } = useAois();
 
   // Scale to use when there are no datasets with data (loading or error)
-  const initialScale = useMemo(() => {
-    return getInitialScale(width);
-  },[width]);
+  const initialScale = useMemo(() => getInitialScale(width) ,[width]);
 
   return (
     <TimelineControlsSelf>
--- a/app/scripts/components/exploration/components/timeline/timeline-controls.tsx
+++ b/app/scripts/components/exploration/components/timeline/timeline-controls.tsx
@@ -68,18 +68,20 @@ interface TimelineControlsProps {
 }
 
 
-export function useInitialScale(width) {
-  return useMemo(() => {
-    const now = new Date();
-    return scaleTime()
-      .domain([startOfYear(now), endOfYear(now)])
-      .range([0, width]);
-  }, [width]);
+export function getInitialScale(width) {
+  const now = new Date();
+  return scaleTime()
+    .domain([startOfYear(now), endOfYear(now)])
+    .range([0, width]);
 }
 
 export function TimelineDateAxis(props: Omit<TimelineControlsProps, "onZoom">) {
   const { xScaled, width } = props;
-  const initialScale = useInitialScale(width);
+
+  const initialScale = useMemo(() => {
+    return getInitialScale(width);
+  }, [width]);
+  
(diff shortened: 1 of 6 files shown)


app/scripts/components/exploration/components/timeline/timeline.tsx
```diff
@@ -499,15 +507,51 @@ export default function Timeline(props: TimelineProps) {
...
+      </TimelineHeader>
+        <TimelineContent>
+        <TimelineDetails />
+        <EmptyTimelineContentInner>
+          <DateGrid width={width} xScaled={mockedScale} />
+          <LayerActionBox>
+            <div>
+              <CollecticonIsoStack size='xxlarge' />
+              <p>No data layer added to the map!</p> 
+              <Button variation='base-text' size='small' onClick={onDatasetAddClick}>Add a Layer here</Button>
```
j08lue, line 550:
```suggestion
              <Button variation='base-text' size='small' onClick={onDatasetAddClick}>Add a layer here</Button>
```

j08lue (approved):
This works fine, I think.

A little nitpick: lowercase "layer".

I also tried different scenarios - what happens when you are in an active analysis and then remove and add layers. Behavior seems to be as I would expect - the analysis state and user geometry stays on and can be reused after new layers have been added. 👌

faustoperez:
Agree with Jonas, this is fine @hanbyul-here 👍

app/scripts/components/exploration/components/timeline/timeline-controls.tsx
```diff
@@ -68,18 +68,20 @@ interface TimelineControlsProps {
...
+    .range([0, width]);
 }
 
 export function TimelineDateAxis(props: Omit<TimelineControlsProps, "onZoom">) {
   const { xScaled, width } = props;
-  const initialScale = useInitialScale(width);
+
+  const initialScale = useMemo(() => {
+    return getInitialScale(width);
+  }, [width]);
```
sandrahoang686, line 83: ⛏️ : I would simplify like this 

`const initialScale = useMemo(() => getInitialScale(width), [width]);`

not a blocker though

app/scripts/components/exploration/components/timeline/timeline.tsx
```diff
@@ -522,9 +522,11 @@ export default function Timeline(props: TimelineProps) {
         </Button>
       </Headline>);
   };
+  // Stub scale for when there is no layers
+  const initialScale = useMemo(() => {
+    return getInitialScale(width);
+  }, [width]);
```
sandrahoang686, line 528: Same as above comment but not a blocker

sandrahoang686 (approved):

PR #824, opened 2024-02-07
Overwrite analytics request result

hanbyul-here:
It is this PR: https://github.com/NASA-IMPACT/veda-ui/pull/817 without changing how the requests are processed. 

This doesn't change public UX at all, nor changes the max number of requests analysis page makes. - @j08lue I think you are right that some sorts of UI change should accompany the change of how we batch process the requests in #817 . 

This approach has the problem of : The number of 'loaded items' is not sequentially increasing (which #817 resolved), having a local copy of the state just to overwrite the global state (which #817 couldn't resolve.)

I think we should move towards more of the approach from #817 very soon, I will open a tech debt ticket about it once we decide to take this pr's route.)


j08lue (approved):
This seems to resolve the issue with the "hanging" requests, while the current UI progress indicator (counter) still makes sense, in contrast to https://github.com/NASA-IMPACT/veda-ui/pull/817#pullrequestreview-1867252860.

I am unable to notice non-sequential counter increase - in the tests I did, once the responses come in, they go so fast that I could not tell. So that is not an issue, I'd say. 👌 

Fine to follow up with changing the request approach, perhaps after sync with the Backend team about how to optimize for performance / efficiency.

app/scripts/components/exploration/hooks/use-analysis-data-request.ts
```diff
@@ -107,23 +114,39 @@ export function useAnalysisDataRequest({
...
+        dataset,
+        queryClient,
+        concurrencyManager: analysisConcurrencyManager,
+        onProgress: (data) => {
+          setAnalysis(data);
+        }
+      });
+      setAnalysisResult(stat);
+    }
+    makeCall();
```
sandrahoang686, line 132: ⛏️  : In this case, we can probably have it self invoke so we dont need to do `makeCall()` here
```
(async  () => {
    ....
})();
```

app/scripts/components/exploration/analysis-data.ts
```diff
@@ -246,20 +246,37 @@ export async function requestDatasetTimeseriesData({
...
     concurrencyManager.dequeue(`${id}-analysis-asset`);
-
-    onProgress({
+    return {
       ...datasetAnalysis,
       status: TimelineDatasetStatus.ERROR,
       error,
       data: null
-    });
+    };
```
sandrahoang686, line 280: 💭 : hmm.. I see that in some places we are returning type `TimelineDatasetAnalysis` for it to be later set to the analysis and then in others, we are invoking the callback. i'm wondering a few things to clean this up.. 

1. can we just pass in `setAnalysis` directly, no need for `onProgress`
2. define a var of type `TimelineDatasetAnalysis` and have it overwrite and then just return that var or pass it to `setAnalysis`. this is really a nit in term of it looking a little more cleaner

What do you think?

app/scripts/components/exploration/hooks/use-analysis-data-request.ts
```diff
@@ -107,23 +114,39 @@ export function useAnalysisDataRequest({
...
+  useEffect(() => {
+    // setAnalysis sets Jotai Atom value for analysis result
+    // It should sequentially set the analysis progress as 
+    // idle => loading (increasing number of loaded items) => success
+    // However, it is not setting the value sequentially, 
+    // ended up loading status as a result even when the request is successful
+    // This just overwrites the analysis result with the final result of analysis, 
+    // so Jotai value gets 'corrected' in case it is wrong
+    setAnalysis(analysisResult);
+  },[setAnalysis, analysisResult]);
```
sandrahoang686, line 150: 💭 : now i'm not sure but instead of this useEffect which watches `analysisResult` as a dependency. Could we just call `setAnalysis` on line 130 instead of doing `setAnalysisResult(stat);` and then in that case we wouldn't need this local state?

hanbyul-here: Yeah, that was my thought too first, too, but that approach doesn't set the state stably. (Observing the same pattern of state bouncing back and forth) 

~~ After a while ~~

OH I THINK I found a more concrete clue, I will post the finding in the main thread.

hanbyul-here:
I opened this PR without knowing the cause of the problem, mainly to brute-fix the problem so we can move on. 

I couldn't shake off the folly of wanting to find the cause...  After trying several different things (ex. Using event emitter for state update, using generator to create a batch of requests...) I think we could scope down the problem a bit better now. 

In short, the problem goes away when we 'disable' automatic batch updates that come out of the box from React. (`flushSync`) Then it means that the timeline literally gets re-rendered every time the counter goes up, and I can see that the timeline 'freezes' when there are ~300 requests expected. 🤔  Then Question is:  Is this an acceptable amount of regression of performance? Or better to just brute force the state update with a local copy? 



+ I thought automatic batching will guarantee the order of the async executions - so I am unsure where the root of the problem is - a simiar issue I could find is: https://github.com/facebook/react/issues/25734

app/scripts/components/exploration/hooks/use-analysis-data-request.ts
```diff
@@ -105,25 +106,44 @@ export function useAnalysisDataRequest({
...
+  //   // TECH DEBT
+  //   // setAnalysis sets Jotai Atom value for analysis result
+  //   // It should sequentially set the analysis progress as 
+  //   // idle => loading (increasing number of loaded items) => success
+  //   // However, it is not setting the value sequentially, 
+  //   // ended up loading status as a result even when the request is successful
+  //   // This just overwrites the analysis result with the final result of analysis, 
+  //   // so Jotai value gets 'corrected' in case it is wrong
+  //   setAnalysis(analysisResult);
+  // },[setAnalysis, analysisResult]);
```
sandrahoang686, line 147: I would leave an `@NOTE` OR `@TECH-DEBT` comment to explain why we have to brute force overwrite

sandrahoang686 (approved):

PR #825, opened 2024-02-07
Add 'reopen guide' button

### LINKED ISSUE #792: [E & A] Guided tour shows up on again on re-visit
If I am not mistaken, the guided tour to the E&A interface shows up every time the user opens the page, also when returning.

Image

## Acceptance criteria

- [ ] Made sure guided tour shows up only on first visit

### LINKED ISSUE #793: [E & A] Give users an option to invoke the guided tour again
During testing, users found the guided tour relevant and helpful.

Some wished to invoke it again later, because they wanted a refresher on the instructions or regretted clicking it away too fast.

## Acceptance criteria

- [x] Designed a way for users to bring up the guided tour again after closing it
- [ ] Implemented it

hanbyul-here:
Close #793 
Close #792 

+ Turning the a&e page on and off only through the environment file is cumbersome during active development. Test by directly typing the URL /exploration.  ex. https://deploy-preview-825--veda-ui.netlify.app/exploration

### DIFF
--- a/app/scripts/components/common/map/style/button.tsx
+++ b/app/scripts/components/common/map/style/button.tsx
@@ -4,15 +4,9 @@ import { themeVal } from '@devseed-ui/theme-provider';
 
 import { TipButton } from '$components/common/tip-button';
 
-// Why you ask? Very well:
-// Mapbox's css has an instruction that sets the hover color for buttons to
-// near black. The only way to override it is to increase the specificity and
-// we need the button functions to get the correct color.
-// The infamous instruction:
-// .mapboxgl-ctrl button:not(:disabled):hover {
-//   background-color: rgba(0, 0, 0, 0.05);
-// }
-
+// &&&: To increase the specificity of css class
+// https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity
+// so we can override the derived selectors from Mapbox .mapboxgl-ctrl button
 export const SelectorButton = styled(TipButton)`
   &&& {
     ${createButtonStyles({ variation: 'surface-fill', fitting: 'skinny' } as any)}
--- a/app/scripts/components/common/map/style/button.tsx
+++ b/app/scripts/components/common/map/style/button.tsx
@@ -1,0 +1,27 @@
+import styled from 'styled-components';
+import { createButtonStyles } from '@devseed-ui/button';
+import { themeVal } from '@devseed-ui/theme-provider';
+
+import { TipButton } from '$components/common/tip-button';
+
+// Why you ask? Very well:
+// Mapbox's css has an instruction that sets the hover color for buttons to
+// near black. The only way to override it is to increase the specificity and
+// we need the button functions to get the correct color.
+// The infamous instruction:
+// .mapboxgl-ctrl button:not(:disabled):hover {
+//   background-color: rgba(0, 0, 0, 0.05);
+// }
+
+export const SelectorButton = styled(TipButton)`
+  &&& {
+    ${createButtonStyles({ variation: 'surface-fill', fitting: 'skinny' } as any)}
+    background-color: ${themeVal('color.surface')};
+    &:hover {
+      background-color: ${themeVal('color.surface')};
+    }
+    & path {
+      fill: ${themeVal('color.base')};
+    }
+  }
+`;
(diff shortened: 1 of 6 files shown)


j08lue (approved):
I can confirm that:
1. Guided tour shows up only on first visit
2. Icon to invoke it looks like expected and works
3. Icon is inactive when the map state is empty
💯 

Icing on the cake would be if the tour, when closed, had some kind of animation towards that button, basically giving the user a visual hint that they can find it behind that button if they need it again. But if that is not quick to add, we can keep that for maybe later.

app/scripts/components/common/map/style/button.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { TipButton } from '$components/common/tip-button';
+
+// Why you ask? Very well:
+// Mapbox's css has an instruction that sets the hover color for buttons to
+// near black. The only way to override it is to increase the specificity and
+// we need the button functions to get the correct color.
+// The infamous instruction:
+// .mapboxgl-ctrl button:not(:disabled):hover {
+//   background-color: rgba(0, 0, 0, 0.05);
+// }
```
sandrahoang686, line 14: ❓ 💭 : Do we want this note here? If this is helpful, might be worth including the link to this documentation. Then this is also a random thought which we can discuss outside this PR... but I was thinking we should probably start marking our in-code documentation with maybe helpful pre-fixes (ex: `NOTE, COMMENT, TODO, TECH-DEBT`) this would help us identify helpful comments in the code, todo's to come back too and then tech-debt meant to be thought about and revisited? Again this is a discussion for a different time but wanted to say it here to start the convo, lmk if you have thoughts

app/scripts/components/common/map/style/button.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export const SelectorButton = styled(TipButton)`
+  &&& {
```
sandrahoang686, line 17: ❓ : Curious - what is this `&&&` ?

app/scripts/components/exploration/tour-manager.tsx
```diff
@@ -200,3 +213,37 @@ export function PopoverTourComponent(props: ExtendedPopoverContentProps) {
...
+    setCurrentStep(0);
+    setSteps?.(introTourSteps);
+    setIsOpen(true);
+  },[setIsOpen]);
+
+  useThemedControl(() => <TourButtonComponent {...props} onClick={reopenTour} />, {
+    position: 'top-right'
+  });
+  return null;
+}
```
sandrahoang686, line 249: 🔧 : can we get rid of passing in the rest of the props? seems like we just need to pass in `onClick`. Can we also move the jotai hooks from `TourButtonComponent` into `TourManagerInvokingButton` so then `TourButtonComponent` would just take in another prop for the `count`? that way we can keep `TourButtonComponent` pure. This can start helping us move hooks and data management pieces into smart components and then allowing dumb/pure components to note care about those pieces. Then we dont do this right now but if we wanted to unit test components in the future, this approach would allow that much easier.

app/scripts/components/common/map/style/button.tsx
hanbyul-here: I think I had the exact same question when I first saw it! It is to bump up the specificity of the selector so we can override: https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity (Maybe this should go into the comment ^-^;)

hanbyul-here: I changed the comment. I hope that the new comment actually answers the questions the code-reviewers would have. I found this comment helpful Especially because Mapbox doesn't offer an official doc about how to override their styles 
- Does the new comment help better?
- Is there any better way to clearly indicate this maneuver for styles?

app/scripts/components/common/map/style/button.tsx
hanbyul-here: Yeah, having a prefix is a great idea. Is there any way to automatically check these prefixes for comments? I do not trust myself to remember it!

app/scripts/components/exploration/tour-manager.tsx
hanbyul-here: (Sorry, I decided to move the file to where other controls are; it is map/tour-control.tsx` now.)
Yeah, this is a really thoughtful comment aligning with what we are trying to do with VEDA. 🙇

app/scripts/components/common/map/style/button.tsx
sandrahoang686: I think we will just have to create that standard ourselves and document it somewhere for development purposes

sandrahoang686 (approved):
lgtm

PR #826, opened 2024-02-07
Manually center geocoding result

### LINKED ISSUE #813: Location from search box on Exploration is off
Searching for a location in Exploration may send users way off the right place. Could be a question of the map projection.

## Acceptance criteria

- [ ] Location search sends users the right place

hanbyul-here:
Close #813 

It seems to be a known bug : https://github.com/mapbox/mapbox-gl-js/issues/12565

I put a manual workaround to center the map with the result from the geocoder for now. But hopefully, we can get rid of this workaround soon with Mapbox resolving the issue. 

(Sorry for confusing branch name, I thought bumping Mapbox would resolve the issue at first, but it seems not all the fixes have made it to production yet.)

### DIFF
--- a/app/scripts/components/common/map/controls/geocoder.tsx
+++ b/app/scripts/components/common/map/controls/geocoder.tsx
@@ -3,8 +3,7 @@ import MapboxGeocoder from '@mapbox/mapbox-gl-geocoder';
 
 import { useControl } from 'react-map-gl';
 
-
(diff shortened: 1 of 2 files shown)


j08lue (approved):
That seems to fix the issue. I tested location search with various non-Mercator projections and if faithfully took me the right places. 🚀

sandrahoang686 (approved):
lgtm

PR #827, opened 2024-02-07
Start replacing with new map in story blocks

sandrahoang686:


app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -166,26 +251,132 @@ function MapBlock(props: MapBlockProps) {
...
+          dataset &&
+          selectedDatetime &&
+          layerId &&
+          baseDataLayer?.data &&
+          (
+            <Layer
+              key={baseDataLayer.data.id}
+              id={`base-${baseDataLayer.data.id}`}
+              dataset={(baseDataLayer as unknown) as TimelineDatasetSuccess}
+              selectedDay={selectedDatetime}
```
sandrahoang686, line 305: 📓 : Casting because type is expecting `TimelineDatasetSuccess` but variable is only partial. We should probably update the type that Layer expect here to accept partial

app/scripts/components/exploration/components/map/layer.tsx
```diff
@@ -18,23 +18,35 @@ import { CMRTimeseries } from '$components/common/map/style-generators/cmr-times
...
+    // 'settings' on undefined value even when dataset has 'settings' key. This is a workaround for now but
+    // should be revisited. Ideally type should be fine with 'Partial<TimelineDataset>'
+    const [getSettings] = useTimelineDatasetSettings(datasetAtom);
+  
+    isVisible = getSettings('isVisible');
+    opacity = getSettings('opacity');
+  } catch {
+    isVisible = true;
+    opacity = undefined;
+  }
```
sandrahoang686, line 44: 📓 : wrapping this jotai logic with a try/catch because the dataset being passed in kept erroring with an error like "unable to find `settings` on `undefined`..." even after transforming the data to include the `settings` attribute key. If we look at the types though... `settings` is not optional, yet all the keys in settings are optional... we probably need to clean up typing here in the future. But for now this is my workaround. Please let me know if there are any better suggestions

j08lue (changes requested):
Looks great on the very nice video walk-through you made, @sandrahoang686.

However, I am unable to review the functionality on the preview. Only the mock Covid19 story has comparison maps and they do not seem to be set up for comparison, just for showing a single layer?

https://deploy-preview-827--veda-ui.netlify.app/stories/air-quality-and-covid-19

What I am able to see on the build preview, though, is that the old attribution icon is still on the map, now overlaying the zoom controls:

image

So this may need a bit of adjustment - I think these:
- [x] Change mock data so we can do a full feature evaluation
- [ ] Remove old attribution icon

sandrahoang686:
@j08lue thanks for taking a look! I made some updates after your comment. I didn't come across the case of if there was no compare date, thanks for pointing to a story that had this

Update 1: Now the compare control only renders when there are two dates to compare, this should match old behavior now! (for example, see last block map in `/stories/air-quality-and-covid-19`)
Update 2: I kept the old attribution icon because I realized this was tied to the content that users contribute too in the MDX files where users added `attrAuthor, attrUrl`. So I just moved the zoom controls over to the "top-left" for these block maps. Exploration page widget locations remain the same though.

This is what it should look like now in the preview...
Screenshot 2024-03-12 at 2 43 16 PM

j08lue (approved):
Looks good now in terms of look and feel and functionality. ✅ 

One issue I've always had with this feature is that text elements on the map pick up when you move the slider past them and the text gets highlighted. That is just annoying. But it has always been like that and fixing it would probably be beyond the scope of this feature. We could also give the feature a face lift in general. I might ticket that later - we have more important things to do right now, though.

image

hanbyul-here:
Thank you so much for your efforts on addressing this issue! 🙇 It is great to see such a progress.

I have a query regarding the data reconciliation process. The new exploration map has a new way of reconciling layer data, with the` Timelinedatasets` atom being updated through `useStacMetadataOnDatasets`. When we use `useAsyncLayers` for blockmap —which saves the data with React context— , it seems we end up with two distinct sets of reconciled data across the application (one in atoms and the other in context). Could we explore the possibility of unifying the methods for reconciling layer data?

Additionally, I wanted to touch on some other points:

- Attribution Control should be used for map-related attribution: https://docs.mapbox.com/help/dive-deeper/attribution/#text-attribution (This is an important thing to do for any osm data consumer - we are using osm data through Mapbox.) We can still use the AttributionControl for map-related attribution, but we should use a separate one for figure attribution.
- What are your thoughts on migrating the necessary components from the `/mapbox` directory to `/map`?This could streamline our process by allowing us to simply remove the /map folder when necessary, instead of navigating through the codebase. (I think duplication is fine if necessary) 

I look forward to your insights on these matters!

sandrahoang686:
@hanbyul-here great points! 
* `useDatasetAsyncLayer` logic was copied over from the old way in the `/common/mapbox` dir. but I can change that so it uses the new way to keep it standardized 👍🏼 
* I will actually remove AttributionControl component. Originally I introduced it into the code base not knowing that the MDX files provided the `Caption` component here. AttributionControl will then take on [default behavior](https://visgl.github.io/react-map-gl/docs/api-reference/attribution-control) like before
* We will definitely need to migrate those necessary components over to the `/map` dir like the `LayerLegend` component. But let me create a ticket for this and keep it separate from this PR. So I can think a little more deeply about how to arrange the files and content especially since we need to start breaking components out into features and re-useable pieces more. 
Ticket for that here: https://github.com/orgs/NASA-IMPACT/projects/17/views/6?pane=issue&itemId=56505746

parcel-resolver-veda/index.d.ts
```diff
@@ -24,20 +24,21 @@ declare module 'veda' {
...
-  interface DatasetLayerCommonProps {
+  interface DatasetLayerCommonCompareProps {
+    mapLabel?: string | DatasetDatumFn<DatasetDatumReturnType>;
+  }
+
+  interface DatasetLayerCommonProps extends DatasetLayerCommonCompareProps{
     zoomExtent?: number[];
     bounds?: number[];
     sourceParams?: Record<string, any>;
   }
```
sandrahoang686, line 35: 📓 : I've inversed this as this is what makes most sense right now with how our interfaces are currently typed. `DatasetLayer` extends `DatasetLayerCommonProps` and `DatasetLayer` is used widely in the TimelineDataset types which we use heavily in the map component logic and hooks. For now, this works and it is fine to do this since `DatasetLayerCommonCompareProps` has one optional key anyways. 

Made a tech-debt ticket here though as we need to revisit many of our types.. there is too much casting going on generally in the codebase and the types could be cleaned up and be more explicit...
https://github.com/NASA-IMPACT/veda-ui/issues/905

sandrahoang686:
@hanbyul-here this ticket is ready for re-review 👀 . I've updated and cleaned up your feature branch PR a good amount and merged it in here. I've also created a new tech-debt ticket for us to revisit typing which is out of scope for this PR. Included that ticket in one of my comments.

hanbyul-here:
I am seeing some unexpected behavior on maps with compare mode on.

1. https://deploy-preview-827--veda-ui.netlify.app/sandbox/mdx-page - I cannot get the map rendered at all with commpare mode on. I only seen an error `Cannot read properties of undefined (reading 'version')` where a map is supposed to be.
2. When the compare mode was interacted enough, an error that `Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.` pops up. Can you take a look at this? It might take a while to trigger this error. I encountered this error on https://deploy-preview-827--veda-ui.netlify.app/stories/air-quality-and-covid-19  Attaching a screen recording in case this helps.

https://github.com/NASA-IMPACT/veda-ui/assets/4583806/8f206f1f-fa4d-4d43-9883-cea1df5b33a3

app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -1,29 +1,42 @@
...
+} from '$components/common/map/controls';
+import { Layer } from '$components/exploration/components/map/layer';
+import {
+  S_SUCCEEDED
+} from '$utils/status';
+import { TimelineDataset, TimelineDatasetSuccess } from '$components/exploration/types.d.ts';
+
+import { reconcileDatasets } from '$components/exploration/data-utils';
+import { datasetLayers } from '$components/exploration/data-utils';
+import { useReconcileWithStacMetadata } from '$components/exploration/hooks/use-stac-metadata-datasets';
```
dzole0311, line 33: I wonder if importing the types, components and hooks from the Exploration will always work, especially if at some point these features diverge (`block-map` VS the usage of the map hooks in the `Exploration` feature which to me it looks like something different). We have (I think?) two remaining places to replace the old with the new map component and we might need some of the types, or reconciliation methods in there (this is just an assumption at this point). If that's the case, maybe it's better to have some common base atoms and hooks that update them, and then each feature could derive sub-atoms based on them if needed?

dzole0311 (commented):
@sandrahoang686 Aside from some small comments, the rest looks good to me. I also learnt a few things on how we handle data and reconciliation through this PR. 

PS. the loom walkthroughs I see in your PRs are really helpful! 🤩

app/scripts/components/exploration/components/map/layer.tsx
```diff
@@ -18,23 +18,35 @@ import { CMRTimeseries } from '$components/common/map/style-generators/cmr-times
...
+    // @TECH-DEBT: Wrapping this logic with a try/catch because jotai errors because it is unable to find
+    // 'settings' on undefined value even when dataset has 'settings' key. This is a workaround for now but
+    // should be revisited. Ideally type should be fine with 'Partial<TimelineDataset>'
+    const [getSettings] = useTimelineDatasetSettings(datasetAtom);
+  
+    isVisible = getSettings('isVisible');
+    opacity = getSettings('opacity');
+  } catch {
+    isVisible = true;
+    opacity = undefined;
```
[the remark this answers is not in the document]
sandrahoang686: I cant exactly remember why I've set it to undefined but i'm thinking because opacity wasn't being defined as part of the dataset from the story block maps originally

app/scripts/components/common/mapbox/map-message.tsx
```diff
@@ -5,6 +5,8 @@ import { Transition, TransitionGroup } from 'react-transition-group';
 import { glsp, themeVal } from '@devseed-ui/theme-provider';
 import { variableGlsp } from '$styles/variable-utils';
 
+// @NOTE: File should be moved under "/common/map" as we are working to deprecate "/common/mapbox" dir
+
 const fadeDuration = 240;
```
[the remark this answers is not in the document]
sandrahoang686: Nope, we do not. I have seen and done a fair amount of camel casing myself but we should better define these types of code standards and document them somewhere... i'll can ahead and plan for something..

app/scripts/components/common/blocks/block-map.tsx
sandrahoang686: a common directory would be great! especially since there are actually a few components using `/exploration` logic & hooks. We haven't gotten to that clean up yet but I think both @hanbyul-here and I are aware. I think we only care to replace the story block maps with the new component and the others i'm actually unsure about.

As part of the refactor we have a story to break out the Map core feature component ([here](https://github.com/NASA-IMPACT/veda-ui/issues/899)) so much of this cleanup can happen then but also I think it makes sense to chip away at this gradually due to scope of use.

What do you think though?

dzole0311:
> I am seeing some unexpected behavior on maps with compare mode on.
> 
> 1. https://deploy-preview-827--veda-ui.netlify.app/sandbox/mdx-page - I cannot get the map rendered at all with commpare mode on. I only seen an error `Cannot read properties of undefined (reading 'version')` where a map is supposed to be.

Still looking into it.

> 2. When the compare mode was interacted enough, an error that `Maximum update depth exceeded 

I managed to reliably reproduce this issue. It seems to occur when we zoom out fast (to zoom level 0) which triggers a lot of identical state updates using the setInitialViewState and this reaches the Maximum update depth exceeded error limit in React. Hopefully it's addressed now 🤞 

https://github.com/NASA-IMPACT/veda-ui/assets/13894251/3714d653-05d0-40e4-8766-6317ecb7dcd9

hanbyul-here:
Ah, it is so great to know the root of the problem 🙇 
re: triggering identical states - If the bug was caused by changing viewState way too fast, shouldn't the callback be debounced?

dzole0311:
> If the bug was caused by changing viewState way too fast, shouldn't the callback be debounced?

Good point, added.

hanbyul-here (approved):
This seems to be the PR that all of us three worked on- and I think it is good to go out now! Thanks a lot for working on this @sandrahoang686 @dzole0311 🙇

Issue #821, opened 2024-02-05
[E & A] Implement dataset/collection information on layers


hanbyul-here:
Implementation ticket of 
- #774

## Refinement required

* Clarification on the timeline needed - ASAP 
* The details of the design can be changed. Check the Figma link to see the latest design.

![Image](https://github.com/NASA-IMPACT/veda-ui/assets/4583806/384acac9-6cf8-4337-bd38-ff3452080763)

Figma Link: https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=855-1241&mode=design&t=MWBrtK0xYXZHT06E-0
## Acceptance criteria
- [ ] The new layer should show the data set it belongs
- [ ] The new layer should match the provided design mockups.
- [ ] No interactivity change - The button interaction (that opens the modal with all the layers of the dataset) will be implemented along with #828 

## Nice to have
- [ ] New look for layer option tooltip

hanbyul-here:
Closed via https://github.com/NASA-IMPACT/veda-ui/commit/b64ab957d293487161467bd1d4eb04336d70e87d

PR #829, opened 2024-02-08
[E&A] Implement dataset/collection information on layers #821

sandrahoang686:
Implements the data layer cards in the E&A page

Ticket: https://github.com/orgs/NASA-IMPACT/projects/17?pane=issue&itemId=51911993

Figma: https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=893-982&mode=design&t=TTmoflPGMFYHuMok-0

**Completed:**
* New menu for each layer item
* Card style at rest
* Hover State Styling
* Active state styling

**Needs Implementation:** 
* Move up and down functionality (ask @faustoperez )
* Card info is still hardcoded

**Questions:**
* The cards are already draggable, do we need to the `Move up` and `Move down` options in the menu? (cc @faustoperez ) If so, I'm not too sure how to actually implement that with the current library we use [framer-motion](https://www.framer.com/motion/). Maybe @hanbyul-here would you know how to quickly implement this? The library supports dragging but not too sure about re-ordering onClick
* During active state, we want to rotate the cards -2 deg. But how dragging works currently is when we drag the card, the entire row item is moving as well (see pic or demo video). This is also tricky because when clicking on button inside the card (info button and dot menu), this will also rotate the card. Maybe this is where we would play with `z-index` which I did... but couldn't successfully figure out how to not rotate when clicking on button in the card... Is it okay if we get rid of the rotation styling on active state (cc @faustoperez)? 

**Demo Video:**
https://www.loom.com/share/2d84f78eab544597913e09c962fa2d43?sid=67cc1cbe-4822-437e-b8db-61866ac49efc


faustoperez:
I love the walkthrough videos @sandrahoang686 !

> Is it okay if we get rid of the rotation styling on active state (cc @faustoperez)?

We can skip the rotation on the card and use the default library behavior. It's a nice to have but it might take us longer to fine-tune 👍 

> The cards are already draggable, do we need the Move up and Move down options in the menu? (cc @faustoperez )

I brought these two actions for discoverability and accessibility, but maybe something we can also park for a future iteration 👍

hanbyul-here:
🥳 🥳 

Reorder of layers: I think we can directly edit the atom value without going through Framer to reorder layers. I made a pr : https://github.com/NASA-IMPACT/veda-ui/pull/847

Event propagation: As far as I know, there is no straightforward way to stop the event propagation from the child elements to parent elements only with CSS. But I agree that the card rotation doesn't seem to be very necessary, considering the timelines don't move together. 

My apologies if you are already working on it. Can we have a coherent name for the components? ex. every component starting with `data-list`...  or put them in one folder. It will help me to see the relationships between the components better!

hanbyul-here:
Ah, one tiny thing I found while playing around in case this helps. The card doesn't seem to stretch out when the title is short!
Screen Shot 2024-02-16 at 8 06 50 AM

hanbyul-here:
- [ ] remove the tilt
- [ ] The card being dragged needs to have a higher z-index
- [ ] comment out the hardcoded value
- [ ] General spacing (alignment of the elements)
- [ ]  (not caused by this PR) :  The handlebar to adjust the size of analysis panel can be styled better, and target across the panel. Timeline marker should also have 'A' (or 'A' marker on the left side of the calendar should not show up if comparison mode is not on.)

hanbyul-here (approved):
:shipit:

Issue #828, opened 2024-02-08
[E & A] Implement layer selection modal


hanbyul-here:
Design: https://github.com/orgs/nasa-IMPACT/projects/17/views/6?pane=issue&itemId=46784228

PR #834, opened 2024-02-09
Add Media as an optional attribute for a layer

hanbyul-here:
Part of #828 

This adds 'meda' as an optional attribute for a layer so that the media can be used for each layer. I am making a small PR so that it is clear how to add a media for each layer.

### DIFF
--- a/mock/datasets/no2.data.mdx
+++ b/mock/datasets/no2.data.mdx
@@ -50,6 +50,9 @@ layers:
   - id: no2-monthly
     stacCol: no2-monthly
     name: No2 PT
+    media:
+      src: ::file ./img-placeholder-3.jpg
+      alt: Placeholder Image
     type: raster
     projection:
       id: polarNorth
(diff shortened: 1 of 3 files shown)


mock/datasets/no2.data.mdx
```diff
@@ -50,6 +50,9 @@ layers:
   - id: no2-monthly
     stacCol: no2-monthly
     name: No2 PT
+    media:
+      src: ::file ./img-placeholder-3.jpg
+      alt: Placeholder Image
```
hanbyul-here, line 55: This is how a media can be added to a layer @faustoperez  I am just reusing an image that was already in the folder (`img-placeholder-3.jpg`) but let's go over how we can add an actual file later together 👯

sandrahoang686 (approved):

hanbyul-here:
@j08lue I think I make a better documentation when all the features are on main branch, I will make a ticket.

PR #837, opened 2024-02-13
Use allDatasets without 'disableExplore' flagging for data catalog page

hanbyul-here:
Fix the problem of the current GHG Instance that the datasets with the `disableExplore` flag don't show up on the catalog page - for the context, this flag was introduced because some datasets cannot use the default explorer. ex. https://earth.gov/ghgcenter/data-catalog/emit-ch4plume-v1 

For the data catalog page, show all the datasets whether 'disableExplore' is flagged. (The 'explore' button won't show up anyway)

### DIFF
--- a/app/scripts/components/data-catalog/index.tsx
+++ b/app/scripts/components/data-catalog/index.tsx
@@ -7,6 +7,7 @@ import { Subtitle } from '@devseed-ui/typography';
 import { Button } from '@devseed-ui/button';
 import { CollecticonXmarkSmall } from '@devseed-ui/collecticons';
 import { VerticalDivider } from '@devseed-ui/toolbar';
+import { datasets } from 'veda';
 
 import DatasetMenu from './dataset-menu';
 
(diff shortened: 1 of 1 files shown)


app/scripts/components/data-catalog/index.tsx
```diff
@@ -47,7 +48,8 @@ import {
   TAXONOMY_TOPICS
 } from '$utils/veda-data';
 import { DatasetClassification } from '$components/common/dataset-classification';
-import { allDatasets } from '$components/exploration/data-utils';
+
+const allDatasetsForCatalog = Object.values(datasets).map((d) => d!.data);
```
sandrahoang686, line 52: 💭  : It would be nice to have this at a global level in the same place & way we would access `datasets` globally as well in the future. Maybe expose this filtered arr from a state / method store too

sandrahoang686 (approved):

app/scripts/components/data-catalog/index.tsx
hanbyul-here: It turns out it already exists: https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/utils/veda-data.ts#L17 :[  It is my fault that I didn't look through more carefully, but also it is quite confusing that our data utils related to veda are spread over. I will try to consolidate after this sprint.

Issue #823, opened 2024-02-05
[E & A] Implement layer info modal


hanbyul-here:
Implementation ticket of 
- https://github.com/NASA-IMPACT/veda-ui/issues/811

## Refinement needed

- Clarification on the timeline needed
- The details of the design can be changed. Check the Figma link to see the latest design.

![Image](https://github.com/NASA-IMPACT/veda-ui/assets/4583806/079eb5ba-2452-4c42-b671-12e3f8ba312e)
![Image](https://github.com/NASA-IMPACT/veda-ui/assets/4583806/e1c6bf95-d9b2-42ea-a59a-b2b4006c626d)

## Acceptance Criteria

- [ ] Offer a way for editors to add a layer info through MDX file
- [ ] Add layer info icon to trigger layer info modal
- [ ] Display of layer info modal

faustoperez:
Edited the comment above to make it link to #811 

<img width="1728" alt="US GHG Center - Timeline + data panel - options" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/b093ca7e-ee46-4478-9316-9a95f136e03e">
<img width="900" alt="Info modal - single dataset - text" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/57977e47-c352-423a-a12e-289703f629bc">

hanbyul-here:
One more question: Where would 'Learn more' take a user to? Should this be another attribute? or is this a link to the dataset catalog page?

j08lue:
> or is this a link to the dataset catalog page?

That, yes.

faustoperez:
@hanbyul-here 

> Can the description under the title be templated from the data below? or is it something editorial that a human needs to write?
> The design includes a link for the Data Latency section. So the expectation for these strings is HTML?

The sample content for this design is taken from the layer overview page here 👉 https://earth.gov/ghgcenter/data-catalog/odiac-ffco2-monthgrid-v2022

This includes the header description, and the bullet list for the dataset attributes (including the external link). We should standardize this info to it doesn't have to be manually entered in two different places. Hope this helps!

hanbyul-here:
Closed via https://github.com/NASA-IMPACT/veda-ui/commit/24a06b30d9b04097493ffebceb2138194d04a1dd

Issue #842, opened 2024-02-13
[E & A] Follow up: documentation


hanbyul-here:
- [ ] media attribute per each layer: https://github.com/NASA-IMPACT/veda-ui/pull/834
- [ ] info attribute per layer: https://github.com/NASA-IMPACT/veda-ui/pull/843
- [ ] infoDescriptionattribute per dataset: https://github.com/NASA-IMPACT/veda-ui/pull/843

PR #843, opened 2024-02-13
Add Layer info data as a layer attribute

hanbyul-here:
Part of #823 

**Heavily edited after the last time I opened this PR**

With this pr, layer can include `info` attribute. `info` consists of
```
info:
  modal: 
    subtitle: |
      ::markdown subtitle in markdown
    contents: |
      ::markdown contents in markdown
  template:
        source: string
        spatialExtent: string
        latency: string
        unit: string
``` 

I added a sandbox page to trigger the modal.   https://deploy-preview-843--veda-ui.netlify.app/sandbox/sandboxLayerInfo 

@faustoperez Currently, there is no easy way to determine if the link the user uses is an external or an internal link. (The link is just part of the HTML that a user passes as a part of Layerdata.) So I think we might have to skip the external link icons. :[ 

I am not implementing the link on top of the header, since I can pick that component up from the other branch currently working on (https://github.com/NASA-IMPACT/veda-ui/pull/829) 
Screen Shot 2024-02-14 at 10 48 14 AM

**Details of technical implementation**: This PR required an Implementation of processing markdown in frontmatter in `stringfy-yml-func`.
- I thought about passing Markdown as it is and delegating the rendering of it to the front-end, but it will increase the load of front-end work at this point. (Integrating it with our current MDX system for rendering stories and datasets seems challenging.) 
- Initially considered using our system's markdown parsers ([remark](https://github.com/remarkjs/remark) and [rehype](https://github.com/rehypejs/rehype)), but they are pure ES modules requiring rewiring in the pipeline, which I wanted to avoid in this PR. Therefore, I chose a different markdown parser.
- It's important to note that `info.modal.subtitle` and `info.modal.contents` should only contain markdown, not MDX, to exclude custom components.

### DIFF
--- a/app/scripts/components/sandbox/layer-info/index.js
+++ b/app/scripts/components/sandbox/layer-info/index.js
@@ -1,30 +1,46 @@
-import React from 'react';
+import React, { useState, useCallback } from 'react';
 import { datasets } from 'veda';
+import LayerInfoModal from '$components/exploration/components/layer-info-modal';
 import Block from '$components/common/blocks';
-
 import { ContentBlockProse } from '$styles/content-block';
 
 export default function SandboxMDXPage() {
-  const data = datasets['no2'].data.layers[0];
-  const layerInfoMdxContent = data.info?.description;
-  const layerInfoTable = data.info.table;
+  const [revealed, setRevealed] = useState(false);
 
-  return layerInfoMdxContent ? (
-    <Block>
-      <ContentBlockProse>
-        <h2>{data.name} Layer Info</h2>
-        <div dangerouslySetInnerHTML={{ __html: layerInfoMdxContent }} />
-        <ul>
-          {Object.keys(layerInfoTable).map((key) => {
-            return (
-              <li key={key}>
-                <b>{key} : </b> <span>{layerInfoTable[key]}</span>
-              </li>
+  const openModal = useCallback(() => setRevealed(true), []);
+  const closeModal = useCallback(() => setRevealed(false), []);
+
+  const layer = datasets['no2'].data.layers[0];
+  return layer && layer.info ? (
+    <>
+      <Block>
+        <ContentBlockProse>
+          <button type='button' onClick={openModal}>
+            Open {layer.name} Layer Information Modal
+          </button>
+          {revealed && (
+            <LayerInfoModal
+              revealed={revealed}
+              close={closeModal}
+              datasetLayer={layer}
+            />
+          )}
+        </ContentBlockProse>
+      </Block>
+      <Block>
+        <ContentBlockProse>
+          <h4> Templated layer info</h4>
+          {Object.keys(layer.info.template).map((key, idx, arr) => {
+            const currentValue = layer.info.template[key];
+            return idx !== arr.length - 1 ? (
+              <span>{currentValue} · </span>
+            ) : (
+              <span>{currentValue} </span>
             );
           })}
-        </ul>
-      </ContentBlockProse>
-    </Block>
+        </ContentBlockProse>
+      </Block>
+    </>
   ) : (
     <div> Cannot find layer info fron dataset no2</div>
   );
--- a/app/scripts/components/sandbox/layer-info/index.js
+++ b/app/scripts/components/sandbox/layer-info/index.js
@@ -4,13 +4,16 @@ import LayerInfoModal from '$components/exploration/components/layer-info-modal'
 import Block from '$components/common/blocks';
 import { ContentBlockProse } from '$styles/content-block';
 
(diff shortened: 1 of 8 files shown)


parcel-resolver-veda/index.d.ts
```diff
@@ -56,11 +56,23 @@ declare module 'veda' {
...
     DAY = 'day'
   }
+  export interface LayerFrontmatterInfo {
+    info: {
+      description: string;
+      table: {
+        source: string;
+        spatialExtent: string;
+        latency: string;
+        unit: string;
```
j08lue, line 66: I guess this is the Sandbox thing, right? Could it basically just take whatever is in the config object and render that, rather than hard-coding keys?

mock/datasets/no2.data.mdx
```diff
@@ -55,6 +55,20 @@ layers:
...
+
+        - list item 1
+        - lis item 2
+
+        [A tag test](https://developmentseed.org)
+      table:
+        source: NASA
+        spatialExtent: Global
+        latency: Monthly
+        unit: unit
```
j08lue, line 71: We will want to have nicer names for the keys - capitalized, with spaces and possibly unicode chars, etc.

```suggestion
        Source: NASA
        Spatial extent: Global
        Latency: Monthly
        Units: 10¹⁵ molecules cm⁻²
```

If that does not work nicely with the current object structure in YAML, perhaps we can use something like this?

```yaml
        - title: Source
          value: NASA
        - title: Spatial extent
          value: Global
```

or

```yaml
        source:
          title: Source
          value: NASA
        spatialExtent: 
          title: Spatial extent
          value: Global
```

j08lue (changes requested):
We need nice property titles, not just slug-like keys, please.

parcel-resolver-veda/index.d.ts
hanbyul-here: Yes it can. But that approach can be error-prone, and we do need to have at least these 4 to display at the layer level on the timeline if I am understanding the design right (unless we are fine with the data being irregular)

mock/datasets/no2.data.mdx
hanbyul-here: This change is fine but to clarify my understanding: 
Maybe my sandbox example was confusing (that the example renders the key name) from the current design, I only see this table data will be used programmatically (key will never be rendered directly, and will be used only to fetch the data from the layer.) Considering we offer description as markdown (which this information will be included as markdown list), do we need to make key name human-friendly?

hanbyul-here:
I think I poorly communicated what the code does in this PR. I will close this for now and open it again after making it clear.

j08lue (approved):
Looks good and the choice of making it all Markdown is solid for now (see comment below).

image

mock/datasets/no2.data.mdx
```diff
@@ -55,6 +55,25 @@ layers:
...
+          Global, monthly 1 km resolution dataset of fossil fuel carbon dioxide emissions, version 2022
+        contents: |
+          ::markdown
+          - Temporal Extent: January 2000 - December 2021
+          - Temporal Resolution: Monthly
+          - Spatial Extent: Global
+          - Spatial Resolution: 1 km x 1 km
+          - Data Units: Tons of carbon per 1 km x 1 km cell (monthly total)
+          - Data Type: Research
+          - Data Latency: Updated annually, following the release of an updated [BP Statistical Review of World Energy report](https://www.bp.com/en/global/corporate/energy-economics.html)
```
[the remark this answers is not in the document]
hanbyul-here: Yeah, I agree. Especially since our current application also asks for structured data (why `info` has a separate section for `template`). We would need more streamlined thoughts about how to handle this layer information before though. I can't think of a straightforward way of having both descriptive layerinfo and structured layerinfo without manual edits and duplication of the data.

app/scripts/components/sandbox/layer-info/index.js
```diff
@@ -0,0 +1,47 @@
...
+              <span>{currentValue} · </span>
+            ) : (
+              <span>{currentValue} </span>
+            );
+          })}
+        </ContentBlockProse>
+      </Block>
+    </>
+  ) : (
+    <div> Cannot find layer info fron dataset no2</div>
```
sandrahoang686, line 45: 🔧 : Mispelled `from` and should `no2` be hardcoded here?

sandrahoang686 (approved):

hanbyul-here:
I moved markdown level data to the dataset (`infoDescription` - please let me know if there is a better way to name these.), left `info` to layer level. Added these two to documentation ticket: https://github.com/NASA-IMPACT/veda-ui/issues/842

PR #845, opened 2024-02-15
828 layer select modal

hanbyul-here:
@faustoperez 

1. This component lives on the other branch right now. I will do proper styling once everything is merged on main branch.
2. I removed this section because the one-liner information belongs to the layer level (not the collection level)

Screen Shot 2024-02-15 at 3 44 24 PM

TODO:
- [x] Text highlight on the description of the collection
- [x] Show more button styling
- [x] Modal height - needs to be close to max

- [x]  2/3 length for layer selection modal main description
- [x] the number (layer selected) showing up on 0

Preview link: https://deploy-preview-845--veda-ui.netlify.app/exploration 
(@faustoperez , let me know if you want to see it in GHG instance context)

### DIFF
--- a/app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
+++ b/app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
@@ -1,34 +1,14 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React from 'react';
 import styled, { css } from 'styled-components';
-import { useAtom } from 'jotai';
-
-import { DatasetData, DatasetLayer } from 'veda';
-import {
-  Modal,
-  ModalBody,
-  ModalFooter,
-  ModalHeader
-} from '@devseed-ui/modal';
 import { glsp, themeVal } from '@devseed-ui/theme-provider';
-import { Button } from '@devseed-ui/button';
-import { Subtitle } from '@devseed-ui/typography';
 import {
   CollecticonPlus,
   CollecticonTickSmall,
-  CollecticonXmarkSmall,
   iconDataURI
 } from '@devseed-ui/collecticons';
+import { DatasetLayerCardProps } from './';
 
-import { timelineDatasetsAtom } from '../../atoms/datasets';
-import {
-  allDatasets,
-  datasetLayers,
-  findParentDataset,
-  reconcileDatasets
-} from '../../data-utils';
-
-import RenderModalHeader from './header';
-
+import { DatasetClassification } from '$components/common/dataset-classification';
 import EmptyHub from '$components/common/empty-hub';
 import {
   Card,
@@ -36,9 +16,8 @@ import {
   CardMeta,
   CardTopicsList
 } from '$components/common/card';
-import { DatasetClassification } from '$components/common/dataset-classification';
+import TextHighlight from '$components/common/text-highlight';
 import { CardSourcesList } from '$components/common/card-sources';
-import DatasetMenu from '$components/data-catalog/dataset-menu';
 import { getDatasetPath } from '$utils/routes';
 import {
   getTaxonomy,
@@ -46,17 +25,9 @@ import {
   TAXONOMY_TOPICS
 } from '$utils/veda-data';
 import { Pill } from '$styles/pill';
-import {
-  Actions,
-  useBrowserControls
-} from '$components/common/browse-controls/use-browse-controls';
-import { prepareDatasets, sortOptions } from '$components/data-catalog';
-import Pluralize from '$utils/pluralize';
-import TextHighlight from '$components/common/text-highlight';
 
 const DatasetContainer = styled.div`
-  height: 100%;
-  min-height: 0;
+  height: auto;
   display: flex;
   margin-bottom: ${glsp(2)};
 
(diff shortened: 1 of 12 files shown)


app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+const allDatasets = rawAllDatasets.map(currentDataset => {
+  return {
+    ...currentDataset,
+    layers: currentDataset.layers.map(l => ({
+      ...l,
+      parentDataset: {
+        id: currentDataset.id,
+        name: currentDataset.name
+        // infoDescription: currentDataset.infoDescription,
```
hanbyul-here, line 91: @sandrahoang686 This is where I set up parentDataset available to layer. 
I thought about putting all the needed information on here (like infoDescription HTML and taxonomies) but decided to keep it minimal since this is enough information for a layer to get the parent data it needs. (Like how data card gets taxonomy data using this info https://github.com/NASA-IMPACT/veda-ui/pull/845/files#diff-a989668bae4be591a5f4ab7f357c63286837ce1702280c2bf988f72c758e5cb6R180)  How do you think?

j08lue (commented):
😍

app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
```diff
@@ -0,0 +1,238 @@
...
+  padding: 0 ${glsp(0.5)};
+  color: ${themeVal('color.surface')};
+`;
+
+const DatasetHeadline = styled.div`
+  display: flex;
+  gap: ${glsp(1)};
+  margin-bottom: ${glsp(1)};
+`;
+const DatasetDescription = styled.p``;
```
sandrahoang686, line 62: 🔧 : lets remove this

app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
```diff
@@ -0,0 +1,238 @@
...
+interface ModalContentComponentProps {
+  search: string;
+  selectedIds: string[];
+  displayDatasets: any[];
+  onCheck:  (id:any) => void;
+}
+
+export default function ModalContentComponent(props:ModalContentComponentProps) {
+  const { search, selectedIds, displayDatasets, onCheck } = props;
+  return(<>
```
sandrahoang686, line 81: 🔧 : Lets get rid of these open tags, we should be good here because everything is wrapped in `DatasetContainer` tag

app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
```diff
@@ -0,0 +1,238 @@
...
+  <DatasetContainer>
+    {displayDatasets.length ? (
+      <div>
+      {displayDatasets.map(currentDataset => (
+        <SingleDataset key={currentDataset.id}>
+          <DatasetIntro>
+            <DatasetHeadline>
+            <h4>{currentDataset.name}</h4>
+            {/* <Subtitle><LayerInfoLiner info={currentDataset.subtitle}/></Subtitle> */}
+            {currentDataset.countSelectedLayers > 0 && <DatasetSelectedLayer><span>{currentDataset.countSelectedLayers} selected </span> </DatasetSelectedLayer>}
```
sandrahoang686, line 90: ❓ : There is commented code sprinkled through out, im guessing this is still kindof WIP? we should get rid of these before merging ~

🔧 : typing `displayDatasets` would allow us to type `countSelectedLayers` to `number` and then we can just do `{currentDataset.countSelectedLayers  && ...` instead

app/scripts/components/exploration/components/dataset-selector-modal/footer.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <>
+      <LayerResult aria-live='polite' className='selection-info'>
+        {selectedIds.length
+          ? <><LayerNumberHighlight>{selectedIds.length} </LayerNumberHighlight> <div>{pluralize({ singular: 'layer', plural: 'layers', count: selectedIds.length})} selected</div></>
+          : 'No data layers selected'}
+      </LayerResult>
+      {!isFirstSelection && (
+        <Button variation='base-text' onClick={close}>
+          Cancel
+        </Button>
```
sandrahoang686, line 43: ❓ : In what case would this show up? I couldn't trigger it to render even when selecting multiple layers 🤔 Also is it okay for both these buttons to appear? if not, we should make this a ternary instead with the `Add to map` button

app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+        firstRevealRef.current = false;
+        return;
+      }
+      onAction(Actions.CLEAR);
+    }
+  }, [revealed]);
+
+  // Filters are applies to the veda datasets, but then we want to display the
+  // dataset layers since those are shown on the map.
+  const displayDatasets = useMemo(
```
sandrahoang686, line 155: 🔧 : I think it would be a good idea to type the return value here and then we can update the `ModalContentComponentProps` to receive this proper type for `displayDatasets`

maybe...
```
interface DisplayDatasetLayerInfo {
    countSelectedLayers: any;
    featured?: boolean | undefined;
    id: string;
    name: string;
    infoDescription?: string | undefined;
    taxonomy: Taxonomy[];
    description: string;
    usage?: DatasetUsage[] | undefined;
    media?: Media | undefined;
    layers: DatasetLayer[];
    related?: RelatedContentData[] | undefined;
    disableExplore?: boolean | undefined;
}
```

But if this could extend a current type, even better. This way we could probably use the same type for the layer cards if need be?

app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
sandrahoang686: 💭 : That makes sense, I think this minimal approach is the better approach to go 👍🏼 . I'm thinking though that it might be better to move this method in `../data-utils` ? And we should type the return value as well even if its just used once or twice including the data-layer-cards.

app/scripts/components/exploration/components/dataset-selector-modal/footer.tsx
hanbyul-here: Ah yes. This should have been in the conditional (Because we allow empty layers now). Thank you for catching this 🙇  but I am not clear on your concerns about both buttons appearing simultaneously. What potential problem do you see?

app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
hanbyul-here: Typing is a great idea. If my code needs to be clarified, this dataset is only to display the datasets on the modal. (This data does not get saved into Atom.)! Which removes the necessities of properties such as `infoDescription`.

app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
```diff
@@ -0,0 +1,238 @@
...
+          })});
+        background-color: ${themeVal('color.base')};
+      }
+      &:hover ::before {
+        background-color: ${themeVal('color.primary')};
+      }
+    `}
+`;
+
+function DatasetLayerCard(props: DatasetLayerCardProps) {
```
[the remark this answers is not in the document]
hanbyul-here: I agree that it is generally better to break down the components. (I broke down the modals into header, content, and footer for this reason.) I think there is a line that I drew. Ex. My thought process to leave this card is that the current structure (for me) clearly shows where each component comes from which component and this card component is only used in the body. - But this kind of personal judgment can confuse the code more, so I am for having a standard to have separate files for components!

sandrahoang686:
I'm pretty much good with this but we should make the effort to remove commented out code

app/scripts/components/exploration/components/dataset-selector-modal/content.tsx
hanbyul-here: typing is a great idea, but `currentDataset.countSelectedLayers > 0 &&` seems to be still necessary; otherwise, `0` will be rendered.

sandrahoang686 (approved):

PR #853, opened 2024-02-22
Update A&E Activation Docs

sandrahoang686:


app/scripts/components/exploration/README.md
```diff
@@ -2,8 +2,8 @@
 
 #### How to activate A&E page
 
-Add a line below in your `.env` file to activate the A&E page. The A&E page will replace Data Catalog, and Data Analysis page.
-Mind that the A&E page requires veda-ui version to be at least v4.0.
+Add a line below in your config instance's `.env` file to activate the A&E page. The A&E page will replace Data Catalog, and Data Analysis page.
+This will require that the A&E page have veda-ui version to be at least v4.0.
```
j08lue, line 6:
```suggestion
This feature is available from veda-ui >= v4.0 - make sure your instance is up-to-date.
```

j08lue (changes requested):
Good call to update this statement.

This is just my own word-smithing - I think the second sentence did not get more clear, since it in your version reads a bit as if the veda-ui version is related to the A&E page, while it is the other way around, I guess.

I made a new suggestion, what do you think? Please feel free to accept or change otherwise.

j08lue (approved):
👌

Issue #856, opened 2024-02-22
[E & A] Fast follow ups after internal review


faustoperez:
This is a placeholder ticket if there are any fast design follow-ups after enabling the new E&A release
```[tasklist]
### Tasks
- [x] select modal - update filter to remove redundant "Name" option
- [x] card - remove link from collection title
- [x] info modal - change button text to "Open in Data Catalog"
- [x] select modal - remove link from collection title on top of the modal
- [x] analysis - add stat analysis menu back to chart ( @faustoperez to follow up with design) https://github.com/NASA-IMPACT/veda-ui/pull/860
- [x] info modal - remove link from collection title
- [x] select modal - update `show/hide` filters to be optional depending on view https://developmentseed.slack.com/archives/C03LGTM2SSH/p1709062753753879?thread_ts=1709050497.517369&cid=C03LGTM2SSH
- [x] Update linking to Data Catalog to open a new tab (ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/861))
```

faustoperez:
@sandrahoang686 @j08lue Here's a quick mockup of what the analysis options could look like. The `cogs` icon should stick to the top right corner and always be visible (during analysis mode).

<img width="1728" alt="US GHG Center - Timeline + analysis options" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/f75c861e-f631-4836-adaa-04918c7db8c3">

As a side note, in the near future we should revisit the analysis chart to work on the following:

- Timeline analysis states
- add 0 reference to vertical legend
- show multiple charts
- expand/collapse etc.
- chart tooltip
- chart styles
- etc.

sandrahoang686:
PR #859 
Accomplishes the first 4 tasks...
- select modal - update filter to remove redundant "Name" option
- card - remove link from collection title
- info modal - change button text to "Open in Data Catalog"
- select modal - remove link from collection title on top of the modal

Analysis change will take a bit longer so separate PR to follow

sandrahoang686:
This is also to be done: https://github.com/NASA-IMPACT/veda-ui/issues/861

sandrahoang686:
For task => select modal - update show/hide filters to be optional depending on view

https://github.com/NASA-IMPACT/veda-ui/pull/862#pullrequestreview-1906792573 merged

Issue #857, opened 2024-02-22
[E & A] Review analysis on new E&A interface


faustoperez:
### Issue
We are missing the previous selector to enable Max, Min, Avg. etc. metrics on the analysis results due to the new design / implementation. Also, the previous analysis results page included more than one chart.

### Acceptance criteria

- [x] Review and propose a design solution to toggle chart metrics
- [x] Review expanded dataset panel styles
- [x] Review how to show multiple charts from a single dataset analysis
- [x] After triggering an analysis: fly and zoom the map to AOI; zoom the timeline to focus on the selected analysis period
<img width="326" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/23f0ceac-a62d-4211-bf27-4fb153713373">

### Implementation (design [linked below](https://github.com/NASA-IMPACT/veda-ui/issues/857#issuecomment-1985810963))
- [ ] ~Add timeline chart widgets~
- [ ] ~The chart widget should open up the analysis chart menu~
- [x] When starting an analysis, both map and timeline zoom to selected area
- [x] Analysis chart lines are updated to design
- [x] Analysis chart tooltip is updated to design
- [ ] ~Edgecase behaviors~

faustoperez:
Review and propose a design solution to toggle chart metrics addressed in #860

faustoperez:
Related: Review date selection on the timeline #833

faustoperez:
Hi, I have updated the Figma file for this ticket.

Link to Figma file 👉 https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=1127%3A2201&mode=design&t=fGtGK3DblwdZpUGr-1

I am experimenting with a new format for design communication and organization to ensure that each design element is linked to its corresponding issue and the reasoning behind the design is documented. Let me know if this works 👍 

<img width="1394" alt="image" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/78dea1fc-9008-4de9-b4de-55b035de6c9f">

sandrahoang686:
Note:
* Download feature available from old analysis view
* The analysis options including collapser should appear for each layer chart
* When layer chart is zoomed out and there is no space for the analysis options, hide them

aboydnw:
@hanbyul-here This looks partially complete, am I interpreting that correctly? Should we break off a new ticket to represent whatever work is incomplete? Might be easier to follow that way.

[...]

PR #860, opened 2024-02-27
Add analysis metrics menu to chart

sandrahoang686:
This PR adds the analysis metrics menu back to each layer list item in the explorations page. 

Ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/856)
Design [here](https://github.com/NASA-IMPACT/veda-ui/issues/856#issuecomment-1964612735)

Question for @faustoperez related to icon button location. I moved the button closer to the right-aligned because of the overlapping arrows. Please let me know what you think. I discuss this more in the demo video if you can take a look. Thanks!

Demo vid:
https://www.loom.com/share/56d013aade7945fdb40bff8cd2f3dc56

### DIFF
--- a/app/scripts/components/exploration/components/datasets/dataset-chart.tsx
+++ b/app/scripts/components/exploration/components/datasets/dataset-chart.tsx
@@ -3,12 +3,15 @@ import { useTheme } from 'styled-components';
 import { extent, scaleLinear, ScaleTime, line, ScaleLinear } from 'd3';
 import { useAtomValue } from 'jotai';
 import { AnimatePresence, motion } from 'framer-motion';
-
+import styled from 'styled-components';
+import {
+  CollecticonCog,
+} from '@devseed-ui/collecticons';
 import { isExpandedAtom } from '../../atoms/timeline';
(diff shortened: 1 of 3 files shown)


faustoperez (changes requested):
Hi Sandra, I love the videos! Good call moving the icon to the right so it doesn't overlap the chart.

Could you please move it together with the units a little bit to the left (say 4-8px) so it has a bit of space? It's a little bit too tight now. Thanks!

image

faustoperez:
As a side note related to this PR @j08lue @sandrahoang686 we need to review the behavior of the interface after triggering an analysis. I suggest:

- Fly and zoom the map to AOI
- Zoom the timeline to focus on the selected analysis period

I suggest we tackle that on #857

j08lue (approved):
Metrics selection works well with the new menu. I am not sure how well it can be found or we should use an icon that hints more at that these are chart-related options, now that the icon is not floating above the lines it is related to, but we can see in user testing, whether that is an issue and address it together with other planned design changes, if need be.

faustoperez (approved):
It's ok for now, might review the design at a later stage 👍 Thanks Sandra!

PR #862, opened 2024-02-28
Make filters button optional for datasets

sandrahoang686:
This is to remediate the issue discussed from this slack thread [here](https://developmentseed.slack.com/archives/C03LGTM2SSH/p1709050497517369)

This PR makes the filters button in the browser controls which currently lives in the `Data Catalog` and Datasets Selector Modal for A&E optional.

**Validation:**
* Make sure the Filters button does not appear on the `Data Catalog` Page
* Make sure the Filters button does appear for the Datasets Selector Modal for A&E

### DIFF
--- a/app/scripts/components/common/browse-controls/index.tsx
+++ b/app/scripts/components/common/browse-controls/index.tsx
@@ -77,6 +77,7 @@ const ButtonPrefix = styled(Overline).attrs({ as: 'small' })`
 interface BrowseControlsProps extends ReturnType<typeof useBrowserControls> {
   taxonomiesOptions: Taxonomy[];
   sortOptions: FilterOption[];
+  showMoreButtonOpt?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


slesaad:
We still want the filters to be available when the button is not present, no?

sandrahoang686:
> We still want the filters to be available when the button is not present, no? `showFilters` state should probably default to `true`?

@slesaad  Didn't realize the filters disappeared, thanks for catching! The original behavior for the data selection modal (A&E) was hiding the filters so I made default state conditional instead.

In Data Catalog now, there should be no more `Show/Hide Filters` button and the filters should be automatically be shown.
In Data Selection Modal now, there should be `Show/Hide Filters` with `Show Filters` as the default behavior and the filters hidden.

slesaad (approved):
thanks 🙇

PR #870, opened 2024-03-05
Open all external links in a new tab

sandesh-sp:
Closes https://github.com/NASA-IMPACT/veda-ui/issues/873

### DIFF
--- a/app/scripts/components/common/smart-link.tsx
+++ b/app/scripts/components/common/smart-link.tsx
@@ -12,7 +12,7 @@ export default function SmartLink(props: SmartLinkProps) {
   const { to, ...rest } = props;
 
   return /^https?:\/\//.test(to) ? (
-    <a target='_blank' rel="noopener noreferrer" href={to} {...rest} />
+    <a target='_blank' rel='noopener noreferrer' href={to} {...rest} />
   ) : (
     <Link to={to} {...rest} />
   );
@@ -29,7 +29,7 @@ export function CustomLink(props: CustomLinkProps) {
   const { href, ...rest } = props;
 
   return /^https?:\/\//.test(href) ? (
-    <a target='_blank' rel="noopener noreferrer" href={href} {...rest} />
+    <a target='_blank' rel='noopener noreferrer' href={href} {...rest} />
   ) : (
     <Link to={href} {...rest} />
   );
--- a/app/scripts/components/common/smart-link.tsx
+++ b/app/scripts/components/common/smart-link.tsx
@@ -12,7 +12,7 @@ export default function SmartLink(props: SmartLinkProps) {
   const { to, ...rest } = props;
 
   return /^https?:\/\//.test(to) ? (
-    <a target='_blank' href={to} {...rest} />
+    <a target='_blank' rel="noopener noreferrer" href={to} {...rest} />
   ) : (
     <Link to={to} {...rest} />
   );
@@ -29,7 +29,7 @@ export function CustomLink(props: CustomLinkProps) {
   const { href, ...rest } = props;
 
   return /^https?:\/\//.test(href) ? (
-    <a target='_blank' href={href} {...rest} />
+    <a target='_blank' rel="noopener noreferrer" href={href} {...rest} />
   ) : (
     <Link to={href} {...rest} />
   );
--- a/app/scripts/components/common/smart-link.tsx
+++ b/app/scripts/components/common/smart-link.tsx
@@ -5,10 +5,6 @@ interface SmartLinkProps {
   to: string;
 }
 
-interface CustomLinkProps {
-  href: string;
-}
-
 /**
  * Switches between a `a` and a `Link` depending on the url.
  */
@@ -16,13 +12,19 @@ export default function SmartLink(props: SmartLinkProps) {
   const { to, ...rest } = props;
 
   return /^https?:\/\//.test(to) ? (
-    <a href={to} {...rest} />
+    <a target='_blank' href={to} {...rest} />
   ) : (
     <Link to={to} {...rest} />
   );
 }
 
+interface CustomLinkProps {
+  href: string;
+}
 
+/**
+ * For links defined as markdown, this component will open the external link in a new tab.
+ */
 export function CustomLink(props: CustomLinkProps) {
   const { href, ...rest } = props;
 
--- a/app/scripts/components/common/smart-link.tsx
+++ b/app/scripts/components/common/smart-link.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 
+import { getBoolean } from 'veda';
+
 interface SmartLinkProps {
   to: string;
 }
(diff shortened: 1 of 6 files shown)


slesaad:
@sandesh-sp i tried the deploy preview link and this external link opened in the same tab 👀 
image

sandesh-sp:
@slesaad I didn't consider the external links on the cards but only within the stories. Let me look at that right away.

sandesh-sp:
@slesaad External links present on cards should also open in a new tab now

j08lue (changes requested):
Thanks for the contribution, @sandesh-sp! Please always reference the ticket you are working off of.

I guess it was this one:

- https://github.com/US-GHG-Center/veda-config-ghg/issues/315

But I had not considered that we would change this centrally in VEDA UI, so I now created a ticket in this repo for it:

- https://github.com/NASA-IMPACT/veda-ui/issues/873

As I stated in the ticket, we need to clarify: Would it be very complex to make this behavior optional?

app/scripts/components/common/smart-link.tsx
```diff
@@ -12,8 +12,25 @@ export default function SmartLink(props: SmartLinkProps) {
...
+ * For links defined as markdown, this component will open the external link in a new tab.
+ */
+export function CustomLink(props: CustomLinkProps) {
+  const { href, ...rest } = props;
+
+  return /^https?:\/\//.test(href) ? (
+    <a target='_blank' rel='noopener noreferrer' href={href} {...rest} />
+  ) : (
+    <Link to={href} {...rest} />
+  );
```
sandrahoang686, line 35: This gives us no other options but to open external links in a new tab. Not sure if we want this behavior though... This should probably be decided on the instance level?

slesaad:
@j08lue @sandrahoang686 yes, we do want to make the behavior optional. @sandesh-sp is looking at making it configurable via the config file in the the veda-config repo probably via the https://github.com/NASA-IMPACT/veda-config/blob/develop/veda.config.js file or an env var?

j08lue:
> via the `veda.config.js` file or an env var?

I'll defer to @sandrahoang686 to advise. 

For the E&A feature flag, we chose `.env`, like [here](https://github.com/US-GHG-Center/veda-config-ghg/pull/312/files#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055d).

sandrahoang686:
@slesaad @sandesh-sp I would recommend to make this type of configuration in the `veda.config.js` file since we use that file to specify overrideable behavior to things like stylistic changes already. Also usually env vars are typically related at the deployment level and the deployment environment while I think of configuration files at the code level and storing info in relation to the app, in this case instance.

sandesh-sp:
Creating a new pull request in veda-config-ghg
https://github.com/US-GHG-Center/veda-config-ghg/pull/326

j08lue:
I see you closed this pull request, @sandesh-sp. Is the change in VEDA UI not needed?

slesaad:
@j08lue I think that was a mistake. Reopening. @sandrahoang686 @hanbyul-here can you guys have a look at the approach followed here and let us know what you think?

sandrahoang686:
@slesaad @sandesh-sp this looks like a better path, thanks for making the updates! Could we possibly move this defining of linkProps into a more global util file? I think `app/scripts/utils/url.ts` should be fine. Then maybe update the method to work for those three cases.. something like below possibly:

```
const getLinkProps = (
  isExternalLink: string[] | boolean,
  linkTo: string,
  as?: React.ForwardRefExoticComponent<LinkProps & React.RefAttributes<HTMLAnchorElement>>,
  onClick?: () => void
) => {
  const externalLinksInNewTab = getBoolean('externalLinksInNewTab');
  return isExternalLink
  ? {
      href: linkTo,
      ...(externalLinksInNewTab
        ? {target: '_blank', rel: 'noopener noreferrer'}
        : {}),
      ...(onClick ? {onClick: onClick} : {})
    }
  : {
      ...(as ? {as: as} : {}),
      to: linkTo,
      ...(onClick ? {onClick: onClick} : {})
    };
}
```

For SmartLink it would then look like...
```
export default function SmartLink(props: SmartLinkProps) {
  const { to, ...rest } = props;
  const isExternalLink = /^https?:\/\//.test(to);
  const linkProps = getLinkProps(isExternalLink, to);
  return isExternalLink ? (
    externalLinksInNewTab ? (
      <a {...linkProps} {...rest} />
  ) : (
    <Link {...linkProps} {...rest} />
  );
}

```

sandesh-sp:
Thanks @sandrahoang686 ! I'll make the changes as suggested

sandesh-sp:
@sandrahoang686 could you have a look to the new changes. I defined getLinkProps on utils/url which was then used to get the needed attributes for the links.
ts-check is throwing an error around the type of onClick argument (currently defined as MouseEventHandler) and I'm not sure what its value should be. 
```
const getLinkProps = (
  isExternalLink: string[] | boolean,
  linkTo: string,
  as?: React.ForwardRefExoticComponent<LinkProps & React.RefAttributes<HTMLAnchorElement>>,
  onClick?: MouseEventHandler
)
```

sandrahoang686:
@sandesh-sp what is your exact error? Maybe try typing `onClick` with `onClick?: (() => void) | MouseEventHandler;`. 

Another error looks like `Link` is expecting the `to` prop and in this case it looks like typescript is not catching on sinceit looks like your code would definitely return that prop if `!isExternalLink`.. a solution might be to check that `linkProps` does return `to` or type the return?

sandesh-sp:
Thanks @sandrahoang686, this seems to be working. Could you please review it.

app/scripts/utils/url.ts
```diff
@@ -0,0 +1,27 @@
...
+    linkTo: string,
+    as?: React.ForwardRefExoticComponent<LinkProps & React.RefAttributes<HTMLAnchorElement>>,
+    onClick?: (() => void) | MouseEventHandler
+  ) => {
+    const externalLinksInNewTab = getBoolean('externalLinksInNewTab');
+    const isExternalLink = /^https?:\/\//.test(linkTo);
+    return isExternalLink
+    ? {
+        href: linkTo,
+        to: linkTo,
```
sandrahoang686, line 16: 🤔 I see

sandrahoang686, line 16: Will this work? I tested in the deploy preview and looks like the external link I clicked wasn't opening in a new tab?

Screenshot 2024-03-28 at 5 20 34 PM

sandesh-sp: @sandrahoang686 that's because we also need externalLinksInNewTab set to true on veda-config/veda.config.js. 
It is working on veda-config-ghg (https://github.com/US-GHG-Center/veda-config-ghg/pull/326) and also on my local machine.
I had to set it this way to prevent typescript error which was expecting "to" prop from the function

sandrahoang686 (approved):
lgtm :shipit:

sandesh-sp:
@j08lue @slesaad Could you please review the changes?

sandrahoang686:
@sandesh-sp Jonas is out and should be back monday

j08lue (approved):
Referring to @sandrahoang686 for an authoritative review of the technical implementation.

From a feature perspective, this seems complete now with the option to enable / disable external links to open in new tabs per-instance.

Issue #844, opened 2024-02-14
Link embedded maps to full-screen interface


j08lue:
As a user reading a data story, I would like to continue the exploration from an embedded map in a full-screen environment (Exploration or E&A), so I can look at the data more closely and change selection parameters.

# Acceptance criteria

- [ ] Validated the request
- [ ] Addressed the request in follow-up tickets, if relevant

aboydnw:
Related PR: https://github.com/NASA-IMPACT/veda-ui/pull/871

PR #871, opened 2024-03-05
added a component to handle linking embedded maps to full map view

rajdangol0077:
Closes https://github.com/NASA-IMPACT/veda-ui/issues/844

Added a FullMapLinkButton component to link embedded maps with full map view.

image

### DIFF
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -188,7 +188,6 @@ function MapBlock(props: MapBlockProps) {
         withScale
       />
     <FullMapLinkButton
-      datasetId={datasetId}
       layerId={layerId}
       date={selectedDatetime}
       compareDate={selectedCompareDatetime}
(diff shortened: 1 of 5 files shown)


rajdangol0077:
@slesaad the date format in the full map view seems to be misbehaving. Will fix this.

j08lue:
Wow, cool to see this function working, @rajdangol0077.

We need to revise the design, though, and consider how to integrate it in the embedded map features.

So I would consider this a great proof of concept that we will need to refine and work further on, which is a bit at the mercy of resources / priorities aligning. Please keep fixing any issues you are seeing, though, then the POC might get us very close to the feature.

j08lue:
> can we get some reviews, design, anything else for this?

I think this needs a good deal of thinking through. Like

1. Should these links open the comparison in E&A or a simpler map viewer - or do we offer options for both?
2. Do we just construct URLs in this component (via hard-coded URL patterns) or could the map components provide an interface for parameterization (and they construct the URLs)?

There might be more. So I think this is not a quick fix and therefore needs to wait until our backlog allows for us to spend time on this. Critical features for GHG Center R2 like the E&A release take priority right now and we also need to move the refactor forward.

aboydnw:
I'm going to close out this PR, just to keep our list of open PRs a little cleaner. We'll track follow up work in #844 in the next couple sprints and this PR is linked there as a good place to start.

Issue #865, opened 2024-02-29
[VEDA2 Refactor] Create ADR Document for decisions around Refactor


sandrahoang686:
We have a miro board that documents the technical discussions and decisions around the refactor but we should properly document in an ADR the context, technical decisions, value, and plans of the Refactor

ADR Google Doc Link [Here](https://docs.google.com/document/d/1MVBxC5xqIOjw0AqgNNr3ZW1PNIEEd_XYmtrzOh2kpNE/edit?usp=sharing)

sandrahoang686:
@j08lue will work on splitting ADR up to a high level product ADR and then a technical ADR. Technical ADR should be moved to github unders docs/adr

PR #875, opened 2024-03-06
Add ADR about VEDA V2 refactor

sandrahoang686:
Closes: https://github.com/NASA-IMPACT/veda-ui/issues/865


docs/adr/veda-v2-refactor.md
```diff
@@ -0,0 +1,98 @@
...
+![Architectural Diagram](./diagrams/veda-v2-refactor-adr-dataprovider-diagram.png)
+
+*[Miro Board Link](https://miro.com/app/board/uXjVN6lkBnc=/?share_link_id=85040810316) which documents team's brainstorming discussions, options considered, technical trade-offs, and final proposed solution in detail*
+
+### Value
+#### Product POV
+This refactor will enhance efficiency of delivering new instances and modifying page views without directly impacting the core UI logic library. It will also facilitate seamless support for a wider range of data integrations. 
+#### Developer POV
+This refactor will allow developers to move faster by removing specific domain knowledge on very custom dependencies. It will also ideally be easier to make customizations without limitations around overriding pages and components. Spinning up a new instance and making modifications to page views would now be easier with routing and composability moving to the instance level. This will also allow different data integrations to be easily added without having to touch core components. 
+
```
sandrahoang686, line 89: @j08lue I know you mentioned moving `Value` into `Context` or `Decision Drivers` but I wasn't too sure how to and if it made sense since the value is generated after the `Decision Outcome` ? Let me know if you think we should remove this and move it else where instead

docs/adr/veda-v2-refactor.md
```diff
@@ -0,0 +1,98 @@
...
+The proposed data layer is designed to handle all data fetching, including STAC calls, and MDX parsing. Think of it as a versatile data utilities library. Introducing this layer would enable components in the core UI library to remain data-agnostic. In the event of scaling to additional or different data sources, expansion can be easily accomplished by integrating them into this centralized layer. Smart components (larger and stateful) would then efficiently consume this data layer.
+> **Integrating with [stac-react](https://github.com/developmentseed/stac-react/tree/main)**: Because these React hooks manage data fetching to the STAC API using the Context API provider pattern, this could just be used directly as it is already its own data layer. We would have to decide where we wrap the Provider though. Ideally, at the instance level, we would wrap the provider at the highest level in the tree either around the router or specific view containers and then consume these hooks down the hierarchy on the instance side. The components in the core ui library would be prop driven so this way they can be truly data agnostic and accept whatever data passed in.
```
sandrahoang686, line 65: @oliverroick @sharkinsspatial in this section, I speak in a little more detail about having a data layer and prop driven data agnostic core ui library which seems to align with what was spoken about in the `STAC UI Planning mtg` hopefully

docs/adr/veda-v2-refactor.md
```diff
@@ -0,0 +1,98 @@
...
+4. Additional pages
+5. New CMS-typical content types like events or announcements
+
+We are expecting new instances to start using VEDA in 2024.
+
+### Challenge 1: Cumbersome customization to the degree required
+Implementing these additional customizations proved to be cumbersome with the current application structure, because they require page overrides and new customization options across the application, which grew fast in complexity.
+
+### Challenge 2: Dependency injection
+The separation of application and content is currently done via Git submodules: the VEDA UI library source tree is injected into a Configuration project and then they are built together. This works and is lightweight, but is not a very common pattern which new contributors to the project need to
```
j08lue, line 25:
```suggestion
The separation of application and content is currently done via Git submodules: the VEDA UI library source tree is injected into a Configuration project and then they are built together. This works and is lightweight, but is not a very common pattern which is an obstacle for new contributors to the project and new instances. 
```

faustoperez (approved):

j08lue (changes requested):
I am not sure the distinction between refactor and rewrite actually makes sense, @sandrahoang686. Even if we refactor - i.e. turn individual features into (compositions of) components - the instance apps will probably be built pretty much from scratch, right?

So, should we remove the fourth option?

faustoperez (approved):

sandrahoang686:
> I am not sure the distinction between refactor and rewrite actually makes sense, @sandrahoang686. Even if we refactor - i.e. turn individual features into (compositions of) components - the instance apps will probably be built pretty much from scratch, right?
> 
> So, should we remove the fourth option?

@j08lue Yeah, I guess there are two parts here. We will be refactoring the core ui logic library / component library (breaking out into core feature components, modularizing for smaller re-useable components, abstracting out data logic inside the current code repo) and then for our instances, we would be rewriting the codebase because of the architecture changes. 

Option 4 was about rewriting the entire core ui logic library from scratch as well which was considered but quickly scrapped.. so I can remove it! And then - It might make more sense for Option 3 to say `[3] Refactor component library + replace design system + rewrite instances`?

j08lue:
> Option 4 was about rewriting the entire core ui logic library from scratch as well which was considered but quickly scrapped.. so I can remove it!

I see. Let us keep the option then, to show that we considered it (and why we rejected it).

> It might make more sense for Option 3 to say [3] Refactor component library + replace design system + rewrite instances?

Yes, let us add that.

docs/adr/002-application-architecture-for-configurability.md
```diff
@@ -0,0 +1,117 @@
...
+3. Changes to navigation - main, secondary, page
+4. Additional pages
+5. New CMS-typical content types like events or announcements
+
+We are expecting new instances to start using VEDA in 2024.
+
+### Challenge 1: Cumbersome customization to the degree required
+Implementing these additional customizations proved to be cumbersome with the current application structure, because they require page overrides and new customization options across the application, which grew fast in complexity.
+
+### Challenge 2: Dependency injection
```
hanbyul-here, line 24: Nitpick: May I suggest considering a different title for this challenge? While 'Dependency Injection' is a well-established design pattern, it might be beneficial to explore a title that better reflects the unique aspects of this specific challenge.

docs/adr/002-application-architecture-for-configurability.md
```diff
@@ -0,0 +1,117 @@
...
+
+#### 2. Feature components in the core ui library composed at the instance level
+Currently, the core UI library handles the composition of each page because of its control over routing. However, pages and what they render should shift to the instance level so developers/stakeholders can control what is rendered within each page view. The core UI library should expose feature components (containers) that deliver a core feature in the VEDA dashboard such as the A&E feature. 
+
+#### 3. Modularize and create smaller reusable components
+The core UI library should also expose smaller reusable components like date pickers, form elements, analysis tools, etc… (“presentational/dumb” components). This way developers/stakeholders for example can compose a page view with the A&E feature from the core UI library and consume other reusable components to construct their page view. This also now allows developers to build their own custom components on the instance side and directly incorporate them. 
+
+#### 4. Create a Data layer that manages all data fetching
```
hanbyul-here, line 66: I'd like to clarify my understanding to ensure I've grasped this correctly.

This data utility will be consumed at the instance level and will pass the 'reconciled' data to the component (in JSON format, I think). Additionally, the utility will be optional, allowing those with a CMS system different from MDX the flexibility to format their data independently and then pass the finalized JSON to the components. In other words, we will only offer data utility for MDX files with specific frontmatter. 

If my understanding aligns with the intended implementation, could we possibly integrate this explanation into the content?

docs/adr/002-application-architecture-for-configurability.md
```diff
@@ -0,0 +1,117 @@
...
+> **Integrating with [stac-react](https://github.com/developmentseed/stac-react/tree/main)**: Because these React hooks manage data fetching to the STAC API using the Context API provider pattern, this could just be used directly as it is already its own data layer. We would have to decide where we wrap the Provider though. Ideally, at the instance level, we would wrap the provider at the highest level in the tree either around the router or specific view containers and then consume these hooks down the hierarchy on the instance side. The components in the core ui library would be prop driven so this way they can be truly data agnostic and accept whatever data passed in.
+
+#### 5. Introduce a new Design System
```
hanbyul-here, line 70: If I remember correctly, I thought we agreed that Design System should be addressed as a separate matter from this refactor, as it isn't essential for the structural modification. Did I miss anything?

hanbyul-here (commented):
Thanks for working on this ADR 🙇 I left some comments!

docs/adr/002-application-architecture-for-configurability.md
sandrahoang686: Yeah we can probably remove this and start a separate ADR for the Design system 👍🏼

j08lue: Yes, we went a bit back and forth between keeping it in one and moving it out. Makes sense to move it out, though.

I created a ticket for that here: https://github.com/NASA-IMPACT/veda-ui/issues/889 and can work on it tomorrow - including responding to the change requests here, so we can get both ADRs merged asap.

docs/adr/002-application-architecture-for-configurability.md
j08lue: Good point. Is the updated title better?

> Challenge 2: Unconventional combination mechanism for instance configuration and UI components (via Git)

hanbyul-here: Yeah I think so. Thanks!

docs/adr/002-application-architecture-for-configurability.md
hanbyul-here: Yes it is. I mainly wanted to make it clear that we are not going to offer various types of contents parser, other than MDX 👍

j08lue (approved):
Thanks for the thorough work on this. I think this is a good record for our decision now. With approval from all tagged approvers, this is good to merge, @sandrahoang686. 🙏

Issue #864, opened 2024-02-29
[E & A] Update styles for long collection and layer names on dataset cards


faustoperez:
### Issue
Some data layers and collections have long titles and potentially overflow their containers or span 2 lines, specially on dataset cards. Design guidance on what to do in those cases is needed.

### Acceptance criteria
- [x] Design guidance is provided
- [x] Implemented

faustoperez:
Hi, here are the proposed rules for overflowing text on dataset cards:

- **Truncate the collection title** so it only spans one line. It is there to provide context and vertical space is precious. It also has an icon and having it in two lines makes the design a bit unbalanced. Make it `font-weight: 400;` so it takes less horizontal space.
- **Have the data layer title span as many lines as it needs**. Align the dataset options icons to the top line if it takes more than one line. Make sure it's aligned with the text and not with the text block (it might need to have a little bit of `padding-top`).

Here's the Figma link 👉 https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=1077%3A1341&mode=design&t=q675OmFlHNwkaJwM-1


<img width="643" alt="US GHG Center - Dataset card with overflowing text" src="https://github.com/NASA-IMPACT/veda-ui/assets/12268629/1ac94e0d-9afe-4508-b153-b3059fbaf9a7">

PR #879, opened 2024-03-08
Update styles for long collection and layer names on dataset cards

sandrahoang686:
Closes Ticket https://github.com/NASA-IMPACT/veda-ui/issues/864

Screenshot 2024-03-08 at 11 17 24 AM

**Changes**
* Updates the dataset layer cards parent collection name to truncate after one line
* Updates the toolbar menu on cards to stay top aligned to layer name
* Padding updated to 16px all around

**Validation**
* Validate that dataset cards look correct with a long title
* Validate that the parent dataset titles in the **layer selection modal** are not affected by these changes and look as they are supposed too


faustoperez (approved):
Looking great @sandrahoang686 ! 👏

Issue #877, opened 2024-03-07
[Analysis] Y-axis label missing (again)


j08lue:
This issue seems to be back with the current Analysis (not E&A) feature:

- #764 

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/437f3ca3-6cbd-45af-aefb-5b9d7fa0db53)

## Acceptance criteria

- [ ] Y-axis label stays on the map

PR #883, opened 2024-03-08
Fix Y-axis label missing in the analysis page.

rajdangol0077:
Fixes https://github.com/NASA-IMPACT/veda-ui/issues/877


slesaad:
@rajdangol0077 i see that there's still some issues, look at the attached images (datasets: NO2 PT and NPP)
image - the label and values are overlapping
image - the superscript 2 is cutoff a bit

Can you see if you can fix those too?

rajdangol0077:
> @rajdangol0077 i see that there's still some issues, look at the attached images (datasets: NO2 PT and NPP) image - the label and values are overlapping image - the superscript 2 is cutoff a bit
> 
> Can you see if you can fix those too?

I could only fix this by adding width and offset to the axis and labels. Do verify and let me know if I need to make other changes.

rajdangol0077:
@slesaad I could only fix this by adding width and offset to the axis and labels. Do verify and let me know if I need to make other changes.

sandrahoang686 (approved):
Manually validated analysis page results. lgtm!

Issue #892, opened 2024-03-19
[VEDA2 Refactor] - Data Catalog Core Feature Breakout


sandrahoang686:
This ticket seeks to break out the Data Catalog into a core feature component in veda-ui. Core features are view containers that deliver a core part of the veda dashboard product and these containers compose of resuseable components - so when breaking this out, think about pieces that can be reuseable and start to modularize the stateless/presentational components as well. 

**Notes:**
* Mark new files for VEDA2 refactor with `@VEDA2-REFACTOR-WORK`

**Acceptance Criteria:**

**General**
- [ ] Reusability => modularize stateless/presentational components to be as reuseable as possible within the core feature view containers
- [ ] Feature Component should be entirely data agnostic => Separate out the data logic from this feature component into the data layer (dependent on data layer scaffolding ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/802)). _Note: Data will now have to be passed in as props so we will need to update the page views to supply this_
   - [ ] Feature Components should have stricter typed interfaces
- [ ] This component should still support current instances and be backwards compatible

**Specific**
- [ ] Users can use DataCatalog component with their own supplied list of data collections (data array of json objects)
- [ ] Users can still filter through catalog with current existing filters
- [ ] DataCatalog component is dynamic and will generate a list of cards based on supplied list of data collections
- [ ] DataCatalog needs to support style overrides but have default style (look into making card style or card component a prop with a default)

**Related to:** https://github.com/NASA-IMPACT/veda-ui/issues/801

**Resources:**
* ADR: https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/adr/002-application-architecture-for-configurability.md#decision-outcome
* [Miro Board outline Core Features](https://miro.com/app/board/uXjVNlIZVJI=/)

sandrahoang686:
Data Layer Scaffolding Draft PR https://github.com/NASA-IMPACT/veda-ui/pull/893

Issue #802, opened 2024-01-17
[VEDA2 Refactor] Exploration - Data layer scaffolding


sandrahoang686:
**Description:**
As part of the refactor work for veda-ui v2, we should decouple data sources from components by creating a data layer as part of the new core ui logic library which can be consumed at the instance level to be then fed into new feature components. We should define how this data layer structure would look like for example should it be a library or should we introduce the Context API Provider / Consumer model to also effectively manage global state?

**Acceptance Criteria:**

- [ ] Define data layer structure and pattern to use, data layer should be able to scale with new data sources
- [ ] Research on whether or not its worth bringing in the context api provider/consumer model here, currently veda-ui has "virtual modules" (`import { datasets, DatasetLayer } from 'veda';`) is it worth having global state replace these?
- [ ] Document Data Layer Architecture here 
- [ ] Create the scaffold PR to `veda-ui` repo. Ideally, this data layer should live at the root level with the core-ui library, but this is to be determined and up for thoughts, for example...
```
repo level
- veda-ui
   - ...current veda ui library....
- data-layer
```

**Resources:**
https://miro.com/app/board/uXjVN6lkBnc=/
[Virtual modules from VEDA Architect document](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/development/ARCHITECTURE.md#veda-modules)

hanbyul-here:
The current architect document has a section about virtual modules: [Virtual modules from VEDA Architect document](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/development/ARCHITECTURE.md#veda-modules)

sandrahoang686:
Also worth mentioning, we currently use [jotai](https://jotai.org/) (state management library) across the exploration components. We should decide whether or not we could keep using it or is it worth implementing the context api ourselves. Personally, from limited use already, I do like it as it looks like we dont have to implement a reducer to update state but whoever takes this ticket on, should explore more.

sandrahoang686:
Spoke with @hanbyul-here here and we agreed to keep managing state with Jotai

PR #893, opened 2024-03-19
Create minimal partial data layer scaffold starting off with Data Catalog for VEDA2 Refactor

sandrahoang686:
**Related Ticket:** 
* [[VEDA2 Refactor] Exploration - Data layer scaffolding ](https://github.com/NASA-IMPACT/veda-ui/issues/802)
* [[VEDA2 Refactor] - Data Catalog Core Feature Breakout](https://github.com/NASA-IMPACT/veda-ui/issues/892)
* 
### Description of Changes
This is a very minimal partial implementation of the data layer scaffolding to start the conversation and work on breaking out the feature components which will involve moving data logic out of the components into the data layer. This PR moves the datasets dependency out from the Data Catalog component to help kick off breaking out this Feature Component to be data agnostic

This Data Layer scaffold will be a WIP throughout the refactor and scaffold WILL EVOLVE. This is just to define a current place where data logic can be migrated too to start core feature breakouts

@hanbyul-here thoughts?

### Validation / Testing
* Data Catalog Page should work the same


tsconfig.json
```diff
@@ -8,7 +8,8 @@
       "$components/*": ["app/scripts/components/*"],
       "$styles/*": ["app/scripts/styles/*"],
       "$utils/*": ["app/scripts/utils/*"],
-      "$context/*": ["app/scripts/context/*"]
+      "$context/*": ["app/scripts/context/*"],
+      "$data-layer/*": ["data-layer/*"]
```
oliverroick, line 12: To keep the repo clean, could we move this into from the repository root to a subdirectory like src?

```suggestion
      "$data-layer/*": ["src/data-layer/*"]
```

data-layer/datasets.tsx
```diff
@@ -0,0 +1,17 @@
+import { datasets } from 'veda';
```
oliverroick, line 1: Does this import the datasets as configured the instance config?

oliverroick (commented):
I looked over this but I don’t fully understand these changes and how it will evolve in the future.

I left two questions in the code but my main question is about the `data-layer/datasets` module: How will this be used in the future? I’m assuming there will be more complexity in the module at some point; what functionality are we planning to implement in the module in the future?

oliverroick:
Thanks @sandrahoang686 that's clear now. As for the work you've put in the PR, that looks all good to me. 

What we need to consider though is how the changes you plan in future relate to and affect to the[ discussion/spike for the parcel builder](https://github.com/NASA-IMPACT/veda-ui/issues/896) the @hanbyul-here is working on. From what Daniel is saying in the ticket, the parcel builder not only parses the content but also generates option lists that allow filtering for content. You mentioned data fetching in hooks, were you planning that this happens on-the-fly or are we still going to use a fair amount of pre-processing?

sandrahoang686:
@oliverroick yeah I didn't realize how much the parcel resolver was doing... and I haven't thought about it. I'm thinking though and if I understood your question, seems like the parcel resolver is generating the list of taxonomies and datasets during build time but i'm thinking can we do something similar with something like NextJs's pre-rendering and static generation? So both? We initially fetch at build time with something like `getStaticProps` if using nextjs and this will not be managed by the hooks but the app can be pre-rendered and then fetch again on the client side when we need to?

What do you think?

tsconfig.json
hanbyul-here: ☝️ Can we put `data-layer` in `app/scripts`? That is where all the 'scripts' are right now.

sandrahoang686: @oliverroick @hanbyul-here should I move it under `app/scripts/*` or create a new directory to do `src/data-layer/*`? I will say I personally dont understand why we have a dir under app named `/scripts` and why it was named that but maybe we can revisit naming later.. lmk which you both prefer here

oliverroick: If we want to keep it all in the same place it should go to `app/scripts`. But I agree, we should revisit that at some point, this `app/scripts` convention we're using in some projects is odd.

sandrahoang686: Updated, can I get a re-review?

oliverroick (approved):

Issue #894, opened 2024-03-20
Sort featured stories based on publication date


hanbyul-here:
Our featured stories are not sorted by publication dates now. It will make the most sense to sort them by the dates. 

![image (1)](https://github.com/NASA-IMPACT/veda-ui/assets/4583806/8b9e14d0-ba2b-4059-9343-5e6a7a29db4d)

sandesh-sp:
Hi @hanbyul-here, just to confirm, should newer stories appear first (ordered from most recent to oldest)?

hanbyul-here:
@sandesh-sp Yes, more recent one should come first!

PR #907, opened 2024-03-29
Sort featured stories based on publication date in descending order

sandesh-sp:
### Description of Changes
Sorted the featuredItems on function FeaturedSliderSection based on their publication date so that featured stories are ordered from most recent to oldest

### Issue

Closes https://github.com/NASA-IMPACT/veda-ui/issues/894

### DIFF
--- a/app/scripts/components/common/featured-slider-section.tsx
+++ b/app/scripts/components/common/featured-slider-section.tsx
@@ -67,14 +67,14 @@ function FeaturedSliderSection(props: FeaturedSliderSectionProps) {
 
   if (!featuredItems.length) return null;
 
-  // sorting featuredItems based on dateProperty in descending order
-  if (dateProperty) {
-    featuredItems.sort((itemA: StoryData | DatasetData, itemB: StoryData | DatasetData) => {
-      const pubDateOfItemA = new Date(itemA[dateProperty]);
-      const pubDateOfItemB = new Date(itemB[dateProperty]);
-      return pubDateOfItemB.getTime() - pubDateOfItemA.getTime();
-    });
-  }
+  // Disable no-mutating rule since the copy of the array is being mutated
+  // eslint-disable-next-line fp/no-mutating-methods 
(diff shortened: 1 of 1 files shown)


app/scripts/components/common/featured-slider-section.tsx
```diff
@@ -67,6 +67,15 @@ function FeaturedSliderSection(props: FeaturedSliderSectionProps) {
...
   if (!featuredItems.length) return null;
 
+  // sorting featuredItems based on dateProperty in descending order
+  if (dateProperty) {
+    featuredItems.sort((itemA: StoryData | DatasetData, itemB: StoryData | DatasetData) => {
+      const pubDateOfItemA = new Date(itemA[dateProperty]);
+      const pubDateOfItemB = new Date(itemB[dateProperty]);
+      return pubDateOfItemB.getTime() - pubDateOfItemA.getTime();
+    });
+  }
```
hanbyul-here, line 77: Mutating the original array can lead to a challenging situation in the future. (You can find many resources with the key sentence such as 'why mutate is bad in js') 

```suggestion
  // Disable no-mutating rule since we are mutating the copy of the array
  // eslint-disable-next-line fp/no-mutating-methods 
const sortedFeaturedItems = dateProperty? [...featuredItems].sort((itemA: StoryData | DatasetData, itemB: StoryData | DatasetData) => {
      const pubDateOfItemA = new Date(itemA[dateProperty]);
      const pubDateOfItemB = new Date(itemB[dateProperty]);
      return pubDateOfItemB.getTime() - pubDateOfItemA.getTime();
    }) : featuredItems;
  }
```

And use `sortedFeaturedItems` instead of `featuredItems ` at line 92.

hanbyul-here (changes requested):

hanbyul-here (approved):
TS error was mainly because TS is confused if sorted array is a mix of StoryData and DatasetData or not. Thanks for picking up the issue.

Issue #876, opened 2024-03-07
[E & A] Catalog to exploration bridge not working for LPJ dataset


j08lue:
Hitting "Explore" for the LPJ model dataset opens the E&A view:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/4e290e4f-faf5-4348-9e9c-bb846f49ec12)

https://deploy-preview-319--ghg-demo.netlify.app/exploration?search=lpjwsl-wetlandch4-grid-v1

but the filtered view does not contain any layers:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/71ab74bc-045a-40bc-a1d3-047bf1fc7989)

Even though they are otherwise available in E&A (not disabled).

Same for the TM5-4DVar dataset, but some datasets work as expected, for example ECCO-Darwin (single-layer).

## Steps to reproduce

1. On an E&A enabled preview like https://deploy-preview-319--ghg-demo.netlify.app/data-catalog
2. Open Data Catalog and find LPJ model dataset
3. Open Dataset overview page
4. Click `Explore`
5. Expected result: E&A with layer selection modal, only showing layers for LPJ dataset
6. Actual result: E&A with layer selection modal, showing the LPJ dataset header, but no layers

## Acceptance criteria

- [ ] Catalog to E&A layer selection works for LPJ
- [ ] Checked whether other datasets are affected by the same issue

j08lue:
I see what we populate the search box with is the Dataset ID (`id`), e.g.

https://github.com/US-GHG-Center/veda-config-ghg/blob/cd9bcc2fd8a73c1ef2841d3a0b2ec0a27378e751/datasets/lpjwsl-wetlandch4-grid-v1.data.mdx?plain=1#L2

https://github.com/US-GHG-Center/veda-config-ghg/blob/cd9bcc2fd8a73c1ef2841d3a0b2ec0a27378e751/datasets/tm54dvar-ch4flux-monthgrid-v1.data.mdx?plain=1#L2

For some reason, in the layer selection modal, searching by Dataset ID does not return all layers for that dataset. Perhaps the search is limited to Layer IDs? Perhaps this is the code in question?

https://github.com/NASA-IMPACT/veda-ui/blob/694a665be632a591011d012c0c9160551685fdb4/app/scripts/components/data-catalog/index.tsx#L92-L95

@hanbyul-here, this is all I could dig up, I hope you can take it from here.

PR #910, opened 2024-04-03
Include dataset id to filter layers

hanbyul-here:
**Related Ticket:** #876 

### Description of Changes
Include dataset id & name for search criteria

I am kind of surprised that dataset id & name were not search criteria yet? Anyway, this will go well with our new approach in A&E to group layers with dataset.

### Validation / Testing
Check LPJ dataset link from veda-config-ghg


j08lue:
I am now using this GHG preview for testing this fix: https://github.com/US-GHG-Center/veda-config-ghg/pull/312

Will review as soon as that is done building.

j08lue (approved):
I can confirm that the link from exploration to E&A now works: where we populate search with the dataset id, all layers for that dataset are shown.

sandrahoang686 (approved):

Issue #882, opened 2024-03-08
[E & A] Some datasets cannot be selectable together with others


j08lue:
We have a dataset that should be shown exclusively on the E&A interface, without the possibility to add others.

This is a complex ask, but let us see whether we can find a simple solution.

## Acceptance criteria

- [ ] Found and designed a solution
- [ ] Added the option
- [ ] Configured it for the relevant GHG dataset

faustoperez:
Could you please elaborate a bit on what this dataset is and the reasons behind this ask @j08lue ? Thank you!

j08lue:
As we discussed, this is basically the response to

- https://github.com/US-GHG-Center/veda-config-ghg/issues/314

In essence, users looking at the EPA inventory dataset layers should not view these layers together with other products **without context about the different datasets**. We will learn soon about what context is required.

We have basically three options:
1. Certain layers cannot be selected together with others -> change in the layer selection modal
2. When certain layer combinations are added to the map, contextual information is provided
3. Certain datasets allow single-layer view only (the old Exploration interface)

In a recent discussion among stakeholders, there was a preference for the first option - hence this ticket here. But I think if we can solve the second once nicely, that could also be a possibility.

j08lue:
Here is a preview of these layers in E&A:

https://deploy-preview-319--ghg-demo.netlify.app/exploration?datasets=%5B%5D&taxonomy=%7B%22Source%22%3A%22epa%22%7D

And this is what we should prevent - EPA totals viewed on top of other totals without context:

[...]

PR #913, opened 2024-04-04
Some datasets can only be analyzed with layers from the same source

sandrahoang686:
**Related Ticket:**

Closes https://github.com/NASA-IMPACT/veda-ui/issues/882#issuecomment-2037327990

### Description of Changes
* New dataset markdown attribute `sourceExclusive` added
* When sourceExclusive, dataset can only be selected to be analyzed with other layers from the same dataset/source
* Updates data layer selector modal
   * Remove show/hide filters
   * Allow only 4 filters with the search bar and then wrap all other filters below
   * Remove SortBy filter
   * Remove heading description to only appear in empty/not found state
   * Move note about limited selection of available datasets from top bar to the message that is shown when search and/or filters do not return any layers.

### Validation / Testing

### DIFF
--- a/parcel-resolver-veda/index.d.ts
+++ b/parcel-resolver-veda/index.d.ts
@@ -178,6 +178,7 @@ export interface LayerInfo {
    */
   export interface DatasetData {
     featured?: boolean;
+    sourceExclusive?: string;
     id: string;
     name: string;
     infoDescription?: string;
--- a/app/scripts/components/common/browse-controls/index.tsx
+++ b/app/scripts/components/common/browse-controls/index.tsx
@@ -13,7 +13,6 @@ import { DropMenu, DropTitle } from '@devseed-ui/dropdown';
 import {
   Actions,
   FilterOption,
-  TaxonomyFilterOption,
   optionAll,
   useBrowserControls
 } from './use-browse-controls';
--- a/app/scripts/components/common/browse-controls/index.tsx
+++ b/app/scripts/components/common/browse-controls/index.tsx
@@ -43,7 +43,7 @@ const FilterOptionsWrapper = styled.div`
   > * {
     flex-shrink: 0;
   }
-`
+`;
 
 const DropButton = styled(Button)`
   max-width: 12rem;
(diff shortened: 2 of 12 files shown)


parcel-resolver-veda/index.d.ts
```diff
@@ -178,6 +178,7 @@ export interface LayerInfo {
    */
   export interface DatasetData {
     featured?: boolean;
+    sourceExclusive?: string;
```
hanbyul-here, line 181: If this is a binary value, would it make sense to make it boolean?

j08lue: I see the value is actually used, so it is not a boolean flag.

https://github.com/NASA-IMPACT/veda-ui/pull/913/files#diff-e7c2bc61a035bdafe3d86d260f7cddb201aadb0195fcf1cd261139932ea00cc4R122

So, basically, you define here, which source value can exclusively be used. That is a bit redundant, though, because the is already defined under `taxonomy`, so a boolean flag would be cleaner for this simple case where only the source of the current dataset can be exclusive.

j08lue (approved):
The logic is as intended, great work! 💯 

A little improvement might be if the filters from the layer selection modal would stick. Currently, they are reset when you add layers to the map, so if you add a layer from a "source-exclusive" dataset and then return to the layer selection modal to add more, you see all layers to select from, without the (source) filter. Changing that does not need to be in scope for this issue.

parcel-resolver-veda/index.d.ts
sandrahoang686: I think I did this because some datasets look like they can have more than 1 source value like [here](https://raw.githubusercontent.com/US-GHG-Center/veda-config-ghg/develop/datasets/oco2-mip-co2budget-yeargrid-v1.data.mdx). So for me to set the filter the easiest, i opted to do it like so. lmk if there are better ideas though and I will go change 🏃🏼‍♀️

j08lue: This is great for now, thanks! We can refine later, if we learn more about the use cases and needs for these restrictions - or find out we can avoid them with other solutions.

hanbyul-here (approved):

Issue #915, opened 2024-04-05
[Data Catalog] Implement catalog enhancement with floating filter sidebar


j08lue:
Upon stakeholder feedback, we reviewed the design of the data catalog filter and dataset listing layout.

- https://github.com/US-GHG-Center/ghgc-architecture/issues/133

This has now been approved as a significant improvement and is ready for implementation

Figma Link here https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=1179%3A5867&mode=design&t=bXYpdYKR3ZMvi9bU-1

## Acceptance criteria

- [x] Implemented layout update with floating filters in data catalog

sandrahoang686:
@j08lue @faustoperez did we want to feature flag this feature? Or a way to switch between, original and new view based on some config var?

j08lue:
> did we want to feature flag this feature? Or a way to switch between original vs new view based on some config var?

Good you are asking. No, this would be a core update for all instances.

j08lue:
Let us consider this closed - the feature is implemented, although some details may still be missing. We can track those here:

- https://github.com/NASA-IMPACT/veda-ui/issues/926

PR #918, opened 2024-04-12
Data Catalog enhancement with floating filter sidebar

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/915

### Description of Changes
This pull request implements the new Data Catalog View with the filters sidebar

* Sidebar is not sticky, wasn't able to get to implementing, should we create a ticket for this as a fast follow @j08lue ?
* Some modularization and breakout was done on the card components but not much to the Data Catalog feature so the [refactor ticket ](https://github.com/NASA-IMPACT/veda-ui/issues/892)still needs to stay in progress

### Validation / Testing
**Shoulds:**
* Selecting multiple of the same taxonomy group should execute OR logic
* Selecting across taxonomy groups should execute AND logic
* Tags should persist when page refreshes
* Tags should add/remove correctly when updating filters
* Clicking on the data catalog card and then click on the pills in the info view should take you back to the data catalog filters page with the correct filters checked and you should be able to add/remove filters from there

**Should Nots:**
* These changes should not affect the `/exploration?datasets=%5B%5D` view
* These changes should not affect other any other card styles


sandrahoang686:
@faustoperez @j08lue this is still WIP but I could use an early user flow review of the new filters menu (cards are still actively being worked on and there are some quirks I still need to figure with the filters). A few questions and things to note which I wanted to flag right now. 

1. In the current view, a user can only select one value per each taxonomy filter. With this new view, that changes, a user can select multiple values per taxonomy filter. This is tricky to figure out with how we have the routing url params set up because right now the routing params works with only selecting a single value per each taxonomy group. But i'm guessing expected behavior is that we want mutually exclusive results vs non mutually exclusive results? is this correct?
2. Should the filter menu's default behavior be opened or closed?
3. When adding and removing tags, this causes the menus to move up and down, is this okay?
4. When s

sandrahoang686:
Checkin 4/18 Update:
@faustoperez @j08lue 
* Remove count and move tags above results
* The results from the filters should be non-mutually exclusive (OR not and) @faustoperez to confirm

j08lue:
> The results from the filters should be non-mutually exclusive (OR not AND)

I would say additive within earth filter, exclusive between them, i.e. `source=(NASA OR NOAA) AND gas=CO2`.

app/scripts/components/common/card/styles.tsx
```diff
@@ -0,0 +1,161 @@
...
+  display: flex;
+  flex-flow: column nowrap;
+  border-radius: ${multiply(themeVal('shape.rounded'), 2)};
+  box-shadow: ${themeVal('boxShadow.elevationD')};
+  height: 100%;
+  overflow: hidden;
+  transition: all 0.24s ease-in-out 0s;
+`;
+
+export const CardList = styled.ol`
```
hanbyul-here, line 26: `CardList` is being used by instances (ex. GHG : https://github.com/US-GHG-Center/veda-config-ghg/blob/89ec4506cd3a172dfbb6dfa6ffbde8662b448fe3/overrides/home/keypoints.tsx#L7) - We should check other instances too - We should either 1. plan to bring the changes to instances or 2. make this backward compatible.

sandrahoang686: Duplicated those two back to `/common/card/index.tsx` to support instance dependency. Once we update the instances to point to a version of veda with these new updates, we should point the imports to the new location and then remove it from this file.

app/scripts/components/data-catalog/index.tsx
```diff
@@ -155,153 +182,209 @@ function DataCatalog({ datasets }: DataCatalogProps) {
...
+        title='Data Catalog'
+        description={getString('dataCatalogBanner').other}
+      />
+      <PageHero
+        title='Data Catalog'
+        description={getString('dataCatalogBanner').other}
+      />
+
+      <FeaturedDatasets />
+      <BrowseSection>
```
hanbyul-here, line 288: This section leaves no margin on top, makes the general style not complete. (It is not obvious on UI because featured-list has margin at the bottom.) 

Screen Shot 2024-04-25 at 11 39 51 AM

app/scripts/components/data-catalog/index.tsx
```diff
@@ -155,153 +182,209 @@ function DataCatalog({ datasets }: DataCatalogProps) {
...
-                          imgAlt={d.media?.alt}
-                          footerContent={
-                            <>
-                              {topics?.length ? (
-                                <CardTopicsList>
-                                  <dt>Topics</dt>
-                                  {topics.map((t) => (
+                            />
+                            <VerticalDivider variation='light' />
+                            {/* TODO: Implement modified date: https://github.com/NASA-IMPACT/veda-ui/issues/514 */}
```
hanbyul-here, line 335: (Not introduced by this PR but) might be time to just get rid of this and close this issue

app/scripts/components/data-catalog/filters-control.tsx
```diff
@@ -0,0 +1,70 @@
+import React, {useCallback } from 'react';
+import styled from 'styled-components';
+import { Taxonomy } from 'veda';
+import SearchField from '$components/common/search-field';
+import CheckableFilters, { OptionItem } from '$components/common/form/checkable-filter';
+import { Actions, useBrowserControls } from '$components/common/browse-controls/use-browse-controls';
+
+const ControlsWrapper = styled.div<{ width?: string; }>`
+  min-width: 20rem;
+  width: ${props => props.width ?? '3rem'};
```
hanbyul-here, line 10: Am I reading right that width is smaller than min-width? then dos this style do anything?

hanbyul-here (commented):
I can do more review but that might not be what this change is needed at this point.
Two blockers I see from this PR are 1. https://github.com/NASA-IMPACT/veda-ui/pull/918/files#r1578769216 2. https://github.com/NASA-IMPACT/veda-ui/pull/918/files#diff-d569f507e8f543c166ce6a250b3c64a0b1b997d8d59ee996e0d390d3af4636bdR288 (See the review comment)  If you have bandwidth, check this change in GHG context to see if there is anything you didn't intend is happening there. 

There are improvements we can bring regarding styles, I left comments about hard-coded values. I can see that the page will benefit from subtle animations since there are many moving pieces on the page. These can be followed up while people are testing the main functionality, so once two blockers are cleared, feel free to merge it.

j08lue:
> If you have bandwidth, check this change in GHG context to see if there is anything you didn't intend is happening there.

Good call, @hanbyul-here. I created a GHG Center feature preview here:

- https://github.com/US-GHG-Center/veda-config-ghg/pull/340

hanbyul-here:
One. more thing to follow up. Footer action is not sticking to the bottom.

j08lue (changes requested):
The EMIT dataset is missing in the GHG Center preview of this feature. I have a suspicion that this is related to the `disableExplore: true` status of this dataset.

https://github.com/US-GHG-Center/veda-config-ghg/pull/340#issuecomment-2076725633

We need to make sure all datasets are shown in the main catalog.

j08lue:
https://github.com/NASA-IMPACT/veda-ui/blob/4c98710497230432c2707053dd6303e314cbbe8b/app/scripts/components/exploration/data-utils.ts#L40-L42

?

j08lue (approved):
All datasets now show up in the main catalog and it looks nice both in our mock and the GHG Center preview. Good to go!

app/scripts/components/common/card/styles.tsx
sandrahoang686: ahh I see and remember what happened now, I changed the original `CardList` name to `CardListGrid` and moved to to `/styles` this is what the ghg instance depended on. Looks like we moved this style dependency up to GHG which is good

Issue #904, opened 2024-03-26
Blank time-series chart with some datasets + selected AOI


slesaad:
# Description
With some of the datasets, a specific example being the Air Sea CO2 flux in the GHG center, the time series chart values are null even tho the `/search` returns results. See the attached screenshots.

This happens when the user draws a polygon on land where the data values don't exist. In such cases, the time-series chart should display a message instead of displaying a blank chart.

All the `/statistics` responses look like this:

```json
 "properties": {
        "statistics": {
            "b1": {
                "min": null,
                "max": null,
                "mean": null,
                "count": 2274.60009765625,
                "sum": null,
                "std": null,
                "median": null,
                "majority": null,
                "minority": null,
                "unique": 0.0,
                "histogram": [
                    [
                        0.0,
                        0.0,
                        0.0,
                        0.0,
                        0.0,
                        0.0,
                        0.0,
                        0.0,
                        0.0,
                        0.0
                    ],
                    [
                        0.0,
                        0.1,
                        0.2,
                        0.30000000000000004,
                        0.4,
                        0.5,
                        0.6000000000000001,
                        0.7000000000000001,
                        0.8,
                        0.9,
                        1.0
                    ]
                ],
                "valid_percent": 0.0,
                "masked_pixels": 4551.0,
                "valid_pixels": 0.0,
                "percentile_2": null,
                "percentile_98": null
            }
        }
```

See an example here: https://earth.gov/ghgcenter/analysis/results?start=2018-01-01T00%3A00%3A00.000Z&end=2022-12-31T00%3A00%3A00.000Z&datasetsLayers=air-sea-co2&aoi=%7C~%60eTwcvbGftcAli%7B%7D%40%7DnohDuhab%40

It will be less of a problem once E&A is out because users can see where data values exist. However, it should still be handled.

# Screenshots
![image](https://github.com/NASA-IMPACT/veda-ui/assets/7830949/6792750e-1597-4559-a59f-cab7165315bc)
![image](https://github.com/NASA-IMPACT/veda-ui/assets/7830949/eb619258-a04f-4766-a7ac-c2d656718cdf)

aboydnw:
@hanbyul-here is this ticket actually complete? I see the merged PR linked above, not sure if it was intentionally left open

j08lue:
Hmm, in the current production GHG Center, the `/statistics` API actually throws a 500 error for analyzing the air-sea flux on land (all-NaN, I assume).

[...]

PR #922, opened 2024-04-23
Handle empty result

hanbyul-here:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/904

### Description of Changes
Handle the case when there is no actual value to display from statistics requests.

- I am not sure the wording is clear. Please feel free to change. It will look like this

Screen Shot 2024-04-23 at 3 24 28 PM

- Is it valid to assume that if the 'mean' (average) value doesn't exist, then other statistical measures also don't exist?

### Validation / Testing
Try datasets that only cover ocean (such as Air-sea co2 flux) , and select aoi that doesn't have ocean coverage.

### DIFF
--- a/app/scripts/components/exploration/analysis-data.ts
+++ b/app/scripts/components/exploration/analysis-data.ts
@@ -235,6 +235,18 @@ export async function requestDatasetTimeseriesData({
       )
     );
 
+    if (layerStatistics.filter(e => e.mean).length === 0) {
+      return {
+        ...datasetAnalysis,
+        status: TimelineDatasetStatus.ERROR,
+        error: new ExtendedError(
+          'No valid statistics in the given time range and area of interest',
+          'ANALYSIS_NO_DATA'
+        ),
+        data: null
+      };
+    }
+
     onProgress({
       status: TimelineDatasetStatus.SUCCESS,
       meta: {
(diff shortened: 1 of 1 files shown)


app/scripts/components/exploration/analysis-data.ts
```diff
@@ -235,6 +235,18 @@ export async function requestDatasetTimeseriesData({
       )
     );
 
+    if (layerStatistics.filter(e => e.mean).length === 0) {
+      return {
+        ...datasetAnalysis,
+        status: TimelineDatasetStatus.ERROR,
+        error: new ExtendedError(
+          'No valid statistics in the given time range and area of interest',
```
j08lue, line 243:
```suggestion
          'The selected time and area of interest contains no valid data. Please adjust your selection.',
```

j08lue (approved):
Great we are catching this now.

I made a suggestion for different wording, adding a note to the user what they should do next - please accept, if you see fit.

(Btw, I tested this by analyzing GEOGLAM over the ocean.)

sandrahoang686 (approved):

hanbyul-here:
This PR was actually missing a component to display a proper error message, I included it in https://github.com/NASA-IMPACT/veda-ui/pull/921/commits/7594dd9e611cdf3747019b67089b047214019483

Issue #926, opened 2024-04-25
[Data Catalog] Follow up on missing details


j08lue:
We completed most of the data catalog feature update and decided to merge

- https://github.com/NASA-IMPACT/veda-ui/issues/915
- https://github.com/NASA-IMPACT/veda-ui/pull/918

but a few details remain to be implemented / fixed (see PR review comments, too)

- Make side bar floating
- Expand all filters
- more?

## Acceptance criteria

- [ ] Listed and addressed missing details

Issue #924, opened 2024-04-24
[E & A] Filter text overflows


j08lue:
When selecting long filter values in the new E&A layer selection modal, the text overflows like this:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/e8858695-ed59-425f-907d-92035995b3b7)

See e.g. https://deploy-preview-319--ghg-demo.netlify.app/exploration?datasets=%5B%5D&date=&aois=%5B%5D&dateRange=&taxonomy=%7B%22Product%20Type%22%3A%22ground-measurements%22%7D to reproduce.

I guess what I would expect is that the text is clipped.

## Acceptance criteria

- [ ] Along long filter names / values are handled nicely

j08lue:
Mostly fixed, but the filter name still overflows:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/1308b872-4328-407c-a024-e156606404af)

hanbyul-here:
It was improved but still overflowing : https://github.com/NASA-IMPACT/veda-ui/pull/927#pullrequestreview-2024276415

aboydnw:
This was inherently fixed by changing the layer selection to better match the updated data catalog UI

Issue #923, opened 2024-04-24
[E & A] Analysis fails for Alaska


j08lue:
We added analysis region presets in and in that context discovered that analysis of Alaska as well as all-US (when it includes Alaska) fails.

- #872

Unsure whether this is because of details in that geometry or the antimeridian crossing.

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/e5f77645-16e8-47a3-ad08-be6aa6c828ec)

## Acceptance criteria

- [ ] Analysis works for Alaska

j08lue:
In the UI, the steps to reproduce this issue are:
1. Go to the Exploration & Analysis interface in this preview: https://deploy-preview-319--ghg-demo.netlify.app/exploration?datasets=%5B%5D
2. Select `Alaska` from the drop-down of regions to analyze
3. Hit "Run Analysis"
4. Observe the above error message

What is going on behind the scenes is a two-step process:
1. A STAC `/search` with the geometry and time range as payload to get the items
2. One request for each item against TiTiler `/statistics`

In this case, it is the STAC `https://ghg.center/api/stac/search` call that throws `403 Forbidden`, for example with this payload: [stac-search-payload-alaska.json](https://github.com/NASA-IMPACT/veda-ui/files/15096112/stac-search-payload-alaska.json).

j08lue:
We are sending the geometry through an antimeridian fixing function before sending it to STAC search: 

https://github.com/NASA-IMPACT/veda-ui/blob/4c98710497230432c2707053dd6303e314cbbe8b/app/scripts/utils/antimeridian.ts#L164 - not sure how well that works.

The data comes from [Census.gov](http://census.gov/)’s [1:20million cartographic state boundaries](https://www.census.gov/geographies/mapping-files/time-series/geo/cartographic-boundary.html) and we store it in our app here: https://github.com/NASA-IMPACT/veda-ui/blob/main/static/public/geo-data/states/Alaska.geojson

You can see how we pull the data and conver it to GeoJSON here: https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/development/geo-data-preparation.ipynb

j08lue:
Pretty sure it has to do with the `MultiPolygon` complexity, actually: When I combine different rings of the Alaska polygon, even staying away from the antimeridian, at some point the failure starts happening.

[...]

PR #927, opened 2024-04-26
Follow up for R2 

hanbyul-here:
**Related Ticket:** Portions of https://github.com/NASA-IMPACT/veda-ui/issues/926 , https://github.com/NASA-IMPACT/veda-ui/issues/924

### Description of Changes
- Generate taxonomies depending on datasets
- Truncate texts on Preset Selector
- Truncate texts on BrowseControl 
- Leave only CONUS for country option

I moved the part that generates taxonomies of the dataset to source code, but did not move the ones which generates taxonomies of stories. (It might be worth moving it since we want to move towards not using veda-resolver in the future?) 

### Validation / Testing
Check if the changes in the descriptions are reflected.

### DIFF
--- a/app/scripts/components/data-catalog/index.tsx
+++ b/app/scripts/components/data-catalog/index.tsx
@@ -1,6 +1,6 @@
 import React, { useRef } from 'react';
 import styled from 'styled-components';
-import { DatasetData, datasetTaxonomies, getString } from 'veda';
+import { DatasetData, getString } from 'veda';
 import { Link, useNavigate } from 'react-router-dom';
 import { themeVal } from '@devseed-ui/theme-provider';
 import { VerticalDivider } from '@devseed-ui/toolbar';
@@ -32,11 +32,12 @@ import TextHighlight from '$components/common/text-highlight';
 import { Pill } from '$styles/pill';
 import { FeaturedDatasets } from '$components/common/featured-slider-section';
 import { CardSourcesList } from '$components/common/card-sources';
-import { DatasetDataWithEnhancedLayers } from '$components/exploration/data-utils';
+import { DatasetDataWithEnhancedLayers, allDatasets } from '$components/exploration/data-utils';
 import {
   getAllTaxonomyValues,
   getTaxonomy,
   getTaxonomyByIds,
+  generateTaxonomies,
   TAXONOMY_SOURCE,
   TAXONOMY_TOPICS
 } from '$utils/veda-data';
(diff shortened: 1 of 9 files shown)


j08lue (approved):
Great follow-up, that clears out remaining issues in E&A. 💯 

Tiny remaining stylistic issue with long filter names, but that is not a blocker:

image

j08lue:
I created simplified versions of the US all-territories and Alaska geometries that seem to work with the current production API, @hanbyul-here 

- #928 

We can use these if we do not solve the issue on the API side
 
- https://github.com/NASA-IMPACT/veda-ui/issues/923

app/scripts/components/data-catalog/index.tsx
```diff
@@ -189,6 +190,8 @@ function DataCatalog({ datasets }: DataCatalogProps) {
   const search = controlVars.search ?? '';
   let urlTaxonomyItems: OptionItem[] = [];
 
+  const datasetTaxonomies = generateTaxonomies(allDatasets);
```
sandrahoang686, line 193: 🔧 : The work surrounding creating the data layer abstraction is around making sure that the data layer remains our single source of truth. `allDatasets` is already doing that because its pulling `veda_datasets` from the data-layer which I had moved in the scaffolding PR. But another part that is important is that in the `data-utils` file we should have these utility methods take in the datasets as a param instead of using datasets directly (made a note of this in code). 

This is because it makes a bit more sense in my opinion to type these methods to expect a certain data structure rather than using the data directly, this is so we can create some abstraction to use these methods more flexibly when we get to pass in the data ourselves. Also with the work around creating the data-layer abstraction, we want the datasets to be provided to the core feature and the core feature should work with the dataset it was explicitly given (hence the data should be provided to these methods directly). We will have to update these methods, I made a note of it but didn't get the chance to do it myself with the scaffolding pr because it was a little out of scope.

app/scripts/utils/veda-data.ts
```diff
@@ -95,6 +97,27 @@ export function useMdxPageLoader(loader?: () => Promise<MDXModule>) {
   return pageMdx;
 }
 
+export function generateTaxonomies(data): Taxonomy[] {
```
sandrahoang686, line 100: 🔧 : lets type this param

app/scripts/components/data-catalog/index.tsx
sandrahoang686: Update made. Updated `data-catalog/container` to instead provide `allDatasets` from `exploration/data-utils` (prev it provided `allDatasetsWithEnhancedLayers` which used `allDatasets` directly). Then updated `allDatasetsWithEnhancedLayers` to be a method instead that takes in the dataset as a param and updated data-catalog to support these changes. 

This makes sure we provide the expected dataset to data-catalog and data-catalog will use the utils to manipulate the dataset explicitly given to it.

sandrahoang686 (approved):

PR #930, opened 2024-04-30
Floating Data Catalog Filters Sidebar Follow-up and starting some cleaning and modularizing

sandrahoang686:
**Related Ticket:**

Closes https://github.com/NASA-IMPACT/veda-ui/issues/926

### Description of Changes
* Make filters sidebar height `100%` when filters sidebar height (when all expanded) is greater than or equaled to the height of all the cards
* Calculate filters sidebar height with `position:sticky` when filters sidebar height (when all expanded) is less than height all of cards
* Display all filters opened now
* Move `prepareDatasets` to own logic file - cleaning
* Move layout components from Data Catalog component into the wrapper - starting to modularize

**DEMO Here:**
https://www.loom.com/share/5f88edba7ca4422d9f2636464cebf7b6?sid=62ae3536-4174-4de0-adfe-a19b2e4a5893

### Validation / Testing
* Make sure when the height of the filters sidebar (when expanded) is greater than the height of all the cards, the sidebar doesn't behave sticky
* Make sure when the height of the filters sidebar (when expanded) is less than the height of all the cards, the sidebar behaves sticky

### DIFF
--- a/app/scripts/components/common/form/checkable-filter/index.tsx
+++ b/app/scripts/components/common/form/checkable-filter/index.tsx
@@ -74,7 +74,7 @@ interface CheckableFiltersProps {
     item?: OptionItem;
     callback?: React.Dispatch<React.SetStateAction<any>>;
   }
-  showFiltersOpened?: boolean;
+  calculateHeightCallback?: (height: number) => void;
 }
 
 export interface OptionItem {
@@ -84,10 +84,18 @@ export interface OptionItem {
 }
 
 export default function CheckableFilters(props: CheckableFiltersProps) {
-  const {items, title, onChanges, globallySelected, tagItemCleared} = props;
-  const [show, setShow] = useState<boolean>(props.showFiltersOpened ?? true);
+  const {
+    items,
+    title,
+    onChanges,
+    globallySelected,
+    tagItemCleared,
+    calculateHeightCallback
+  } = props;
+  const [show, setShow] = useState<boolean>(true);
   const [count, setCount] = useState<number>(0);
   const [selected, setSelected] = useState<OptionItem[]>([]);
+  const targetRef = React.useRef<HTMLDivElement>(null);
 
   const handleChange = useCallback((e: React.ChangeEvent<HTMLInputElement>) => {
     const idInfo = e.target.id.split('&&');
@@ -115,6 +123,13 @@ export default function CheckableFilters(props: CheckableFiltersProps) {
   const isChecked = (item: OptionItem) => globallySelected.some((selected) => selected.id == item.id && selected.taxonomy == item.taxonomy);
   
   useEffect(() => {
+    if(targetRef.current && calculateHeightCallback) {
+      const height = targetRef.current.offsetHeight;
+      calculateHeightCallback(height);
+    }
+  }, [targetRef]);
+
+  useEffect(() => {
     if(!globallySelected || globallySelected.length === 0) {
       setCount(0);
     }
@@ -135,7 +150,7 @@ export default function CheckableFilters(props: CheckableFiltersProps) {
   }, [tagItemCleared, globallySelected]);
 
   return (
-    <FilterMenu>
+    <FilterMenu ref={targetRef}>
       <FilterTitle>
         <div id='title-selected'>
           <CardTitle>{title}</CardTitle>
--- a/app/scripts/components/common/form/checkable-filter/index.tsx
+++ b/app/scripts/components/common/form/checkable-filter/index.tsx
@@ -74,7 +74,6 @@ interface CheckableFiltersProps {
     item?: OptionItem;
     callback?: React.Dispatch<React.SetStateAction<any>>;
   }
-  calculateHeightCallback?: (height: number) => void;
 }
 
 export interface OptionItem {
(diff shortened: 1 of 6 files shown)


sandrahoang686:
_Demo of Local and GHG instance will be following soon..._

sandrahoang686:
Demo here: https://www.loom.com/share/5f88edba7ca4422d9f2636464cebf7b6?sid=62ae3536-4174-4de0-adfe-a19b2e4a5893
@j08lue does this ^ help answer your comment

j08lue:
Very nice and helpful demo, thanks!

> I'm guessing you closed all the filters and was hoping it would still be sticky?

Exactly

> I calculate the height of all the filters when they are open and if all the filters height is greater than height of all cards, position sticky doesn't apply.

I imagine this is quite tricky, because you may also need to scroll the side bar in some cases and it does not have its own scroll bar? 🤯 

Let me test a bit more on a GHG Center preview (I can set that up shortly) and also see what @faustoperez says, before you make any more changes.

j08lue:
Created a GHG Center preview here:
- https://github.com/US-GHG-Center/veda-config-ghg/pull/359

faustoperez (approved):
Ah I see it's more complicated than I thought in the first place, as the height of the element with the accordions and the parent need to be calculated for it to work 🤯 .

To move things forward, I am approving this PR, but let's revisit the behavior after the v2 release.

Ideally, the sidebar would stick at the bottom of the main content when scrolling, it wouldn't scroll up.

Here's a plugin I found in case it helps: https://github.com/blixhavn/sticky-sidebar-v2?tab=readme-ov-file

And a demo: https://blixhavn.github.io/sticky-sidebar-v2/examples/scrollable-element.html

j08lue:
> Ideally, the sidebar would stick at the bottom of the main content when scrolling, it wouldn't scroll up.

+1

hanbyul-here:
I pushed some changes that include
- Sidebar "sticking" to the bottom when scrolling
- Change sticky behavior depending on the height of the sidebar
- Change the sticking point depending on the page header status

`position: sticky` does a lot by default (ex. "sticking" to the bottom when contents are all scrolled up, do not stick when a sticky element is longer than the contents) I think what this PR was missing was to give an actual height value to the sticky element. I got rid of the logic to compare the heights of 'area' and 'sidebar' since it is not needed anymore. I did not clean up further than `data-catalog/filters-control`, `data-catalog/index`

app/scripts/components/common/form/checkable-filter/index.tsx
```diff
@@ -83,10 +84,18 @@ export interface OptionItem {
...
+    title,
+    onChanges,
+    globallySelected,
+    tagItemCleared,
+    calculateHeightCallback
+  } = props;
+  const [show, setShow] = useState<boolean>(true);
   const [count, setCount] = useState<number>(0);
   const [selected, setSelected] = useState<OptionItem[]>([]);
+  const targetRef = React.useRef<HTMLDivElement>(null);
```
hanbyul-here, line 98: Is there any need to introduce the pattern of using `React.useRef`? I am not sure we can set it up as a lint rule, but all other parts of the code explicitly import ` {useRef} `

app/scripts/components/data-catalog/index.tsx
```diff
@@ -223,31 +148,31 @@ function DataCatalog({ datasets }: DataCatalogProps) {
...
     setAllSelectedFilters([]);
   }, [setAllSelectedFilters]);
 
-  React.useEffect(() => {
+  useEffect(() => {
     if (clearedTagItem && (allSelectedFilters.length == prevSelectedFilters.length-1)) {
       onAction(Actions.TAXONOMY_MULTISELECT, { key: clearedTagItem.taxonomy, value: clearedTagItem.id}); 
       setClearedTagItem(undefined);
     }
   }, [allSelectedFilters, clearedTagItem]);
```
hanbyul-here, line 166: I noticed warnings about `useEffect` has missing dependencies throughout this component. Do you mind putting comments on why those dependencies are excluded? It would help me navigate the code better.

app/scripts/components/common/form/checkable-filter/index.tsx
sandrahoang686: not a pattern, can just explicitly import or use like this

app/scripts/components/data-catalog/index.tsx
sandrahoang686: Sure. We have missing dependencies sprinkled throughout the existing codebase already. We can make start making notes going forward why some arn't included

j08lue:
Awesome, the scrolling experience is great now - the filters are visible until the very end of the dataset list and the behavior takes into account the actual height of the side bar. This way, users will always be able to see the filters.

One deviation from [the design](https://www.figma.com/file/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?type=design&node-id=1179-5867&mode=design&t=xzFogX2b2WdP2uPk-0) is that the pills that indicate the active filters and give a shortcut to removing them are added in the right column, above the dataset list:

image

Instead of above the filters:

image

That means that the shortcut to removing individual filters disappears from view quite quickly as the user scrolls down the filtered list. Was this a change you discussed, @faustoperez and @sandrahoang686?

We could keep that refinement to yet another PR, though, and ship this increment already. 👌

sandrahoang686:
@j08lue yes that was a change Fausto and I discussed this during our checkin and in the original PR to have the tags above the dataset list instead of above the filters. Decision change documented [here](https://github.com/NASA-IMPACT/veda-ui/pull/918#issuecomment-2063320262)

j08lue (approved):
As noted in https://github.com/NASA-IMPACT/veda-ui/pull/930#issuecomment-2096624708, the implementation works very nicely!

The small difference from the designs in terms of placement of the filter removal pills is ok for now for me.

Issue #934, opened 2024-05-02
On Exploration make geometries non-movable


j08lue:
It may be more disturbing than helpful that you can (also by accident) easily move predefined geometries like the US country boundary.

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/7fe2f712-bf82-43c4-994d-f2064abd12ac)

This was intended differently in the designs, but was implemented this way for simplicity - all geometries behave the same, whether they come from presets, uploads, or are drawn on the map.

The easiest would probably to disable moving for any geometry and see whether any user ever requests this to be enabled.

## Acceptance criteria

- [ ] Locked all geometries from being moved

PR #936, opened 2024-05-06
Make geometries non-movable in the Exploration feature

dzole0311:
**Related Ticket:** 

Addressing [#934](https://github.com/NASA-IMPACT/veda-ui/issues/934)

### Description of Changes

Two changes made, both affecting only the new map component in the Exploration page:

1. Geometries added from presets are non-draggable & non-editable
2. User-drawn & uploaded geometries are also non-draggable, but their vertices can be moved

- Added a new package in the project to initiate StaticMode (mode with no interactions on polygons): https://github.com/mapbox/mapbox-gl-draw-static-mode
- Override the `dragMove` and the `dragFeature` methods for the `simple_select` and `direct_select`
- We might intend to do more complex interactions in the future that need moving / dragging, so aside from custom modes, an alternative could be some kind of filtering in the original mode itself based on feature props? Ie. if a feature has a prop `isPreset`, then it should be handled differently etc.
- This will probably affect the Data Analysis page as well, once [827](https://github.com/NASA-IMPACT/veda-ui/pull/827) is merged

### Validation / Testing

1. In E&A, select a preset -> it should be `static` so that no vertices are visible nor it can be dragged
2. On the same page, draw a polygon -> the polygons vertices should be draggable, but the polygon itself should be non-movable
3. The same behavior as step 2 applies to uploaded files

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/index.tsx
+++ b/app/scripts/components/common/map/controls/aoi/index.tsx
@@ -1,4 +1,5 @@
 import MapboxDraw from '@mapbox/mapbox-gl-draw';
+import StaticMode from '@mapbox/mapbox-gl-draw-static-mode';
 import { useTheme } from 'styled-components';
 import { useAtomValue } from 'jotai';
 import { useRef } from 'react';
@@ -9,11 +10,23 @@ import { computeDrawStyles } from './style';
 
 type DrawControlProps = MapboxDraw.DrawOptions;
 
+// Overriding the dragMove and dragFeature methods of the 
+// 'simple_select' and the 'direct_select' modes to avoid
+// accidentally dragging the selected or hand-drawn AOIs 
+const customSimpleSelect = {
+  ...MapboxDraw.modes.simple_select,
+  dragMove() {}
+};
+
+const customDirectSelect = {
+  ...MapboxDraw.modes.direct_select,
+  dragFeature() {},
+};
+
 export default function DrawControl(props: DrawControlProps) {
   const theme = useTheme();
   const control = useRef<MapboxDraw>();
   const aoisFeatures = useAtomValue(aoisFeaturesAtom);
-
   const { onUpdate, onDelete, onSelectionChange, onDrawModeChange } = useAois();
 
   useControl<MapboxDraw>(
@@ -21,6 +34,12 @@ export default function DrawControl(props: DrawControlProps) {
       control.current = new MapboxDraw({
         displayControlsDefault: false,
         styles: computeDrawStyles(theme),
+        modes: {
+          ...MapboxDraw.modes,
+          simple_static: StaticMode,
+          simple_select: customSimpleSelect,
+          direct_select: customDirectSelect
+        },
         ...props
       });
       return control.current;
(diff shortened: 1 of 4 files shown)


faustoperez (approved):
I confirm the presets and the custom geometries are working as expected. Great job @dzole0311 🙌

app/scripts/components/common/map/controls/aoi/index.tsx
```diff
@@ -9,18 +10,36 @@ import { computeDrawStyles } from './style';
...
 
   useControl<MapboxDraw>(
     () => {
       control.current = new MapboxDraw({
         displayControlsDefault: false,
         styles: computeDrawStyles(theme),
+        modes: {
+          ...MapboxDraw.modes,
+          simple_static: StaticMode,
+          simple_select: customSimpleSelect,
```
hanbyul-here, line 40: At this point, what is the difference between `simple_static` and `simple_select` mode?

hanbyul-here:
Thanks for a clean and concise solution 👏 I left a question to understand more clearly about when to use `simple_static` and `simple_select`.

app/scripts/components/common/map/controls/aoi/index.tsx
dzole0311: I should have really named this only `static` instead of `simple_static`, but it's basically two different modes. `simple_static` disables all user interactions (including dragging, and prevents vertices from showing up per polygon) because that's the behavior we want to try out for the presets. The `simple_select` mode is almost the same as before, aside from the dragging of the feature as a whole being disabled (but vertices are still shown and can be interacted with).

So `simple_static` is only for the presets at the moment, while the `simple_select` is for uploaded assets or hand-drawn features.

sandrahoang686: https://github.com/mapbox/mapbox-gl-draw-static-mode

sandrahoang686 (approved):
Congrats on your first couple of PRs! lgtm

j08lue:
Got the tiniest of nitpicks - even though user-drawn geometries are not draggable (as intended), the mouse cursor still seems to indicate that it is (four-headed arrow ✥ ish).

Not a blocker, though - please go ahead and merge this as-is!

hanbyul-here:
My apologies that we missed this one bug:  https://github.com/US-GHG-Center/veda-config-ghg/pull/372#pullrequestreview-2041535798

We can find a fix this problem in many ways, but the profoundational problem is that the state of MapboxDraw and React are not syncing (This problem is stated in this ticket: https://github.com/NASA-IMPACT/veda-ui/issues/710#issuecomment-2094324182) - We assume that [any newly added feature is selected and mark it through atom](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/map/controls/aoi/atoms.ts#L36) but the feature added through preset is not selected in MapboxGL Draw. 

Now we have 3 ways of adding AOIs. We should walk-through what is expected for each mean and match the behavior between MapboxGLDraw and React.

j08lue:
> My apologies that we missed this one bug

That is on me for rushing this through. 🙇 

Let us see what the best resolution is - short-term and longer-term.

PR #941, opened 2024-05-07
Set public url as a prefix for preset dasta

hanbyul-here:
Close https://github.com/US-GHG-Center/veda-config-ghg/issues/377
I somehow thought that we were handling baseurl on axios configuration level, but it seems we are not.  

Thanks @slesaad for putting a work-around. - I have a question based on my curiosity - why did preset request work on staging? I think it should have failed since the requests were made directly to staging.earth.gov, not staging.earth.gov/ghgcenter but they worked. Do you have an idea ?

### DIFF
--- a/app/scripts/components/common/map/controls/hooks/use-preset-aoi.tsx
+++ b/app/scripts/components/common/map/controls/hooks/use-preset-aoi.tsx
@@ -4,7 +4,7 @@ import axios from 'axios';
 
 import { getAoiAppropriateFeatures } from './use-custom-aoi';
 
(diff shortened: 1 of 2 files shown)


j08lue:
Re staging works, @hanbyul-here: https://github.com/US-GHG-Center/veda-config-ghg/issues/377#issuecomment-2098574006

hanbyul-here:
Ah I see 😓 Thanks @j08lue and @slesaad

slesaad (approved):

Issue #942, opened 2024-05-07
Delete button doesn't work with static aoi


hanbyul-here:
Issue originally described in this pr thread: https://github.com/US-GHG-Center/veda-config-ghg/pull/372#pullrequestreview-2041535798

👉  An issue with the non-moveable geometries: removing a pre-defined geometry with the "Delete all areas" button fails and leaves the users without a way to get the geometry off:

![image](https://github.com/US-GHG-Center/veda-config-ghg/assets/3404817/53adc3cc-3ae6-4c58-99e1-075d10fb74cd)

A brief description of why this happens in this pr thread: https://github.com/NASA-IMPACT/veda-ui/pull/936#issuecomment-2096926385
👉  The foundational problem is that the state of MapboxDraw and React are not syncing. We assume that [any newly added feature is selected and mark it through atom](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/map/controls/aoi/atoms.ts#L36) but the feature added through preset is not selected in MapboxGL Draw.

PR #945, opened 2024-05-08
Fix delete button not working for static AOI + persist MapboxDraw mode across reloads

dzole0311:
**Related Ticket:**

Addressed [942](https://github.com/NASA-IMPACT/veda-ui/issues/942)

### Description of Changes
In our `aoisUpdateGeometryAtom` we assumed that each feature added on the map via MapboxDraw is selected. This worked well until we started using the `static_mode` for presets where we disable any interactions attached to them. As a result, we introduced a state mismatch between MapboxDraw (the preset features were marked as not selected) and what we held in our atom in React (the preset features were marked as selected).

To fix this in the short-term, two changes were made:

- We switch to simple_select mode if static_mode has been selected when using the trash icon
- A new url param has been added to track whether a present aoi should be editable or is only there for analysis (static). This will track the correct mode across page reloads

Due to the urgency of the fix that should go out by EOW, I opted for the fix in this PR because I wasn't sure if I'll manage to implement / test a long-term solution without breaking something (the approach nicely outlined [here](https://github.com/NASA-IMPACT/veda-ui/issues/710#issuecomment-1829639686). My understanding of a longer-term solution is as follows:

1. Introduce a `drawStateAtom` that keeps track of various draw modes, features, selected features etc.
2. Add a custom hook `useSyncDraw` that reads/writes to the atom and "syncs" the state between React and MapboxDraw
3. Add event handlers in the custom hook that update the atom as a replacement of the current programmatic way

@hanbyul-here @sandrahoang686 If we ticket this for the upcoming sprint/s, I can come up with a draft PR

### Validation / Testing
- Validate that hand-drawn and uploaded polygons can be edited and deleted as usual
- Validate that presets can also be deleted


j08lue (changes requested):
Very nice PR, great you are mindful of the short timeline. 🙏 

## Validation / Testing
1. Validate that hand-drawn and uploaded polygons can be edited and deleted as usual

   - ❌ Neither uploaded nor hand-drawn areas can be moved any longer. While we do not _have_ to offer moving as an option, at least we should change the mouse icon.

   - Btw, uploaded geometries should probably also be immutable (`static_mode`), no?

2. Validate that presets can also be deleted

   - ✅  Deletion now works

3. Review whether there were any unintended side effects

   - ❌ When adding a geometry from preset, I am unable to start an analysis - the "Run Analysis" button is not shown. Probably because the static geometry is not actually "selected"?
   
      image

   - ❌ When I add a geometry from preset and then reload the page, the geometry seems to be treated as editable and the preset selection is gone.

      image

dzole0311:
Okay, this seems trickier than I initially thought.

>    * ❌ Neither uploaded nor hand-drawn areas can be moved any longer. While we do not _have_ to offer moving as an option, at least we should change the mouse icon.

This is the behavior that we intended to do: https://github.com/NASA-IMPACT/veda-ui/issues/934#issuecomment-2092650298, but I agree about the dragging cursor. Will see if I can easily change the styling here.

>    * Btw, uploaded geometries should probably also be immutable (`static_mode`), no?

I generally agree about this, but would need input from @faustoperez 

> When adding a geometry from preset, I am unable to start an analysis
> When I add a geometry from preset and then reload the page, the geometry seems to be treated as editable and the preset selection is gone

😵‍💫 Yes, correct, we don't seem to track the origin of the AOI (hand-drawn / uploaded / preset) anywhere. The solution here would be to keep track of it (maybe as part of a url param) and initialize the correct state for MapboxDraw. We could also use that to make sure that analyses can be ran also for presets.

faustoperez:
> Btw, uploaded geometries should probably also be immutable (static_mode), no?
> I generally agree about this, but would need input from @faustoperez

I also generally agree, my initial assumption was that users won't want to move it from its position but they might want to edit the individual points so they match the map or make final adjustments, as that geometry is generated in a different software. Happy to make it immutable too 👍

hanbyul-here:
Ah, now I can see that the problem is we introduced a new scenario(state) that the geometry can be selected for analysis but not selected for editing. I'd like to point out that there is also a mismatch in how we indicate 'selected' geometry to users. Before, 'selected' geometry (or geometries) was highlighted by editable points. Because being selected meant being editable and analyzed. Now, we are breaking the sync between these three; Therefore we need a way to indicate when polygons are only chosen for analysis.

To clear out my thoughts.

1. Selected by preset 
  - It should be 'internally selected' to be analyzed  * We are missing a unique internal state for this case and a way to indicate this internal selection to users
  - It should not be editable - also, it should not be displayed as editable

2. Selected by drawn AOIs
  - The most recently drawn AOI should be selected both internally and for MapboxGLDraw by default
  - What happens when a user draws multiple polygons and selects some of them? I think the logical answer is the selected polygons are 'internally selected', but not editable.

3. Uploaded file
  - The expectation is the same as that for 1.  

I am okay with a fast solution for now. I wrote this down mainly to clear out my thoughts - and to flag the need to visually indicate the users about 'geometries being chosen only for analysis'.

dzole0311:
Thanks for the overview @hanbyul-here. I made some additional changes and cleaned out the code from my previous commit. I added a url param that will help us to track if a feature is only for analysis or editing. Based on this, we can initiate the initial state of MapboxDraw (either `simple_select` or `static_mode`).

What we should further do:
 
1. We need to do point 2 (second bullet point) from your [comment](https://github.com/NASA-IMPACT/veda-ui/pull/945#issuecomment-2100687700). In the url we keep track of the aois including their ids, so a solution here would be to loop through them when the component is mounted and imperatively "select" them so that they are highlighted. This is only to improve the visual feedback to the user
2. Introduce a different color for static features (like the presets), so they can be differentiated
3. Change the cursor when attempting to drag the immutable features

I also noticed a small bug which was I believe was there from before, but seems like a easy fix: select a preset -> reload the page -> toggle to drawing mode -> the preset is not deleted automatically + it becomes editable.

app/scripts/components/common/map/controls/aoi/index.tsx
```diff
@@ -10,6 +10,8 @@ import { computeDrawStyles } from './style';
 
 type DrawControlProps = MapboxDraw.DrawOptions;
 
+export const STATIC_MODE = 'static_mode';
```
hanbyul-here, line 13: It is probably smart to treat other modes in the same way.

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -168,17 +185,18 @@ function CustomAoI({
     });
     const pids = mbDraw.add(fc);
     setPresetIds(pids);
-    mbDraw.changeMode('simple_static', {
+    mbDraw.changeMode('static_mode', {
```
sandrahoang686, line 188: ⛏️ : Since you created the constant literal string `STATIC_MODE` do you want to replace these hard coded places with that?

app/scripts/components/exploration/atoms/selectedFor.ts
```diff
@@ -0,0 +1,12 @@
...
+const hydrateSelectedForEditing = (value) => value === 'true';
+
+const dehydrateSelectedForEditing = (value) => value ? 'true' : 'false';
+
+export const selectedForEditingAtom = atomWithUrlValueStability({
+  initialValue: hydrateSelectedForEditing(new URLSearchParams(window.location.search).get('selectedForEditing')),
+  urlParam: 'selectedForEditing',
+  hydrate: hydrateSelectedForEditing,
+  dehydrate: dehydrateSelectedForEditing
+});
```
sandrahoang686, line 12: ⛏️ : This is a nit but can we change the file name to `selectedForEditing` or something more information to what this file is for?

dzole0311:
Hey all, I created a [follow-up issue ](https://github.com/NASA-IMPACT/veda-ui/issues/950) for some of the findings here because that's a behavior we already had before the static mode got introduced and I would like to minimize the code changes in this PR. The rest should work fine, please let me know of any issues 🙇

app/scripts/components/exploration/atoms/selectedForEditing.ts
```diff
@@ -0,0 +1,12 @@
+import { atomWithUrlValueStability } from '$utils/params-location-atom/atom-with-url-value-stability';
```
hanbyul-here, line 1: Doesn't it make more sense to keep this atom with other aoi-related atoms (in `common/map/controls/aoi/atoms.ts`)?

dzole0311: Makes sense, moved 👍

j08lue (approved):
1. Validate that hand-drawn and uploaded polygons can be edited and deleted as usual

   - ✅ Behaviour changed as expected - no more moving of geometries
   - ✅ Uploaded geometries behave like presets - not editable
   - ✅ Mouse cursor is consistent with behaviour (panning map even on hover over geometry)

2. Validate that presets can also be deleted

   - ✅ Deletion works, both via "Clear preset" on drop-down and "Delete all areas"

4. Review whether there were any unintended side effects

   - ✅ Preset or uploaded geometries are still not editable after reload

hanbyul-here (approved):
🙇 LGTM

Issue #932, opened 2024-05-02
[Design] Bring new data catalog layout also to the Exploration layer selection


j08lue:
We recently implemented a new data catalog layout with rows and a (floating) filter sidebar:

- [#918 ](https://github.com/US-GHG-Center/ghgc-architecture/issues/133#issuecomment-2009192700)

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/d0a2811e-3781-40a9-8a73-15af9ced5194)

The same layout would also help in the Exploration layer selection, so that users can better browse the available layers:

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/41b834dc-d92b-44bc-8e16-bf41e9876c7c)

It will probably need some adaptation?

## Acceptance criteria

- [ ] Adapt row/sidebar data catalog layout for Exploration layer selection

faustoperez:
Self-assigning this 👍

j08lue:
Once the design change is implemented, we can also close these two issues:

- https://github.com/NASA-IMPACT/veda-config/issues/355
- https://github.com/US-GHG-Center/veda-config-ghg/issues/59

faustoperez:
Designs ready for revision on Figma 👉 https://www.figma.com/design/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?node-id=1593%3A4526&t=awwUsQrQgc9qB1Fr-1

![Desktop Flow](https://github.com/NASA-IMPACT/veda-ui/assets/12268629/2bc40dce-484c-4c7c-87a1-b2f95c6f7185)

PR #946, opened 2024-05-08
Data Catalog Feature Component Breakout

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/892

Data Catalog now takes in datasets as a prop and to support the current architecture, a wrapper container was created to interact with the data-layer to provide data to this feature component ([PR#893](https://github.com/NASA-IMPACT/veda-ui/pull/893)). Note: The template instance wouldn't need this wrapper container as they would have their own page views that would interact with the data layer. Page layout / styling is also removed from Data-catalog component into the wrapper ([PR#930](https://github.com/NASA-IMPACT/veda-ui/pull/930/files)).

@j08lue @aboydnw  @faustoperez The data-catalog feature component is now just what is highlighted in the red box. So when this feature is imported, they would only get what i've highlighted here. Page layout and headers would be something the actual page manages now. Does this seem fine to you both?

Screenshot 2024-05-09 at 3 14 00 PM

### Description of Changes
* Allow card override
* Clean up unused / dead code

### Validation / Testing


app/scripts/components/data-catalog/container.tsx
```diff
@@ -17,14 +17,8 @@ import { FeaturedDatasets } from '$components/common/featured-slider-section';
...
-      <LayoutProps
-        title='Data Catalog'
-        description={getString('dataCatalogBanner').other}
-      />
-      <PageHero
-        title='Data Catalog'
-        description={getString('dataCatalogBanner').other}
-      />
+      <LayoutProps title='Data Catalog' description={getString('dataCatalogBanner').other} />
+      <PageHero title='Data Catalog' description={getString('dataCatalogBanner').other} />
```
sandrahoang686, line 21: 📓 : Moving layout component from data-catalog into wrapper instead

app/scripts/components/data-catalog/index.tsx
```diff
@@ -103,9 +94,16 @@ export const sortOptions = [{ id: 'name', name: 'Name' }];
 
 export interface DataCatalogProps {
   datasets: DatasetData[];
+  cardComponentOverride?: { // Allow user to override with a different cardType with optional props
+    CardComponent: ComponentType<CardComponentProps>;
+    props: Partial<CardComponentProps>;
+  }
```
sandrahoang686, line 100: 📓 : Do we care to have the ability to override the card component? I've added this but i'm thinking we dont currently need it. If we do want to override card logic or styles, we can defer till we have that usecase and need to solve for. thoughts?

faustoperez:
There's work planned to bring the data catalog layout to the layer selection modal on the E&A page.

The difference is that on the Data Catalog we're searching data collections and on the E&A page we're searching data layers within collections.

This is the ticket for that 👉 https://github.com/NASA-IMPACT/veda-ui/issues/932

sandrahoang686:
Note: @hanbyul-here @dzole0311 I created the `content-views` directory and moved the data-catalog logic under there and then renamed it to `catalog-view`. The `data-catalog/` directory just holds the wrapper container component to create the data-catalog page now. This is to align with what we talked about for creating different views a user can import to render some content list

app/scripts/components/data-catalog/index.tsx
hanbyul-here: I think we discussed this issue before, but to make it as a record: I don't think we need to offer ways of overriding specific components inside of the feature components (at lest for now.) If this is needed, a user can make their own list without overriding the specific components inside of the feature.

hanbyul-here:
Some of my thoughts on the bigger scheme of changes!

Is the plan to put everything under the `content-view` folder? (For example, will there be `content-view/exploration` and `content-view/stories`?) Or is this only because you were considering consolidating StoriesHub and Data Catalog? If it is a previous case, I have some concerns that there will be too many codes that we have to move. I am not sure if this fully represents the components under this folder.

I thought the Data catalog component would export the whole page (without navigation - blue rectangle in the screenshot), but it will have three components inside that users can opt-in to use separately (red rectangles) if they don't need a whole page as it is. I am mainly thinking in this way for an easy integration of the existing instances later. 🤔 How do you think/

Screen Shot 2024-05-13 at 4 38 09 PM

app/scripts/components/data-catalog/index.tsx
sandrahoang686: I think this makes sense, if the user wants alot more flexibility or to use whatever card component they like, they will have the flexibility now to do this in their page views. i'll go ahead and remove

sandrahoang686:
@hanbyul-here thanks reviewing and asking
> Is the plan to put everything under the content-view folder?...

No, I created the `content-view/` dir to just hold "views/looks" for displaying some set of cards given a list of content types, so the usecase would be for the data catalog page and stories hub/insights page. There would be two views, one that looks like the new data catalog view with the horizontal cards and floating side bar and then the other view would be the older data catalog view or current stories hub view with the classic cards and top filters. This way the user can explicity import the view they would like to use to render some list of contents.

> I thought the Data catalog component would export the whole page...

With feature components, we should just isolate them to the actual product feature they are trying to deliver. So instead of thinking of entire pages to export and import, its more like we dont worry about layout and layout components like headers and footers. Those can be reuseable components we breakout though and the instance gets to dictate whether or not they'd want to use those layout components and the layout in general (for example, the reuseable headers we've broken out or just entirely different ones). This is where we give alot more control in terms of composability to the instance. No more override components and passing in JSX elements as props to the page. Instead the page is built on the instance side with layout and layout components controlled at that level. (we will still provide a set of reuseable layout components though like the existing header and footers and other reuseable non layout components like the datasets slider component)

In regards to easy integration for existing instances later, It is still very easy, as you can see in the [container component](https://github.com/NASA-IMPACT/veda-ui/pull/946/files#diff-d8d2a213812e43fcf52661aedb8a2cde3e50898398b36f2c8f723dd79a6c0292) created. It easily recreates the page but gives so much more flexibility in terms of layout control. (this container element is supposed to replicate how it would look like on the instance side and was created here to just support our current instances until they are migrated over) 

Does this make sense?

hanbyul-here:
It makes total sense, @sandrahoang686 🙇 thanks for your reply. My apologies for revisiting some items that we already discussed, such as layouts from VEDA-UI. My thoughts go back and forth as I start working on the test instance.

Thanks for pointing out the container component. I think the migration should be okay as long as we export the existing layout components from VEDA-UI. 

I am still a bit concerned that the global style from the instances interrupts how the components can be composited—for example, can our catalog-view component accommodate the dark mode set by the instance? But the way that I suggested won't necessarily resolve this problem now I think. 


>  the case would be for the data catalog page and stories hub/insights page

If this component can be reused inside of Veda-ui, how about moving it to `common/catalog` along with other reused components?

sandrahoang686:
@hanbyul-here as per our discussion, i've moved the catalog view under common as `/common/catalog/`. I will think about whats best for the stories hub view but defer that until i work on that issue 🙇🏼

app/scripts/components/data-catalog/container.tsx
```diff
@@ -17,16 +17,10 @@ import { FeaturedDatasets } from '$components/common/featured-slider-section';
...
     <PageMainContent>
-      <LayoutProps
-        title='Data Catalog'
-        description={getString('dataCatalogBanner').other}
-      />
-      <PageHero
-        title='Data Catalog'
-        description={getString('dataCatalogBanner').other}
-      />
+      <LayoutProps title='Data Catalog' description={getString('dataCatalogBanner').other} />
```
dzole0311, line 20: Nit: Should we make the titles here more dynamic (ie. to arrive as props from the DataCatalogContainer?)

app/scripts/components/data-catalog/container.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { getString } from 'veda';
 import { allDatasets } from '$components/exploration/data-utils';
```
dzole0311, line 3: Nit: To avoid coupling features, should we create a `common/data-utils.ts` to move the `allDatasets` util there and use it in both, the exploration and the data-catalog?

dzole0311 (commented):
The last changes and the discussion points make sense to me 👍 Aside from two nits, the rest lgtm, thanks @sandrahoang686

app/scripts/components/data-catalog/container.tsx
hanbyul-here: Ah this is a great point - we would need this as a prop once we start exporting layout components.

sandrahoang686: Discussed! This is layout related which will eventually be dictated at the page level, the page level can decide which layout component they would use and the title to pass in, this Layout Component already takes in title as a prop as well and title passed in could be made to be dynamic on the instance side. Given that this container is temporary to just support current instances, this shouldn't matter

hanbyul-here (approved):
Thanks for working on this feature breakout 🙇

PR #962, opened 2024-05-21
Map core feature breakout

dzole0311:
**Related Ticket:** [899](https://github.com/NASA-IMPACT/veda-ui/issues/899)

### Description of Changes

This PR breaks out the Exploration feature as a core feature of the veda UI that consists of two components: `ExplorationMap` and the `Timeline`.

The changes include the following:

1. Created a `container.tsx` that wraps the `ExplorationAndAnalysis` feature and passes the `datasets` as a prop. The `setDatasets` is also used when adding new datasets or updating existing ones from the children
2. Lifted up some of the jotai state from the children (`ExplorationMap` and the `Timeline`), such as the `selectedDay`, `selectedCompareDay` etc and passed it down as props to the children
3. Some smaller clean ups based on the discussion points throughout the PR

I've also added examples of the ExplorationMap and the MapBlock in the sandbox page: https://deploy-preview-962--veda-ui.netlify.app/sandbox

### Validation / Testing
1. Validate the exploration map and that it works as before without regressions
4. Validate the stories where the MapBlock was used and that it works as before without regressions


app/scripts/components/exploration/index.tsx
```diff
@@ -69,11 +75,91 @@ const tourProviderStyles = {
...
+    setBasemapId,
+    labelsOption,
+    boundariesOption,
+    onOptionChange
+  } = useBasemap();
+
   const datasets = useAtomValue(timelineDatasetsAtom);
+  const setDatasets = useSetAtom(timelineDatasetsAtom);
+
+  useReconcileWithStacMetadata(datasets, setDatasets);
```
hanbyul-here, line 91: From this line to line 158, some of the features can't be overridden (or can't be optional) to use Exploration properly. ex. `onStyleUpdate`  attaches TileURLs as metadata to datasets. If this function gets overridden by a user-passed function, `LoadIntoGIS` functionality from the timeline can't work anymore. (Attaching a screenshot of LoadIntoGIS functionality). Same with reconciling StacMetadata. In my mind,  this is one of the crucial features of ExplorationMap that can be done out-of-box.

Screen Shot 2024-05-21 at 2 57 35 PM

Now that I can see more clearly that Exploration is a tricky component. There are global states that exist on this page that only make sense inside of the container together (Timeline and Map), so I am not sure if each component can be used outside of its wrapper. How do you imagine users will construct the exploration page? (I think it would be ok if we ask users to use exploration as it is like`<Exploration datasets={datasets}>` at least for the short-term future? but I might be missing something - !  lmk if you want to discuss further etc. )

sandrahoang686 (commented):
Thanks for taking the initial stab at this 🙇🏼 ! Left some thoughts, comments, and things to re-consider

app/scripts/components/exploration/index.tsx
dzole0311: @hanbyul-here @sandrahoang686 Thanks, lots of good pointers here!

> How do you imagine users will construct the exploration page?

Not all is done in the PR, but I imagined few things on how the Exploration page could evolve:

1. Allow more functionality overriding
2. Allow hiding / showing of certain controls and elements within the ExplorationMap and the Timeline (not necessarily done in this PR)

Reading through the PR comments, point 1 above seems that is easier said than done, and I don't have a good answer here. 

> I think it would be ok if we ask users to use exploration as it is like<Exploration datasets={datasets}> at least for the short-term future

It seems like this is the way to go, or until the feature prompts for more flexibility. But I can't imagine a scenario where I would like to only use the timeline widget on a page, or where I would only like to use the Exploration map on a page. They really seem like one logical core feature, both in design and functionality.

sandrahoang686: I think it makes sense to couple both the map and timeline as a core feature as the way forward. Lets not worry too much about functionality overriding till we meet those usecases requiring it, we can build to be flexible though where we see current usecases may need it 👍🏼

hanbyul-here: An additional two cents here is we might want to see the composability of more broken-down components, not the overridability (is this a word) of already composed components. In my mind, Exploration map is already a composed component that needs to serve a specific purpose; I think the goal of this component doesn't have to include the flexibility of how it can be used by users. (To find an example, disabling AOI control doesn't make sense in the Exploration map because it is directly related to the analysis feature)

app/scripts/components/exploration/components/map/index.tsx
```diff
@@ -21,60 +18,37 @@ import {
...
+  selectedCompareDay: Date | null;
+  mapBasemapId: BasemapId;
+  setBasemapId: (id: BasemapId) => void;
+  labelsOption: boolean;
+  boundariesOption: boolean;
+  onOptionChange: (option: string, value: boolean) => void;
+  onStyleUpdate: (style: any) => void;
+  projection: ProjectionOptions;
+  setProjection: (projection: ProjectionOptions) => void;
+}
```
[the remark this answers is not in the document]
dzole0311: Typing this might be obsolete now because I kept some of these (ie. the `projection`) within the component (as it was before) instead of passing it as a prop since we're setting the projection based on the passed datasets to the component. Let me know if I'm missing something by not lifting that up to the parent.

app/scripts/components/exploration/index.tsx
dzole0311: Okay, I think all of the comments should be addressed now

app/scripts/components/sandbox/exploration-map/index.js
```diff
@@ -0,0 +1,114 @@
...
+  height: 40rem;
+`;
+
+const Wrapper = styled.div`
+  position: relative;
+  grid-column: 1 / -1;
+  height: 42rem;
+`;
+
+const mockDatasets = [
```
hanbyul-here, line 17: I'm not sure if this is a nitpick or not. But any reason that you are passing the incomplete dataset here? (This data does not seem to be enough to render the actual layer.)

dzole0311: Thanks for catching this! The sandbox should load the raster layers correctly now.

hanbyul-here (approved):

Issue #947, opened 2024-05-09
[VEDA2 Refactor] Build NextJs Instance Prototype


sandrahoang686:
**Description:**
Build a NextJs Instance prototype to see how it would parse and transform static content and then consume the veda-ui library. 

It would be a good idea to just build this prototype with for a single page / route. The **_Data Catalog page_** is a good option as it looks at our static content and relies on this dataset as a prop and is already being broken out.

This should just be a _sandbox prototype_ and not the official template instance. 

**Dependencies / Blockers:**
* [Build Pipeline Implementation](https://github.com/NASA-IMPACT/veda-ui/issues/800)
* Core feature breakouts (doesn't have to be all, can import these along the way once they are broken out and ready)

**Questions to answer:**
* How will we support parsing of JSX in the mdx files?
* Is any interference from the parcel resolver needed to support how our current content?
* Is nextJs ideal?

hanbyul-here:
I briefly played with NextJS's default blog template, which supports JSX, could think of more broken-down steps as below. Each item can be better defined but am leaving notes for now as I wrap my day.

- Update Node : We might not need to update Node if we mark the Node version more flexibly (like >=16) but I think it is time to update Node anyway as 16 reaches its end of life. We would need to update instances' Nodes versions as well.
- Parse frontmatter in a way that VEDA UI expects (This is what Parcel resolver does now. I believe a lot of snippets from the resolver directly used)
- Make a linkage between VEDA-UI and the test instance: As https://github.com/NASA-IMPACT/veda-ui/issues/799 is not resolved but this test instance needs VEDA UI as a dependency, the instance needs a way of using VEDA-UI. I think the fast way is just to use GitHub branch (and use `yarn link` if any local dev is needed) until #799 is resolved.

hanbyul-here:
This is the repo that I set up: https://github.com/developmentseed/next-veda-ui
I copied one dataset.mdx and one stories.mdx to the test repo, to see if we can use them in NextJS environment.

hanbyul-here:
Notes on findings:
1. There are some MDX components that make sense for users to opt out. 2x.Next offers some components such as Image (can be server side-renered), Link component. 
2. Parcel-specific file urls (`new URL(...).href`) should be replaced
3. Components are not getting exported as expected: We strictly check what the children components are for `Block`, allowing only specific combinations so the layout can be styled as expected. We do this by checking displayName of the components. displayName

hanbyul-here:
[...]

PR #917, opened 2024-04-10
Update Node Version

sandrahoang686:
I (@hanbyul-here ) took this PR over from @sandrahoang686 

**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/947

### Description of Changes
Update Node to v20.

### Notes & Questions About Changes

I will sit this PR until the currently active PR is getting merged, (I can see that Netlify will act out changing the Node versions between previews)

### Validation / Testing
There should not be any noticable changes.

PR #971, opened 2024-05-22
Make Node version more flexible

hanbyul-here:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/947 follow up of https://github.com/NASA-IMPACT/veda-ui/pull/917

### Description of Changes
I realized that it is better to make the Node version flexible in package.json, so VEDA-UI users don't have to be on exact Node v20. 

I think v18 is a reasonable choice to be the minimum version number since 1. We don't have a package that requires specific Node version and 2. 18 is the actively maintained Node at this point (https://nodejs.org/en/about/previous-releases) 🤔 Is there any more thing to consider?

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test": "jest"
   },
   "engines": {
-    "node": "20.x.x"
+    "node": ">=18.0.0"
   },
   "devDependencies": {
     "@babel/core": "^7.16.0",


dzole0311 (approved):

dzole0311 (approved):
I think it makes sense to be flexible, but also we should note that we for example deploy our instances to prod using a [specific version](https://github.com/NASA-IMPACT/veda-config/blob/develop/.github/workflows/deploy-prod.yml#L15). If someone decides to use the latest experimental Node version and that causes issues in their instances, we might not catch it during the deployment / testing stage until it's reported as an issue by them 🤔

hanbyul-here:
We talked during stand-up, but for the record: The major version of UI needs to go up with the Node version change. (Meaning that all the instances need to update Node version to use the latest version of UI) And we will pin the Node version for instances (to v20.)

Issue #800, opened 2024-01-17
[VEDA2 Refactor] Implementation - Create layer / build pipeline to access core ui library 


sandrahoang686:
**Description:**
As part of the refactor work for veda-ui v2, we need to figure out how to access components from the instance level to consume as a library. This will likely require us to wrap the current ui logic with a layer to easily access components by importing through npm registry (to be determined).

**This ticket is the implementation ticket of the findings from [#799](https://github.com/NASA-IMPACT/veda-ui/issues/799)**

**Acceptance Criteria:**

- [ ] Cleanly wrap the current veda-ui code with a layer to be able to easily access components as a library
- [ ] We should be able to easily import components at the instance level for consumption
- [ ] It shouldn't have to worry about the loader dependencies which is how the current veda-ui is rendered

for example... at the instance level, we can regularly import any component that has been broken out in veda-ui library

```
import { Card } from 'veda-ui/components/card';

function Content() {
   return (
      <Card/>
   )
}
```

**Resources:**
https://miro.com/app/board/uXjVN6lkBnc=/

hanbyul-here:
A very rudimentary level of the experiment is going on the `test-dependency` branch now. (Also, on this codepen which uses this branch as a dependency: https://codesandbox.io/p/sandbox/using-veda-ui-nsz24n?file=%2Fsrc%2FApp.js)

aboydnw:
@hanbyul-here @sandrahoang686 this seems like a pretty critical step. I'm wondering if there are any dependencies that must come first before we can do this?

PR #976, opened 2024-05-27
Add library build configuration

oliverroick:
Related ticket: #800 

**Work in progress:** This PR contains the changes necessary to publish a library version of veda-ui. The current state is published to DevSeed’s private registry and can be installed using `npm install @developmentseed/veda-ui`.

## Changes implemented so far

- Added the build command (`yarn buildlib`) which you can use to build the library. The outputs are written to the `lib` directory. 
- Added a parcel target `veda-app` to configure a different entry point for the application build. 
- Added `app/scripts/index.ts` as the entry point for the library. It exports all components we want to make available through the library. 
- Added `.npmignore` to reduce the size of the packaged library. 

**Remove dependencies on faux `veda` module**

The main issue we faced was the faux `veda` module that is created when running the application build. Because the `veda` is directly imported in many components, the library build depends in the veda module too. As a result the library build also includes the mock contents and in turn the library depends on the content. 

To resolve the dependency on the faux `veda` module, we needed to replace all imports from the `veda` module, resulting in the following changes:

- Added `app/scripts/types.d.ts`, which contains the types exported by the `veda` module. 
- Changed relevant imports of types to the new types declaration 
- Refactored the `layout-root` component, so the veda module is only imported for the components that need it. 
	- Separated hooks `useFeedbackModal.ts` and `useSlidingStickyHeaderProps.ts` into their own files. 
	- Separated the LayoutContext into its own file
- *Temporarily* removed the use of `getBoolean` in `app/scripts/utils/url.ts`. I couldn’t find a way to resolve that dependency.


hanbyul-here:
Thanks for working on it. I am trying to use one component from the library. But it returns undefined when I try to use import only one component. ex. https://github.com/developmentseed/next-veda-ui/blob/main/app/components/mdx-components/block.tsx#L3-L4  

```
import VEDA from "@developmentseed/veda-ui"; 
console.log(VEDA.Block) // This works
```
But
```
import { Block } from "@developmentseed/veda-ui"; // This returns undefined
```
Do you have an idea why this happens?

hanbyul-here:
It looks like it was the problem of missing a named export. I added it, along with some other changes I needed (Export of Theme provider)

hanbyul-here:
Since this branch is outdated, I will merge main branch to this branch. 

Flagging that some of the efforts on catalog components: https://github.com/NASA-IMPACT/veda-ui/pull/976/files#diff-ae1c3e259f1b00b01daf0efa008fb2259441e4a5cce04b3dbf4573d0573e58f2R89 will be reverted since there were some changes introduced to the file - we will do the work to make data catalog free from faux module soon though.

oliverroick:
I’ve found the culprit for this issue after doing some spelunking, here’s what I found. Buckle up!

Ts-check only fails for components that use `Modal` of the UI library. Leaving out the majority of properties for brevity, the `ModalProps` type is defined as:

```
type Modal = JSX.IntrinsicElements["section"] & {
  content?: React.ReactNode;
}
```

`JSX.IntrinsicElements["section”]` is defined as `React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>` with the `HTMLAttributes` interface defining a content property with `content?: string | undefined`. 

Now there’s a conflict in the type intersection, a `string` can be a `ReactNode` and a `ReactNode` can be `string` but also `Element`, `boolean`, `number` and many more, so there isn’t sufficient overlap between the two types. 

You can verify this using this simple example:

```
type Modal = JSX.IntrinsicElements["section"] & {
  content?: React.ReactNode;
}

const stringContent: Modal = {
  content: "blah"
};

const nodeContent: Modal = {
  content: <p>Blah</p> // <-- This throws an error because it's not a string
}
```

I’m assuming the behaviour is correct and the `Modal` type in the UI library should be fixed. (I think the correct solution would be to rename the `content` property)

Now the question is why does this surface now? I’ve done a couple of experiments and found:

- tscheck passes on `main`.
- tscheck passes on `publish-lib` with the dependencies from `main` installed.
- tscheck fails on `publish-lib` with the dependencies from `publish-lib` installed.

It’s seems to be down to the dependencies my branch installs. I narrowed it down to the `@types/react` package. To make the library build work, I had to make some changes how we use the React types. We’re using several libraries that depend on different, incompatible versions of `@types/react`. Styled components uses the React 17 types, for example, while we’re explicitly using the React 18 types. This led to issues when building the library and type definitions, which I’ve worked around by forcing the `@types/react` package to version 18.2.12, the most recent release at the time. To pin the version I added the following to `package.json`, which means only this particular version of `@types/react` will be installed. 

```
"resolutions": {
  "@types/react": "18.2.12"
}
```

We have used `@types/react` at 18.2.5 before, my hunch is that the library that is causing the error now was on an older version of `@types/react`, which is why we weren’t seeing the error. 

The latest version of `@types/react` that works for us is 18.0.32. Everything from 18.0.33 breaks the TS checks. So I’m pinning the `@types/react` to 18.0.32. I don’t know what has changed between the two versions, I can’t find any changelog or release notes for the package.

hanbyul-here:
@oliverroick Thanks for the investigation 🔍 ! 

It does look like `@types/react` added `content: string` attribute from `18.0.33`
https://npmdiff.dev/@types%2freact/18.0.33/18.0.32/react/index.d.ts ! 

Honestly, I haven't paid attention to the version of the type that the UI depends on. I can't think of an immediate problem with pinning the `@types/react` for now, but it probably is not an ideal solution for the project's compability and flexibility. 

I can go ahead and put a measure below on ui-library. So we don't have to pin down the dependency so precisely. (I am trying not to change how the library is used since the package is deprecated at this point. - but I think we can still do the release for maintenance? )

```
type Modal = Omit<JSX.IntrinsicElements["section"], "content"> & {
  content?: React.ReactNode;
};
```

app/scripts/components/common/layout-root/useFeedbackModal.ts
```diff
@@ -0,0 +1,22 @@
+import { useContext, useCallback } from "react";
+import { LayoutRootContext } from "./context";
+
+/**
+ * Hook to access the feedback modal.
+ */
+export function useFeedbackModal() {
```
hanbyul-here, line 7: Flagging that GHG instance is using this hook https://github.com/US-GHG-Center/veda-config-ghg/blob/develop/overrides/development/component.tsx#L9 , expecting it to be in `layout-root`

sandrahoang686:
I'm comfortable pinning down the types version for now and going on to merge this. Then we can revisit how we type the modal later. Let me work through the conflicts

sandrahoang686 (approved):
lgtm 

ghg preview with this branch generated [here](https://github.com/US-GHG-Center/veda-config-ghg/pull/425) - QA'ed and looks good to go

hanbyul-here (approved):
I have some clean-up items I would like to tackle as a follow-up such as
 - Consolidate `veda-data` and `veda-data-no-faux-module`
 - Remove commented out imports and unused variables (in Gulp related file) 

But we would like to merge this branch in so we don't have to spend times just to make the branch up to date. so let's merge this and do ff.

@oliverroick 🙇 Thank you so much for working on this branch and lending your help to VEDA team.

Issue #973, opened 2024-05-24
[Dev] Bring new data catalog layout also to the Exploration layer selection


dzole0311:
We recently implemented a new data catalog layout with rows and a (floating) filter sidebar. The same layout would also help in the Exploration layer selection, so that users can better browse the available layers.

Designs: https://www.figma.com/design/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?node-id=1593%3A4526&t=awwUsQrQgc9qB1Fr-1

Related design ticket: [932](https://github.com/NASA-IMPACT/veda-ui/issues/932)

dzole0311:
As [commented here](https://github.com/NASA-IMPACT/veda-ui/issues/932#issuecomment-2100106697), we can close the two issues as well once this is merged:

- https://github.com/NASA-IMPACT/veda-config/issues/355
- https://github.com/US-GHG-Center/veda-config-ghg/issues/59

PR #988, opened 2024-05-31
WIP - Use new data type (+ simplify block map)

hanbyul-here:
**Related Ticket:** #969, #967

### Description of Changes

Opening this to share the current progress of work.

PR #989, opened 2024-06-03
Port the Catalog View in the E&A dataset selector modal

dzole0311:
**Related Ticket:** [973](https://github.com/NASA-IMPACT/veda-ui/issues/973)

### Description of Changes
Ported the CatalogView to the E&A modal.

1. Created a configurable Catalog Card, that either can be selected (E&A Modal) or can navigate to a link (Data Catalog)
2. Re-factored the Data Catalog component to a reusable Catalog Content component (used in Data Catalog and E&A Modal)
3. Few other tweaks and changes + addressed Fausto's design-related comment below

### Validation / Testing
1. Validate that the Catalog works with navigation in the Data Catalog view (counters are correct, filters work, navigation works)
2. Validate that the Catalog is shown in the Dataset Selector modal in E&A and works as expected (is selectable, can add datasets, counters are correct, filters work)


faustoperez (changes requested):
Great job Gjore! Two comments only:

- I think it is safe to have the filters collapsed by default so they don't take a lot of space.
- Make the whole filter box label clickable to open/close and not just the chevron 

Thanks!

app/scripts/components/common/catalog/controls/atoms/taxonomy-atom.ts
```diff
@@ -0,0 +1,19 @@
+import { atomWithUrlValueStability } from '$utils/params-location-atom/atom-with-url-value-stability';
+
+export const taxonomyAtom = atomWithUrlValueStability<Record<string, string | string[]>, Record<string, string | string[]>>({
+  initialValue: {},
+  urlParam: 'taxonomy',
```
dzole0311, line 5: I created a `taxonomy` and `search` atoms as the `useBrowseControls` did not seem to consistently sync the url param after clearing all the taxonomy filters.

For example, I would select few taxonomies, select a dataset, add it to a map. Then re-open the modal, deselect the prev selected taxonomies, select another dataset and add it to a map. Reopen the modal, the taxonomies are persisted because they pop up again in the `url`.

This might have to do with the fact that in E&A we sync query params from two places, once using jotai atoms (ie. `datasets`, `aois`) and once now using the `useBrowseControls` which uses a custom url string management lib. 

To fix the issue I came up with these two atoms and a custom hook which updates / reads from them (similar to what we already have for the rest of the E&A url params).

Let me know what do you think about this.

hanbyul-here:
🤔 I am not sure I am complicating the matter or not - but I am unsure if we need URLs to be manipulated when users interact through modal. Especially dinvr the query parameters are not getting reset when the datasets are seleted. 

In a scenario like below 
1. A user selected some datasets through selecting specific taxonomies
2. The user navigates datasets, finds the cool thing, so shares the url
3. Another user opens the shared url, then opens the modal to add more datasets 

In Step 3, that user will see the taxonomies that were not chosen by the user. 

The current behavior makes sense in a way that the same component does the same URL manipulation. On the other hand, I think mixing up URL parameters from two components (data catalog modal and e&a components) for one page (new E&A) creates an edge case like above. What do you think? Am I missing anything?

app/scripts/components/common/catalog/controls/atoms/taxonomy-atom.ts
hanbyul-here: Adding a few more problems that I found

- Atom values perseist through pages ex. users interact with the modal on the E&A page and navigate to the catalog page. Search and Taxonomy values that were selected in E&A page pop up. I don't think this is the expected behavior
- Something is weird when there is more than one selection happens. The interaction from the third interaction doesn't seem to work at all. ex. in the video below, I selected 1. Topics>Air Quality, 2. Gas>Co2, then unselected Gas>Co2, but the result doesn't show the result from step 1.

https://github.com/NASA-IMPACT/veda-ui/assets/4583806/7c435e46-f9c9-440f-85e5-3621db8c2f73

dzole0311: Thanks for the QA and the insightful comment! All should be addressed now. 

> I am unsure if we need URLs to be manipulated when users interact through modal

I agree (I was trying to recreate the same behavior from the legacy modal). I went ahead and stopped persisting the taxonomies in the URL, but they will persist in a local state if I close and reopen the modal until i refresh the page as I think it's nice to have the immediate previous selection I've made available to me as a user. But it's an easy tweak for me to clear this state out if you think otherwise!
 
 > Atom values perseist through pages ex. users interact with the modal on the E&A page and navigate to the catalog page
> Maybe we can use the hook only for Data Catalog page?
 
Fixed. We'll use the hook in the Data Catalog page and a local state in the Modal, and pass the taxonomies down to the Catalog Content (remember the point from yesterday when we discussed how we should move the taxonomies up one level? 😃)
 
 > Something is weird when there is more than one selection happens
 
 Should be fixed

app/scripts/components/common/form/checkable-filter/index.tsx
```diff
@@ -90,7 +91,7 @@ export default function CheckableFilters(props: CheckableFiltersProps) {
     globallySelected,
     tagItemCleared
   } = props;
-  const [show, setShow] = useState<boolean>(true);
+  const [show, setShow] = useState<boolean>(false);
```
hanbyul-here, line 94: If I remember correctly, we decided that the filter menu should be open by default.

dzole0311: That change is coming from @faustoperez above: https://github.com/NASA-IMPACT/veda-ui/pull/989#pullrequestreview-2093840574

app/scripts/components/common/form/checkable-filter/index.tsx
```diff
@@ -90,7 +91,7 @@ export default function CheckableFilters(props: CheckableFiltersProps) {
...
   const [count, setCount] = useState<number>(0);
```
hanbyul-here, line 95: (Not introduced by PR) - `count` and `selected` are not syncing with the state correctly. Ex. On this page: https://deploy-preview-989--veda-ui.netlify.app/data-catalog?search=&taxonomy=%7B%22Topics%22%3A%5B%22air-quality%22%2C%22experimental%22%2C%22agriculture%22%2C%22our-planet%22%2C%22climate%22%5D%7D if I select one more item from Topics filter, I get a message that only 1 topic is selected.

Screen Shot 2024-06-05 at 11 47 08 AM

app/scripts/components/common/browse-controls/use-browse-controls.ts
```diff
@@ -8,11 +8,10 @@ export enum Actions {
   SEARCH = 'search',
```
hanbyul-here, line 8: Is this hook only going to be used for Stories Hub? (If so, can we put a comment about it?)

app/scripts/components/common/catalog/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+            : [taxonomies[key]];
+
+        if (taxonomyGroupValues.includes(val)) {
+          const updatedValues = taxonomyGroupValues.filter((x) => x !== val);
+          updatedValues.length
+            ? setTaxonomies(set({ ...taxonomies }, key, updatedValues))
+            : setTaxonomies(omit(taxonomies, key));
+        } else {
+          taxonomyGroupValues.push(val);
+          setTaxonomies(taxonomies);
```
hanbyul-here, line 44: Any reason not following the pattern for  `setTaxonomies(set({ ...taxonomies }, key, updatedValues))` this case? I think it is generally a better practice to avoid mutating the object.

app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
```diff
@@ -208,21 +109,30 @@ export function DatasetSelectorModal(props: DatasetSelectorModalProps) {
...
+        <RenderModalHeader />
       )}
       content={
-        <ModalContentRender 
-          search={search} 
-          selectedIds={selectedIds} 
-          displayDatasets={datasetsToDisplay} 
-          onCheck={onCheck}
-        /> 
+        <CatalogContent
```
hanbyul-here, line 115: 😎 that we can pass the Catalog Content directly here. Can we adjust the styles so the filter component and selected pills component can align? (To be fair, they don't perfectly align on Catalog page either but it looks more obvious in a smaller modal)

Screen Shot 2024-06-05 at 10 23 09 AM

app/scripts/components/common/catalog/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  TAXONOMY_MULTISELECT = 'taxonomy_multiselect'
+}
+
+export type CatalogViewAction = (action: CatalogActions, value?: any) => void;
+
+export function onCatalogAction(
+  action: CatalogActions,
+  value: any,
+  taxonomies: any,
+  setSearch: (value: string) => void,
```
hanbyul-here, line 15: Oh! I realize that we would need more of this type of utility function - which takes either `setState` or `setAtom`as a setter function - like the one that we use for reconciling STAC metadata here: https://github.com/NASA-IMPACT/veda-ui/pull/988/files#diff-b0d6bd8ff29016fd81054f35bfcc97dd5b50077310a81f6043a99240cb43a4eaR167

I'd like to ask your thoughts how we should type these util functions moving forward. Should we explicitly type the setter function as either setAtom or setState? ex. I tried this method in  https://github.com/NASA-IMPACT/veda-ui/pull/988/files#diff-b0d6bd8ff29016fd81054f35bfcc97dd5b50077310a81f6043a99240cb43a4eaR169 - and I am still unsure if this just added complicacy or not. But I also think `any` is a bit too broad. Any thoughts?

hanbyul-here (commented):
Wooh, yes, all the problems I mentioned are gone! I left some comments, but most of them are not blockers! 

+ comment for the future: this filter function is actually a great feature to have a unit test 😬

app/scripts/components/common/form/checkable-filter/index.tsx
hanbyul-here: Oh but I think it is possible that @faustoperez meant it only for the Exploration page Modal. Can you clarify? @faustoperez

app/scripts/components/common/form/checkable-filter/index.tsx
dzole0311: Fixed anyway.

app/scripts/components/common/catalog/utils.ts
dzole0311: Re-used from before, changed.

app/scripts/components/common/form/checkable-filter/index.tsx
faustoperez: > Oh but I think it is possible that @faustoperez meant it only for the Exploration page Modal

Yes, only on the E&A page modal. Space is tighter and filters should be collapsed there 👍

app/scripts/components/common/catalog/utils.ts
dzole0311: I think both can be verbose, even for simpler types like a setter for a string (ie. `React.Dispatch<React.SetStateAction<string>>`).

How about we create aliases for the setters and use that? For example: `type SetState<T> = Dispatch<SetStateAction<T>>` and then we can use it like this:

`const [count, setCount]: [number, SetState<number>] = useState(0);`

app/scripts/components/common/catalog/catalog-content.tsx
```diff
@@ -0,0 +1,339 @@
...
+  selectedIds,
+  setSelectedIds,
+  filterLayers,
+  emptyStateContent,
+  search,
+  taxonomies,
+  onAction,
+}: CatalogContentProps) {
+  const [exclusiveSourceSelected, setExclusiveSourceSelected] = useState<string | null>(null);
+  const isSelectable = selectedIds !== undefined;
```
hanbyul-here, line 54: It might be better to separate selectable card component  from this component since the behavior is pretty different, and the conditional blocks are pretty big. But not a blocker for now.

app/scripts/components/common/catalog/utils.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  beforeEach(() => {
+    setSearchMock = jest.fn();
+    setTaxonomiesMock = jest.fn();
+    taxonomies = {
+      Topics: ['air-quality', 'climate'],
+      Sectors: ['electricity', 'energy']
+    };
+  });
+
+  it('should clear search and taxonomies on CLEAR action', () => {
```
hanbyul-here, line 18: I was wondering if this behavior is intended or not. Great to see it on test to clarify the expected behavior

app/scripts/components/common/catalog/utils.ts
hanbyul-here: Yeah, aliases sound great. I will leave it up to you if you want to apply this change to this pr or not.

hanbyul-here (approved):
👍 Left additional comments, but nothing is a blocker. thanks for awesome work

app/scripts/components/common/catalog/catalog-content.tsx
dzole0311: Yeah I started it like that, then merged them into one. Should've probably kept it separate.

app/scripts/components/common/catalog/utils.test.ts
dzole0311: It's matching prod, but the CLEAR should probably only remove the tags (the input already has a clear button by itself). Will leave it in for now to match prev behavior.

faustoperez (approved):
Great work! 👍

hanbyul-here:
Flagging that I missed the `searchTerm` URL Parameters are needed to make a connection between the Data Catalog and the modal.

Issue #967, opened 2024-05-21
Replace Mapbox component to the new Map component in ScrollyMap component


hanbyul-here:
## Context
This is a part of https://github.com/NASA-IMPACT/veda-ui/issues/712 , continued work of https://github.com/NASA-IMPACT/veda-ui/pull/827

## Acceptance Criteria

- [ ] Scrolly Map uses new map component

## Related Tickets

https://github.com/NASA-IMPACT/veda-ui/issues/712

hanbyul-here:
Same with https://github.com/NASA-IMPACT/veda-ui/issues/969, This is not necessarily a blocker, but it will be nice if we can get to this ticket before ArcGIS layer: https://github.com/NASA-IMPACT/veda-ui/issues/943 so we have to touch only one component instead of putting almost duplicated components for both old map and new map.

PR #992, opened 2024-06-07
Replace legacy Mapbox component for Scrollytelling

dzole0311:
**Related Ticket:** [967](https://github.com/NASA-IMPACT/veda-ui/issues/967)

### Description of Changes

This PR replaces the old Map component with the new Map component for the Scrollytelling block: https://deploy-preview-992--veda-ui.netlify.app/stories/air-quality-and-covid-19

Changes:

1. Re-factored the Scrollytelling block component to work with the new `Map` and the `<Layer />` components
2. Deprecated `useAsyncLayers` in favor of using `useReconcileWithStacMetadata` 
3. Removed obsolete code and adjusted some of the types (ie. `TimelineDatasetStatus` became `DatasetStatus` as I feel we could keep this more generic, added aliases)
 
### Validation / Testing


app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -430,58 +415,50 @@ function Scrollytelling(props) {
           </CSSTransition>
         </SwitchTransition>
 
-        <Styles>
-          <Basemap />
+        <Map
```
hanbyul-here, line 418: Oh I see that only the components are replaced, and work around how the data is shaped through a utilify function(`reconcileVizDataset`). Now that I am deep in rewriting `s-explore`, I can see how this approach can help break down the task better.  On the other hand, I think we need to deprecate  `useAsyncLayers` sooner since it unnecessarily diverges how we reconcile stac data with our layer data. 

How do you think about deprecating `useAsyncLayers` ?

dzole0311: Thanks for pointing that! I'm definitely in favor of deprecating unnecessary diverging code, but just to clarify: `useAsyncLayers` accepts an array of `datasetId` & `layerId` (taken from the `Chapter's` from the MDX, validates it and then calls `useLayersInit` with the layer ids. Do you maybe meant that we could deprecate `useLayersInit` instead by switching to `useReconcileWithStacMetadata` (which I think is definitely something we could & should do)?

hanbyul-here: Yes! I meant deprecating all the old methods including `useLayersInit` , and replacing it with `useReconcileWithStsacMetadata`!

app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -492,3 +469,12 @@ Scrollytelling.propTypes = {
...
 }
+
+export function reconcileVizDataset(dataset): VizDatasetSuccess {
+  return {
+    status: DatasetStatus.SUCCESS,
+    data: dataset,
+    error: null,
+    settings: dataset.settings
+  };
+}
```
sandrahoang686, line 480: 💭 : Should this move over to a more general place like `data-utils` file if other components that need visualization type might use this?

dzole0311: Yeah, probably it will have to be moved and re-used

app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -134,103 +123,84 @@ function getChapterLayerKey(ch: ScrollyChapter) {
...
 
-  // Create a ref to cache each of the async layers.
-  // After the async layer data is loaded from STAC, the layer functions have
-  // to be resolved by the `resolveConfigFunctions`. This function will return a
-  // new object every time causing useEffects that depend on this data to fire
-  // multiple times, even though the data didn't actually change. An example of
-  // this is the `sourceParams` in `MapLayerRasterTimeseries`.
-  // Since the these values only have to be computed once, when the layer loads,
-  // we can use this cache. On every hook run the asyncLayers.map below will
-  // return the cached value if it exists or compute and cache.
```
dzole0311, line 173: @hanbyul-here @sandrahoang686 I think this cache might not be needed with the new Layer and Map component. Any issues that you see by removing this?

hanbyul-here: I am not 100% confident, but I think it is already being handled at the layer level at this point? https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/components/map/layer.tsx#L53

app/scripts/components/common/map/map-component.tsx
```diff
@@ -70,8 +77,10 @@ export default function MapComponent({
       initialViewState={initialViewState}
       mapStyle={style as any}
       onMove={onMove}
+      onLoad={onMapLoad}
       projection={mapboxProjection}
       maxBounds={maxMapBounds}
+      interactive={interactive}
```
hanbyul-here, line 83: What is this flag for? (Also I couldn't find this attribute from the doc: https://visgl.github.io/react-map-gl/docs/api-reference/map am I missing anything?)

app/scripts/context/layer-data.tsx
```diff
@@ -248,45 +248,3 @@ export const useDatasetAsyncLayers = (datasetId) => {
...
   return useLayersInit(dataset?.data.layers ?? []);
 };
-
-interface ReferencedLayer {
-  datasetId: string;
-  layerId: string;
-  skipCompare?: boolean;
-}
-
-export const useAsyncLayers = (referencedLayers: ReferencedLayer[]) => {
```
hanbyul-here, line 258: 👏

app/scripts/components/common/map/map-component.tsx
dzole0311: Yeah, we need them for the ScrollyMap. The 'interactive' prop is coming from `MapBox`

> Aside from the props listed below, the Map component supports all parameters of the Map class constructor ([Mapbox](https://docs.mapbox.com/mapbox-gl-js/api/map/) | [Maplibre](https://maplibre.org/maplibre-gl-js-docs/api/map/)).

https://visgl.github.io/react-map-gl/docs/api-reference/map#properties

app/scripts/components/common/blocks/scrollytelling/index.tsx
dzole0311: Thought so as well, and there don't seem to be side effects by not using it for the new layer. I guess we can proceed like this.

app/scripts/components/common/blocks/scrollytelling/index.tsx
```diff
@@ -134,103 +123,86 @@ function getChapterLayerKey(ch: ScrollyChapter) {
...
-        // "center: [number, number]" to "center: number[]" which fails
-        // validation.
-        resolvedLayersCache.current[index] = resolved as ResolvedLayer;
-
-        return resolved;
-      }),
-    [uniqueChapterLayers, asyncLayers]
-  );
+  const resolvedLayers = resolvedDatasetsWithStac.map((layer, index) => {
+    if (layer.status !== DatasetStatus.SUCCESS) return null;
```
sandrahoang686, line 187: ❓ : Curious to know if you had a reason to why you got rid of memoization here, if there are no benefits, can probably get rid of it in line 200 as well?

app/scripts/components/common/map/map-component.tsx
hanbyul-here: Ah right, thanks for the explanation!

app/scripts/components/common/blocks/scrollytelling/index.tsx
dzole0311: Good call, I intended to remove the [caching only](https://github.com/NASA-IMPACT/veda-ui/pull/992#discussion_r1636522808) as I think was obsolete together with the `useMemo`, but looking at the this part again I think it's better to keep the memo part. 

Updated.

sandrahoang686 (approved):

Issue #1000, opened 2024-06-11
How to integrate titiler-cmr


abarciauskas-bgse:
## Context

[Titiler-cmr](https://github.com/developmentseed/titiler-cmr) offers a way to dynamically return map tiles via [NASA's Common Metadata Repository (CMR)](https://www.earthdata.nasa.gov/eosdis/science-system-description/eosdis-components/cmr). The special sauce of titiler-cmr is that it does not require an asset url for tiling - it handles querying CMR for items and assets and then delivers tiles in response. This is similar to how titiler-pgstac works.

We want to demonstrate the use of titiler-cmr in VEDA in case it could be useful to other NASA applications or extend the set of data available to instances of VEDA.

titiler-cmr integration poses a challenge because it doesn't fit into the same model as any of the existing layer configurations. It is similar to how the `MapLayerRasterTimeseries` layer works with pgSTAC in that both submit a query to the tile API which handles querying the data catalog. But in the case of the existing "raster" type, there is no initial query to the STAC API (for tiling that is). A query is sent to the tile api (titiler-pgstac) which returns a mosaic URL which is used in the tile layer as the source. In this new case, we want to submit the catalog query parameters (to CMR in this case) as part of the tile layer source.

Having a STAC request where the STAC response is used as the parameters to the tile API (i.e. not a query to the mosaic tile registration endpoint) is more similar to how the "zarr" and "cmr" types work. However, the existing "zarr" and "cmr" types both query a STAC endpoint for an asset URL and use that asset url in the parameters sent to the tiling request. In the case of "zarr" it uses the asset URL from the collection's assets and in the case of "cmr" it uses the asset url from an item search[^1]. 

In this new scenario, we don't want to use an asset URL in the tile parameters, but the collection concept id and renders data in the STAC response to the tile endpoint.

Here is an imperfect diagram of the scenarios we support for raster layers, with the scenario in red with dotted lines representing the scenario we want to support:

![zarr-viz-arch(20)](https://github.com/NASA-IMPACT/veda-ui/assets/15016780/8c137bb6-19e8-4213-a904-dccb9b42eb91)

### How things work now

From what I understand about how the code currently works for raster layers:

1. There is a `type` value you can set on the collection which can be "raster", "cmr" or "zarr" (not including vector in this discussion)
2. If the type is raster, the `MapLayerRasterTimeseries` will submit a search payload to the STAC endpoint and then use response's href (mosaicUrl) to create a tile layer.
[...]

PR #1001, opened 2024-06-12
Titiler-cmr layer

abarciauskas-bgse:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1000

### Description of Changes

Application code changes - necessary changes
* renames ZarrPaintLayer to RasterPaintLayer, since it is not specific to Zarr
* adds a new type `titiler-cmr` and renames the `cmr` type to `cmr-stac` for disambiguation (although it is sure to be a bit confusing, since cmr-stac is for the use case of using the CMR STAC catalog and titiler-cmr is for the use case of tiling data in CMR but those requests are setup via collection records catalogued in VEDA STAC)
* adds a third hook `useTitilerCMR`. Because this third hook needs to return tiling parameters besides assetUrl, refactor the other hooks to also return a tileParams dictionary.

Application code - optional changes, simplifying some things
* Merges/removes duplication across MapLayer*TimeseriesProps to just one interface `MapLayerRasterTimeseriesProps` since the only thing that seemed to separate these types was `assetUrlReplacements`. This change is not strictly necessary for this functionality to work but it seemed to simplify things IMO.
* Simplify conditionals which set the datetime extent and periodicity. The VEDA STAC metadata for these CMR  collections uses `null` as the second datetime value in the temporal extent to indicate ongoing collections. This is the declared way to do so in the STAC collection specification, so it seems valid to me to allow for all non-vector collections to set the end datetime to the current date if it is null in the metadata.

Dataset changes
* Adds 2 datasets to demonstrate titiler-cmr is working for those datasets. We could go down to one since it seems cumbersome to manage multiple mock datasets for a single use case.
* Removes the GPM IMERG cmr-stac collection, since this overlaps with the titiler-cmr example and we can already cover the cmr-stac example with the TRMM dataset
* Adds the cmip6 kerchunk reference dataset, so there is a dataset to test with the "zarr" type

I would be happy to go over these changes with one or more of the core UI developers.

### Validation / Testing

I tested these changes by opening up all of the mock datasets explore pages.


abarciauskas-bgse:
Will address linting and ts-check later today

abarciauskas-bgse:
Checks are passing but unsure about the changes in style-generators and why it seems like there is duplication of files there. Would love to pair with UI team member(s) to review!

sandrahoang686:
@abarciauskas-bgse thanks for detailed description and for the explanation on #1000. I dont mind campfiring on this together for review! It will be more interactive though since I haven't worked with the style-generators much. I'm guessing the duplication of files may be related to this past PR https://github.com/NASA-IMPACT/veda-ui/pull/805 (specifically maybe this commit https://github.com/NASA-IMPACT/veda-ui/pull/805/commits/0b2dece29f2001707e231e6bccccee3509c207ef)

sandrahoang686:
@abarciauskas-bgse the veda dashboard team is also working on deprecating the `/common/mapbox/` directory as we are are moving to consolidate our maps and mapping logic to a the new map provider and logic under dir `/common/map/`. This is why you may be seeing duplicated logic in both these places. Work is still happening on migrating current maps and logic to the new directory. I recommend if possible to try to minimize references to anything under `/common/mapbox/` and if the logic doesn't already exist under the new dir `/common/map/` if you would be comfortable moving this logic over, I can work with you to do that if as well

app/scripts/components/common/mapbox/layers/raster-paint-layer.tsx
```diff
@@ -0,0 +1,114 @@
+import { useEffect, useMemo } from 'react';
```
sandrahoang686, line 1: 🔧 : We are working to deprecate this path. The new path for this logic should go under `/common/map/`. I believe you would want to migrate this over to `/common/map/style-generators/`.

app/scripts/components/common/map/style-generators/hooks.ts
```diff
@@ -23,8 +23,13 @@ interface CMRResponseData {
   }[]
 }
 
-export function useZarr({ id, stacCol, stacApiEndpointToUse, date, onStatusChange }){
-  const [assetUrl, setAssetUrl] = useState('');
+interface STACforCMRResponseData {
+  collection_concept_id: string;
+  renders: Record<string, any>;
+}
```
sandrahoang686, line 29: I wonder 🤔 if the backend team has json schemas we could generate into typescript interfaces to import, interested.. I'll follow up with them on this

app/scripts/components/common/map/style-generators/hooks.ts
```diff
@@ -90,11 +103,73 @@ export function useCMR({ id, stacCol, stacApiEndpointToUse, date, assetUrlReplac
...
+        const data: STACforCMRResponseData = await requestQuickCache({
+          url: `${stacApiEndpointToUse}/collections/${stacCol}`,
+          method: 'GET',
+          controller
+        });
+
+        let tileParams = {
+          concept_id: data.collection_concept_id,
+          datetime: date,
+          ...sourceParams
```
sandrahoang686, line 152: 🔧 : This matches the naming of local state variable, can we change the name so it isn't confusing.

app/scripts/components/common/map/style-generators/hooks.ts
```diff
@@ -38,11 +43,19 @@ export function useZarr({ id, stacCol, stacApiEndpointToUse, date, onStatusChang
           controller
         });
 
-        setAssetUrl(data.assets.zarr.href);
+        const tileParams = {
```
sandrahoang686, line 46: 🔧 : This matches the local state variable's name, can we change this so it isn't confusing? Or i'd just set state directly with the object like so...
```
        if (data.assets.zarr.href) {
          setTileParams({
             url: data.assets.zarr.href,
             time_slice: date,
             ...sourceParams
           });
        }
```

abarciauskas-bgse:
just a update that I am making some changes based on discussion with @sandrahoang686 (to use the new map component, to fix conflicts and to fix some variable naming). so will move this to draft until that is done and this is ready for review again.

hanbyul-here:
Can you also provide details on how the temporal text/summaries should be handled? @abarciauskas-bgse   I remember we briefly talked about this before, but it's been a while. We are currently shimming the end date to the current day, and density to DAY if the data is missing the density & temporal extent's end date. Is this still the case?

abarciauskas-bgse:
Closing to propose a new design for different layer types

PR #1003, opened 2024-06-17
[DO NOT MERGE] Spike: USWDS experiment

dzole0311:

### Description of Changes

### Validation / Testing


hanbyul-here:
👋  Re: css tree shaking - I am not sure if I am missing anything, but Parcel documentation says only named imports can take advantage of tree shaking: https://parceljs.org/languages/css/#tree-shaking . (I do not know how I can use named imports with scss? Is this even doable?) 

Considering this issue is still opened in uswds - https://github.com/uswds/uswds/issues/5210 I wonder if it is not doable with uswds style at this point? - If so, it might be beneficial to build css without any hash for now so it gets cached?

dzole0311:
> 👋 Re: css tree shaking - I am not sure if I am missing anything, but Parcel documentation says only named imports can take advantage of tree shaking: https://parceljs.org/languages/css/#tree-shaking . (I do not know how I can use named imports with scss? Is this even doable?)

Yeah, that's what I noted in the spike doc as well. 

> Considering this issue is still opened in uswds - [uswds/uswds#5210](https://github.com/uswds/uswds/issues/5210) I wonder if it is not doable with uswds style at this point? - If so, it might be beneficial to build css without any hash for now so it gets cached?



> 👋 Re: css tree shaking - I am not sure if I am missing anything, but Parcel documentation says only named imports can take advantage of tree shaking: https://parceljs.org/languages/css/#tree-shaking . (I do not know how I can use named imports with scss? Is this even doable?)

Yeah, I had the same problem which I mentioned in the spike doc. I think if there were uswds component mixins (ie. 'card'), we could have found a way to use the named import in Parcel but IMO on the expense of the dev UX:

```
// card.module.scss
.card {
    @include card <- mixin that doesn't exist in uswds
}

// card.tsx

import * as CardStyle from './card.module.scss'

<SomeCard className={CardStyle.card} />
```

> Considering this issue is still opened in uswds - [uswds/uswds#5210](https://github.com/uswds/uswds/issues/5210) I wonder if it is not doable with uswds style at this point? - If so, it might be beneficial to build css without any hash for now so it gets cached?

Fully hashless? Not sure how caching is set up in the instances but we have to assess potential cache invalidation issues for some clients browsers. Thinking of which, each hash is generated per veda instance release, so as long as it stays the same it should be cached across soft reloads of the page 🤔

hanbyul-here:
@dzole0311 Ah, yes, you did mention it in the Spike doc. I'm sorry I blanked on it!

I meant detaching hash only for USWDS CSS since the CSS file content won't change if we include every style from USWDS—all other assets should still be hashed. We will likely need to tweak how assets are built for this - as far as I understand, we build assets per page, and any imported asset gets included into `${page}.hash.css`. We would need to separate uswds css from other assets on the page if we want the hashless cache strategy to work 🤔 

I still want to see if there are any more options to purge css afterward, such as https://www.npmjs.com/package/purgecss.

dzole0311:
Work continues in [1029](https://github.com/NASA-IMPACT/veda-ui/pull/1029)

Issue #905, opened 2024-03-26
Clean up Compare, Base, and Reconciled Typed interfaces to Prevent Casting


sandrahoang686:
Description:

We need to revisit our typing as there is a ton of casting going on that is masking our bad types. Our types should be more explicit so we dont need to cast or optionally chain as much.

Acceptance Criteria:
* Revisit Compare Types in `/parcel-resolver-veda/index.d.ts`
* Revisit Reconciled Types returned from `useReconcileWithStacMetadata`

hanbyul-here:
Assuming this need came when we are trying to make the 'old' component ( the components that live outside of `/exporation` work with 'new' state management, I wonder if this will be better resolved by structuring Atoms. We are trying to reuse timelinedatasetAtoms, which is not meant to serve the other parts of the applications, and I think type mismatch happens while repurposing the `timelinedatasetAtoms`?

It makes sense that the types that `useReconcileWithStacMetadata` return might not align with our requirements elsewhere in the application. I initially thought the types defined in `/parcel-resolver-veda/index.d.ts` were more generalized, mainly reflecting the MDX structure. But now, I'm curious about your thoughts on how we should reshape the type. Can you detail your findings when you have a chance?

aboydnw:
this is one of those tickets where I feel like all of the words you used are english but somehow I have no freaking clue what you said 😆

aboydnw:
@hanbyul-here @sandrahoang686 with the linked PR I am wondering if we can close this ticket?

sandrahoang686:
@aboydnw i'm fine with closing for now with the linked PR that was merged, this might come back up again because Im unsure how much we cleaned but if thats the case, i'll just create another ticket that is more specific with types to re-look at 👍🏼 .

PR #1006, opened 2024-06-18
Use VizDataset data type for block map, set a default value for interactive

hanbyul-here:
**Related Ticket:** 
Kind of related to https://github.com/NASA-IMPACT/veda-ui/issues/905 ? We would still need to revisit 'compare' - but I think we can deprecate it soon?

### Description of Changes
- Make BlockMap use the new type `VizDataset.` 
- Simplify some redundant logic (ex. `resolveConfig` now happens at the layer level, so I don't think we need to do it on block map level)
- While working on it, I realized that a new flag, `interactive,` blocks the interaction such as dragging to pan - I set the default value as `true` (which was an implicit value before.)

### DIFF
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -178,9 +178,9 @@ function MapBlock(props: MapBlockProps) {
     return totalLayers;
   },[layerId]);
 
-  const [layers, setLayers] = useState<VizDataset[] | undefined>(layersToFetch);
+  const [layers, setLayers] = useState<VizDataset[]>(layersToFetch);
 
-  useReconcileWithStacMetadata(layersToFetch, setLayers);
+  useReconcileWithStacMetadata(layers, setLayers);
 
   const selectedDatetime: (Date | undefined) = dateTime
     ? utcString2userTzDate(dateTime)
@@ -295,7 +295,7 @@ function MapBlock(props: MapBlockProps) {
         }}
       >
         <Basemap basemapStyleId={mapBasemapId} />
-        {selectedDatetime && layerId && baseDataLayer && (
+        {selectedDatetime && baseDataLayer && (
           <Layer
             key={baseDataLayer.data.id}
             id={`base-${baseDataLayer.data.id}`}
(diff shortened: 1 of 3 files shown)


app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -197,57 +193,39 @@ function MapBlock(props: MapBlockProps) {
...
+    if (compareLayer.status !== S_SUCCEEDED) return null;
+
+    return {
+      ...compareLayer,
+      settings: {
+        isVisible: true,
+        opacity: 100
+      }
+    };
+  }, [compareLayers]);
```
sandrahoang686, line 226: ⛏️ : Just a though to make it less redundant. Combine baseDataLayer and compareDataLayer logic into a function instead?

something like...
```  
const getDataLayer = (layers: VizDataset[] | undefined): (VizDatasetSuccess | null) => {
    if(!layers || layers.length !== 1) return null;
    const someLayer = layers[0];

    if(someLayer.status !== S_SUCCEEDED) return null;
    return {
      ...someLayer,
      settings: {
        isVisible: true,
        opacity: 100
      },
    }
  };

  const baseDataLayer = useMemo(() => getDataLayer(baseLayers), [baseLayers])
  const compareDataLayer = useMemo(() => getDataLayer(compareLayers), [compareLayers])
  ```
  
  But current way is also fine which is just more explicit. whichever

File: app/scripts/components/common/blocks/block-map.tsx [hunk 2]
```diff
@@ -1,8 +1,7 @@
 import React, { useMemo, useState, useEffect } from 'react';
 import styled from 'styled-components';
-import { DatasetDatumFnResolverBag, ProjectionOptions, datasets } from 'veda';
+import { ProjectionOptions } from 'veda';
```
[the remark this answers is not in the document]
hanbyul-here: Yes! I was thinking about how we could replace this type import soon. Since we will need to replace most of type imports once Oliver's branch is merged, I don't think we need to make it as a dependency now. but a note can help

app/scripts/components/common/blocks/block-map.tsx
hanbyul-here: This is a great suggestion. I also made some changes so that both layers can be fetched with one call of `  useReconcileWithStacMetadata(layersToFetch, setLayers);`, and combined with your suggestion.

sandrahoang686 (approved):
🙇🏼‍♀️

hanbyul-here:
I did not benchmark yet, but I notice that BlockMap seems to fail to load more often on this PR. I am investigating it. @sandrahoang686

Issue #1007, opened 2024-06-18
setProjection for scrollytelling not working


hanbyul-here:
## Context

While migrating Scrollytelling component to the new map component, a bug that setProjection not working was introduced. 
Refer to the original PR and the comment: 
https://github.com/NASA-IMPACT/veda-ui/pull/992/files#diff-af7791d70ec6eacf00bc029b1e7b44b287a9d2c6cc737a56acd4d1b20cda8f1cR306-R307

## Acceptance Criteria

- [ ] Given when Scrollytelling block can set the projection with a paassed `projectionId` value .

## Related Tickets
https://github.com/NASA-IMPACT/veda-ui/issues/967

PR #1008, opened 2024-06-18
use getMap for access to hidden member from Scrollytelling

hanbyul-here:
**Related Ticket:**
Fix for #1007 

### Description of Changes
Use `getMap` method to access to hidden method (`setProjection`) 

- Would it be better to return the referenced object from react-map-go instance level? https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/map/map-component.tsx#L71
- Would it be better to wrap MDX Map component with MapProvider so we can use `useMap` to access mapRef? https://visgl.github.io/react-map-gl/docs/api-reference/use-map

### Validation / Testing
Check the scrolly telling example on sandbox page, see the projection changes.

### DIFF
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -303,8 +303,9 @@ function Scrollytelling(props) {
           : undefined;
 
         projection &&
-          // @ts-expect-error setProjection is missing on type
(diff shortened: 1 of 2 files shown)


hanbyul-here:
Hm this makes the error go away, but doesn't solve the problem 🤔 I will close this for now.

Issue #1010, opened 2024-06-24
E&A layer selection modal should omit datasets with disableExplore


j08lue:
Some datasets are disabled for Exploration & Analysis, because they do not work well with that interface (optimized for raster timeseries with consistent spatial coverage).

This is done via the `disableExplore` setting.

https://github.com/NASA-IMPACT/veda-ui/blob/7b672a27618307e3a956978de3c63537454ae8e0/docs/content/CONTENT.md?plain=1#L115-L117

It appears that this setting is not respected, since some datasets that should not be shown in the E&A layer selection modal are selectable in the latest release:

https://github.com/US-GHG-Center/veda-config-ghg/blob/cead60acf57c2c6cb6d0c46de20f579d7498c096/datasets/emit-ch4plume-v1.data.mdx?plain=1#L36

https://github.com/US-GHG-Center/veda-config-ghg/blob/cead60acf57c2c6cb6d0c46de20f579d7498c096/datasets/noaa-cpfp-ch4-point.data.mdx?plain=1#L42

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/a6cd080f-884a-435a-94ad-3d9271f465a2)

## Acceptance criteria

- [ ] Datasets configured with `disableExplore` are not shown in E&A layer selection modal

j08lue:
Maybe we just need to use `allExploreDatasets` here instead of `allDatasets` here:

https://github.com/NASA-IMPACT/veda-ui/blob/7b672a27618307e3a956978de3c63537454ae8e0/app/scripts/components/exploration/components/dataset-selector-modal/index.tsx#L116

https://github.com/NASA-IMPACT/veda-ui/blob/7b672a27618307e3a956978de3c63537454ae8e0/app/scripts/components/exploration/data-utils.ts#L47-L49

PR #1013, opened 2024-06-24
v5.1.2: Use explore only datasets for exploration page

hanbyul-here:
**Related Ticket:** #1010 

### Description of Changes
Use Exploration only datasets

### Validation / Testing
Check if the new E&A page returns the dataset that has the `disableExplore` attribute as true.


hanbyul-here:
Sneaking in patch version bump to release the fix right after merge.

PR #1015, opened 2024-06-25
Create bug.md

aboydnw:
Added a light bug template for creating issues quickly

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -6,9 +6,13 @@ Blocking a release? (y/n):
 
 Priority: 
 
+<!--
+Choose P1, P2, or P3
+
 - P1 - Blocking major workflow, pull into sprint and fix now
 - P2 - Poor reflection of quality, find a way to fix in the next few sprints
 - P3 - Minor annoyance or issue, place in backlog for future consideration
+-->
 
 
 ### Steps to reproduce the issue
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,18 @@
+### Context
+
+URL:
```
j08lue, line 3: What URL?

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,18 @@
...
+
+URL: 
+
+Blocking a release? (y/n): 
+
+Priority: 
+
+- P1 - Blocking major workflow, pull into sprint and fix now
+- P2 - Poor reflection of quality, find a way to fix in the next few sprints
+- P3 - Minor annoyance or issue, place in backlog for future consideration
```
j08lue, line 11:
```suggestion
<!--
Choose P1, P2, or P3

- P1 - Blocking major workflow, pull into sprint and fix now
- P2 - Poor reflection of quality, find a way to fix in the next few sprints
- P3 - Minor annoyance or issue, place in backlog for future consideration
-->
```

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -0,0 +1,18 @@
...
+
+
+### Reproduction Steps
```
j08lue, line 14:
```suggestion
### Steps to reproduce the issue
```

j08lue (changes requested):
Good call to have a bug report template.

I noticed in the past that contributors do not remove parts from the issue template - e.g. when there are three options and they should choose one, they leave in all three. Therefore, I made a suggestion to include the priority levels to choose from only as an instruction that is hidden once you submit.

Issue #1016, opened 2024-06-25
Clicking on a dataset tag does not filter data catalog


aboydnw:
### Context

URL: https://staging.earth.gov/ghgcenter/

Blocking a release? (y/n): no

Priority: P2

- P1 - Blocking major workflow, pull into sprint and fix now
- P2 - Poor reflection of quality, find a way to fix in the next few sprints
- P3 - Minor annoyance or issue, place in backlog for future consideration

### Reproduction Steps

1. Visit data catalog page
2. Open any dataset
3. Click on any tag/topic at the top the dataset
4. You should be redirected to the data catalog page, filtered for the tag you clicked
5. Notice the URL has some parameters, but it is not correctly filtering

hanbyul-here:
There were more problems related to query paramter generated by data-catalog. Putting those bugs here for the record.

- The Search query parameter is not respected in the Data Catalog page
  - Type 'air' as a search term. The query parameter for 'search' gets manipulated, but when I refresh, it gets wiped out as soon as the page loads.
- Navigating back to Data Catalog through Taxonomy Pills on the overview button does not work. 
  - Click on any taxonomy pill from data overview page. `taxonomy` query paramter gets wiped out as soon as the page loads.

hanbyul-here:
Closed via https://github.com/NASA-IMPACT/veda-ui/pulls?q=is%3Apr+is%3Aclosed

PR #1020, opened 2024-06-25
Put a hot fix for URL search paramter *WIP*

hanbyul-here:
**Related Ticket:**  #1016 

### Description of Changes

Several issues are occurring with data catalog and URL manipulations. I fixed some of them in this PR and hot-glued some of them (I think we can fix them better, though). 

1. The `Search` query parameter is not respected in the Data Catalog: I 'hotfix' it by reading the query parameter directly from the URL when the `search` parameter exists in the URL.
**Now it shows a weird behavior of URL bouncing back to data-catalog When 'Explore data' was clicked from a data overview page, without actually bouncing the page back. 🤔**  
2. The `search` query parameter gets wiped out when the page is loaded: This seems to be because the `CLEAR` action gets triggered on page load, which includes clearing both search terms and taxonomies. I separated these actions out 
3. The `taxonomies` parameters are not being respected when clicked from data overview page. 

### Validation / Testing


hanbyul-here:
Closing for https://github.com/NASA-IMPACT/veda-ui/pull/1021

PR #1021, opened 2024-06-26
Use different query parameter managements between data-catalog page and new E&A page

hanbyul-here:
**Related Ticket:** #1016 

### Description of Changes

This PR makes Catalog View on the data-catalog page use QS-state for query parameters, while Catalog View on the new E&A page uses Jotai. The current Atoms related to URL management is not designed for cross-page navigation. I drew a diagram to describe the navigation problem between catalog page and overview page.

Screenshot 2024-06-26 at 1 19 05 PM

More context: 
We moved our query parameter management to use Jotai because of the concerns about the mix-use of QS-State and Jotai for query parameters on the same page. (Specifically - for the new E&A page, other parameters are managed by Jotai). More details on this PR that initially introduced data-catalog view to new E&A page: https://github.com/NASA-IMPACT/veda-ui/pull/989#discussion_r1625643361 In the end, we ditched the Query parameter manipulation thinking that we don't need. (Details are in this comment: https://github.com/NASA-IMPACT/veda-ui/pull/989#issuecomment-2148023391) But the bug reported #1016 made me realize that was a wrong call. 

- We need a better solution for long-term url management. 

### Validation / Testing
Navigate back and forth from data catalog to overview, from overview to data catalog (especially through taxonomy pills on the overview page)
Navigate to the new E&A page through the 'Explore Data' button on the overview page.

I opened a pr on ghg for testing: https://github.com/US-GHG-Center/veda-config-ghg/pull/421


app/scripts/components/common/catalog/catalog-content.tsx
```diff
@@ -108,11 +108,11 @@ function CatalogContent({
...
     if (!selectedFilters.length) {
-      onAction(CatalogActions.CLEAR);
-
-      if (!isSelectable) {
-        navigate(DATASETS_PATH);
-      }
+      onAction(CatalogActions.CLEAR_TAXONOMY);
+      // @NOTE: When is this used? Is this safe to remove?
+      // if (!isSelectable && !search) {
+      //   navigate(DATASETS_PATH);
```
hanbyul-here, line 114: This was wiping search query parameter, so I commented this out. But I am not sure what this was trying to achieve. Do you see any problem removing this line @sandrahoang686 ?

sandrahoang686: 🤔 i cannot remember, I think its safe to remove but I can help QA again before its merged

hanbyul-here:
This PR is ready for the review @sandrahoang686  I ended up with a few more changes to fix the bug that @aboydnw  reported in https://github.com/NASA-IMPACT/veda-ui/pull/1022 

I deprecated `useBrowseControls`, now StoriesHub also uses `useCatalogAction` (I am not sure if this hook might need to be renamed? but maybe not since everything is a catalog? 🙃 ). So I would appreciate if extra attention could be paid to stories + URL query parameters 🙇

aboydnw:
I tried to break the Dataset overview -> exploration link, adding/editing layers, and the stories catalog. The only thing that I noticed is the following:
 
 1. Open the add layer modal (either manually or by clicking explore data from the dataset overview page)
 2. Make sure there is a a search term or filter selected
 3. Select a layer to view
 4. Click add to map
 5. Click Add Layer to reopen the select layer modal
 6. Notice that the filters persist

This is different from how it behaves in prod. I could see arguments for both experiences, but I wanted to call it out in case we intentionally wiped filters in the old layer selection modal. cc @faustoperez

hanbyul-here:
Thank you @aboydnw , that is a sharp catch! That behavior you described, is precisely why we ditched 'keeping the current state of filtering as URL parameters' at first (Details are: https://github.com/NASA-IMPACT/veda-ui/pull/989#issuecomment-2148023391). However, we learned we needed it to manipulate the current filter state through query parameters. (So the modal can show the filtered layers on landing when a user comes from the Data Overview page.) 

But I agree that it is a weird behavior that the search state persists after the purpose of filtering is fulfilled - so I added 'reset' of search terms and taxonomies whenever the layer selections are confirmed, or the layer selection modal is closed (hidden).

app/scripts/components/common/catalog/controls/hooks/use-catalog-view.ts
```diff
@@ -17,6 +25,44 @@ export function useCatalogView() {
...
     search,
     taxonomies,
-    onAction
+    onAction,
   };
-}
+} 
+// @NOTE: A hook using qs-state for query parameter management for cross-page navigation
+// Details: https://github.com/NASA-IMPACT/veda-ui/pull/1021
+export function useCatalogViewQS(): UseCatalogViewResult {
```
sandrahoang686, line 33: ⛏️ : What do you think about changing the name to `useFiltersUrlManagement` because this is mostly tied to the filters instead of the greater catalog view. We can probably change the interface to be `UseFiltersUrlManagementResults`... its long though..

app/scripts/components/common/catalog/prepare-datasets.ts
```diff
@@ -1,22 +1,30 @@
...
-    sortField: string | null;
-    sortDir: string | null;
-    filterLayers: boolean | null;
-  }
-) => {
+const isDatasetData = (data: DatasetData | StoryData): data is DatasetData => {
+  return 'layers' in data;
+};
+
+interface OptionsType {
```
sandrahoang686, line 9: ⛏️ : more explicit? maybe `FilterOptions`?

app/scripts/components/common/catalog/utils.ts
```diff
@@ -1,9 +1,15 @@
 import { omit, set } from 'lodash';
+import { optionAll } from '$components/common/browse-controls/constant';
 
 export enum CatalogActions {
+  TAXONOMY_MULTISELECT = 'taxonomy_multiselect',
   CLEAR = 'clear',
```
sandrahoang686, line 6: I think we can get rid of the CLEAR option if we are going to be explicit in the about which CLEAR operation we are doing (taxonomies or search)

sandrahoang686 (commented):
🙇🏼‍♀️Nice and fast work on this! Left some comments but none if it huge, logic makes sense, comments are minor

sandrahoang686:
QA Findings
| Test | Result |
| --- | ----------- |
| Selecting different filters across groups and clearing tags | ✅  |
| Selecting filters and clicking into dataset and then pill tag takes you to correct filter view | ✅  |
| Search persist | ✅  |
| Selecting filters and Search work correctly in modal | ✅  |
| Clicking on "Explore data" navigates to the modal with search correct | ✅  |

app/scripts/components/common/catalog/utils.ts
hanbyul-here: I think it can go either way, but `CLEAR` might be a better option when we intend to reset the state of filtering? I changed the modal to use `CLEAR` whent it needs to reset the state!

app/scripts/components/common/catalog/controls/hooks/use-catalog-view.ts
```diff
@@ -17,6 +25,44 @@ export function useCatalogView() {
...
+  const navigate = useNavigate();
+  const useQsState = useQsStateCreator({
+    commit: navigate
+  });
+
+  const [search, setSearch] = useQsState.memo(
```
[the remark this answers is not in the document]
hanbyul-here: Oh, That sounds like how the current main branch is using the filters with the states in the data selection modal, but in a separate hook. Am I getting it right? 

Aside from how we will handle the query parameter in the future... I wonder if there is a better place than the code base where this kind of discussion should go so it can get deserving attention. I changed `@NOTE` prefix to `@TECH-DEBT`to signal that actions need to be taken.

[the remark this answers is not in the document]
sandrahoang686: Not sure how its managed in the modal but yep talking about not combining taxonomies state and url changes likes whats happening in the `atomWithUrlStability` and to have them separate

sandrahoang686 (approved):

app/scripts/components/common/catalog/controls/hooks/use-catalog-view.ts
hanbyul-here: (This also required renaming of related variables to prefix by `Filter`, instead of `Catalog` !)

app/scripts/components/common/catalog/catalog-content.tsx
dzole0311: Hmm, I don't think we need it; I should have asked when porting the Data Catalog to the modal if I should remove it but missed it.

PR #1029, opened 2024-07-02
Base setup for using react-uswds in the veda-ui

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1043

### Description of Changes
This is a base setup to start using react-uswds inside of the veda-ui

### Example usage

To use the `Card` component from `react-uswds`, we can simply import it and use it as is from react-uswds, with the default styles applied.

### Validation / Testing


postcss.config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  syntax: 'postcss-scss',
+  parser: 'postcss-safe-parser',
+  plugins: [
+    require('autoprefixer'),
+    require('postcss-import'),
+    require('postcss-uncss')({
```
hanbyul-here, line 7: I put this to purge the styles not used, but I think this significantly slows down the dev server. I will need to put some conditions to use it only for production build

postcss.config.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  syntax: 'postcss-scss',
+  parser: 'postcss-safe-parser',
+  plugins: [
+    require('autoprefixer'),
+    require('postcss-import'),
+    require('@fullhuman/postcss-purgecss')({
```
hanbyul-here, line 7: @dzole0311 @sandrahoang686 
Here, I am trying to purge out CSS that is not being used. I think we are close based on the preview generated from this PR: https://github.com/NASA-IMPACT/veda-ui/pull/1031 Preview: https://deploy-preview-1031--veda-ui.netlify.app/ 

But this definitely slows down the dev process a lot, so we likely want to include this process only for the build process. And I would like to check what happens when we start customizing scss with tokens.

FYI: I think our use case is very similar to this example from purgecss doc : https://purgecss.com/ant_design.html

dzole0311: Nice! Seems like `purgeCss` can't understand nested selectors, while `uswds` makes a heavy use of that pattern. I digged into the `purgecss` issues and [it seems to be the case](https://github.com/FullHuman/purgecss/issues/72)? I pushed a change to safelist some classes and [now it works](https://deploy-preview-1031--veda-ui.netlify.app/), but we should streamline the configuration so that we don't have to add each class manually (looking at the `defaultExtractor` it could be that we can customize it to include nested selectors?).

hanbyul-here: Thanks for looking into the safelist! 
Yes, I agree that manually adding the classes to the whitelist will not be good for the long term. 🤔 I guess we need a 'dynamic safelist'—Making any selector related to the used one not to be purged. I feel like it should be doable, but I can't seem to find a quick way to do it. I will dig a little bit more.

sandrahoang686: Hmm sorry just catching up.. but just reading up on this and other options that we could do. I read up on this article [here](https://www.useanvil.com/blog/engineering/purgecss-styled-components/#feature-comparison-what-about-nested-selectors). Which has a great explanation but then offers a solutions of considering other alternatives other than `purgeCss` - like `purifyCss` https://www.useanvil.com/blog/engineering/purgecss-styled-components/#alternatives-to-purgecss, could we considering using that instead? Looks like there are pros/cons for each but good to know there are other options that can also help with this issue

hanbyul-here:
One thing that blocks me from merging this PR is how slow the dev server is. First, are you also experiencing the slow dev server problem @dzole0311 ? Would it be possibly related to warning from the transformer that @sandrahoang686 pointed out? `@parcel/transformer-postcss: WARNING: Using a JavaScript PostCSS config file means losing out on caching features of Parcel` - but I am unclear how I can migrate postcss.config.js to .postcssrc with plugins configuration now. (https://github.com/parcel-bundler/parcel/issues/5699) I will continue this tomorrow

postcss.config.js
dzole0311: Thanks for the suggestion @sandrahoang686. I did some changes [today](https://github.com/NASA-IMPACT/veda-ui/pull/1029#issuecomment-2217268369), and it seems that we might not need to purge the css (at least for now) 🤔

hanbyul-here:
Ah, thank you so much for redirecting the configuration @dzole0311 ! (Now I also can see that the verbose warning comes from the instruction with how to turn it off 😓 ) - I misunderstood how uswds design system (or even just scss) works, in a way that we need a whole scss to make use of its full capacities. 

One last thing I am checking is if any change we are introducing is interrupting the library build ( I should just make it as a part of the build command). I can't seem to make the build work with the new version of Parcel. Sorry to keep dragging the pr, but this seems to be a big enough issue to be a blocker. I am looking at it, but lmk if you have a clue about it. This is the error message I get when I run `publishlib` with `Parcel@2.12` . I am not sure how accurate this error is, since we didn't have any problem with aliases with the previous version of Parcel for library build.

```

@parcel/core: Failed to resolve '$components/common/layout-root/useSlidingStickyHeaderProps' from './app/scripts/components/common/catalog/index.tsx'

  /Users/hanbyul/ds/git/veda-config-ghg/.veda/ui/app/scripts/components/common/catalog/index.tsx:9:8
     8 |   useSlidingStickyHeaderProps
  >  9 | } from '$components/common/layout-root/useSlidingStickyHeaderProps';
  >    |        ^
    10 | 
    11 | import {

@parcel/resolver-default: External dependency "$components" is not declared in package.json.

  /Users/hanbyul/ds/git/veda-config-ghg/.veda/ui/package.json:111:3
    110 |   },
  > 111 |   "dependencies": {
  >     |   ^^^^^^^^^^^^^^
    112 |     "@codemirror/lang-markdown": "^6.1.1",
    113 |     "@codemirror/state": "^6.2.1",

  💡 Add "$components" as a dependency.
```

dzole0311:
> I can't seem to make the build work with the new version of Parcel

Hmm, yeah I ran the `buildlib` now and I do get that error. I also noticed errors (albeit not the same) with the previous Parcel v2.7 and using the `main` branch. Might be related to something on `main`? I rebased this branch with `main` a couple of days ago.

hanbyul-here:
I looked into the problem more today. I gradually increased the minor version of Parcel and its related dependencies to see when the bug was introduced. The problem appeared from Parcel 2.9.0, when Parcel rewrote its resolvers extensively.https://parceljs.org/blog/v2-9-0/ I dipped around many things, but eventually, Parcel doesn't compile aliases when the target is not a web app. I found a similar issue from the repo: https://github.com/parcel-bundler/parcel/issues/9519 

I am kind of blocked here now 🤔 Hopefully I will have more ideas tomorrow.

dzole0311:
Thanks for looking into it @hanbyul-here , I'm throwing some ideas below.

One option we could try is to downgrade to Parcel v2.7 and try the `errorRecovery` option for "@parcel/transformer-css" (https://github.com/parcel-bundler/website/issues/1071) and see if Parcel parses the CSS without problems. If this works then we could ditch the postcss transformer.

A second alternative would be to downgrade to Parcel v2.7 and stop using postcss to fix the invalid CSS from USWDS, and to write a custom Parcel Transformer that patches the css using Parcel v2.7. Something like this: https://github.com/parcel-bundler/lightningcss/issues/39#issuecomment-1098078026. The downside is that I can't tell how many invalid selectors we have to fix.

hanbyul-here:
Thank you @dzole0311 for the suggestion! I think all your suggestions make a lot of sense. I was wondering if there is any way not to lock ourselves to Parcel 2.7, since 2.7 is quite old anyway at this point. Your comment made me think about using Parcel's Resolver API to resolve aliases. I made a PR: https://github.com/NASA-IMPACT/veda-ui/pull/1045

hanbyul-here:
I found another problem when I tried to build the UI from ghg-config side. The build fails 
```

@parcel/transformer-sass: Can't find stylesheet to import.
  ╷
1 │ ┌ @use 'uswds-core' with (
2 │ │     $utilities-use-important: true,
3 │ │     $theme-show-notifications: false
4 │ │ );
  │ └─^
  ╵
  .veda/ui/app/scripts/styles/_uswds-theme.scss 1:1  @forward
  .veda/ui/app/scripts/styles/styles.scss 1:1        root stylesheet

  Error: Can't find stylesheet to import.
    ╷
  1 │ ┌ @use 'uswds-core' with (
  2 │ │     $utilities-use-important: true,
  3 │ │     $theme-show-notifications: false
  4 │ │ );
    │ └─^
    ╵
    .veda/ui/app/scripts/styles/_uswds-theme.scss 1:1  @forward
    .veda/ui/app/scripts/styles/styles.scss 1:1        root stylesheet
```

I think it is the path problem, and should be fixable. I will update once I make more progress.

hanbyul-here:
I think now the build for library, instance, and UI dev environment should all work. phew 💦  I will give an approval for thispr, but please test for all the builds before the merge !

hanbyul-here (approved):

dzole0311:
:white_check_mark: veda-ui: serving the app, building the app, building the lib, the uswds design system works as before
:white_check_mark: veda-config: updating the submodule, building and running the instance: https://github.com/NASA-IMPACT/veda-config/pull/417
:white_check_mark: veda-config-ghg: updating the submodule, building and running the instance https://github.com/US-GHG-Center/veda-config-ghg/pull/438
:white_check_mark: veda-config-eis: updating the submodule, building and running the instance https://github.com/NASA-IMPACT/veda-config-eic/pull/123

PR #1031, opened 2024-07-02
Announcement component using USWDS

hanbyul-here:
I made banner as a part of veda config: https://github.com/NASA-IMPACT/veda-ui/pull/1031/files#diff-828262a6a39b2c45c85b5fcdf1b4615ff5d3d69f662078fd27aa9846fb29f36fR20 (Compare this configuration to the current setup: https://github.com/US-GHG-Center/veda-config-ghg/blob/develop/veda.config.js#L45-L49) 

I structured the component with the ongoing refactoring in mind. But let me know if you see any problem 🙇 

I also re-introduced purge-css because I wanted to use utility classes from uswds. This comes with some prices ex. manual safelist and no hot reloading. I left a comment in the code, I am open to taking a different approach,

- [x] Add a documentation about configuring Banner


app/scripts/components/common/banner/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+      BANNER_KEY,
+      actionUrl
+    );
+    setIsOpen(false);
+  }
+
+  return (
+    <div>
+      {isOpen && 
+        (<div className='position-relative'>
```
hanbyul-here, line 29: I really liked how I don't need any custom styles with built-in utility classes from uswds. However, then we need to add the line here: https://github.com/NASA-IMPACT/veda-ui/pull/1031/files#diff-8890a38014798c861bfe6ddab91bf46ec3c56d330831731a47453f7a1e7bce72R3 which makes the output css huge. (>400kb) So I went back to manual safelists with purgecss: https://github.com/NASA-IMPACT/veda-ui/pull/1031/files#diff-05ed1c7f99e45b485708115502a1e0f28c8547e9a9a29c1b8664f79103cf7873R12

app/scripts/main.tsx
```diff
@@ -3,7 +3,7 @@ import { createRoot } from 'react-dom/client';
 import T from 'prop-types';
 import { BrowserRouter, Route, Routes, useLocation } from 'react-router-dom';
 
-import '@trussworks/react-uswds/lib/index.css';
+// import '@trussworks/react-uswds/lib/index.css';
```
hanbyul-here, line 6: I think I read frmo uswds-react docs that we need to include this, but I don't really see the point of including this. @dzole0311 do you have more context?

postcss.config.js
```diff
@@ -4,5 +4,15 @@ module.exports = {
   plugins: [
     require('autoprefixer'),
     require('postcss-import'),
+    require('@fullhuman/postcss-purgecss')({
```
hanbyul-here, line 7: The styles doesn't get hot-reloaded unfortunately :[ (I have to restart the dev server if I want to include a new class name etc.)

app/scripts/components/common/banner/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <USWDSBanner aria-label={appTitle} className={type !== infoTypeFlag? 'bg-secondary-lighter': ''}>
```
hanbyul-here, line 43: Floagging that this component uses uswds default color theme now: https://designsystem.digital.gov/design-tokens/color/theme-tokens/ 

I don't want to drag this PR with another configuration setup especially considering there hasn't been a request for a different type of banner. (But then it might be better to just get rid of this special styling for now?)

app/scripts/main.tsx
dzole0311: I was in-between commenting & uncommenting it myself in my PRs. Let's keep it commented or we remove it until we see a need for it.

app/scripts/components/common/banner/index.tsx
dzole0311: Until needed I'm in support of the default theme 👍

app/scripts/components/common/banner/index.tsx
dzole0311: Yeah, an alternative to html util classes for more complicated components that works without bloating the CSS would be to use the tokens via mixins, functions or variables in the .scss files. Something like this:

```
// import path to the wanted mixin package (layout mixin, typography mixin, etc)

.some-new-class {
    @include u-padding-x(2);
}
```

But no need to complicate it here:)

dzole0311:
@hanbyul-here could you add steps on how to configure the banner to test this PR? Thanks! 🙇

sandrahoang686:
Hmm this is more of a thought pattern of how we may want to use USWDS components within our codebase. My thought is it might be a good pattern to create a wrapper for all USWDS components we use in our codebase and then forward props... this way it creates a pattern where it can be easily decoupled from breaking changes that the DS could introduce later on in the future (this may or may not happen but we'd be easily prepared)... and then just also it creates a decoupling layer in general.  If there are any changes that happen to the DS component, it would also isolate us to just having a make a change in one place (the wrapper) instead of throughout our codebase, if that makes sense.

What this would look like is having a Design System directory for USWDS with our wrappers around these components. Then other components in our codebase would reference our internal directory. `common/banner/index.tsx` isn't too complicated and is already pretty close to this pattern since it doesn't add much functionality on top, so its mostly already a simple wrapper.

What do you think?

hanbyul-here:
@dzole0311 No wonder you couldn't see the banner—I configured it wrong 😓 . You should be able to see it now. I also added a document about adding it through the configuration file.

@sandrahoang686 I never had experience with the pattern that you suggested. I tried to educate myself a bit, and some people like the pattern because of the pros you mentioned, and others caution that it may introduce additional complexity and maintenance requirements. It is difficult for me to form an opinion on this issue without hands-on experience. Since we are changing our design system, do you mind giving me a concrete example of how we could have benefited from the pattern if we used it for the devseed ui library?

sandrahoang686:
@hanbyul-here sure, you already are implementing the pattern actually 😅. What you did was wrap the DS `USWDSBanner` element with our own custom wrapper without applying much additional complex logic on top of it, a simple enough wrapper. Our components or `layout-root` ([here](https://github.com/NASA-IMPACT/veda-ui/pull/1031/files#diff-ccbefb6a7f3470de8911afb077446ff62bdc3f0e114b32d1c75490fda4debb5dR9)) references this wrapper instead of directly importing the `USWDSBanner` element. This created a nice abstraction layer where if there are breaking changes to the DS then we can just manage it in our wrappers creating isolation. For example, if DS rolls out breaking changes to some of their ui elements - like banner is deprecated (random example), we just have to update our wrapper file with a new banner component instead of touching every file in our codebase that imports from uswds directly. Another example is when props change..

I'm just thinking though, does it make sense to create a directory `/design-system/components` where these wrappers can live instead so its more explicit and we keep to this pattern, never importing from uswds directly in our own components. Its cleaner, creates a nice easy to manage abstraction layer, and is quick to decouple. There isn't much overhead risk.

dzole0311 (approved):
Looks great, thanks for adding the first USWDS component in the Dashboard 🥳 

To follow up on our discussion from the other day, I'm in favor of a simple structure like the one used in the PR, but I’m also open to the idea of nesting the components in an additional subdirectory (as @sandrahoang686 suggested). It’s a minor adjustment, so I'll leave the decision to you.

hanbyul-here:
@dzole0311 @sandrahoang686  Another note that I made purgecss not to kick in when `NODE_ENV` is `development ` - I am aware that this might have a problem that what you see on development is not what you get from the preview build. But purgecss makes dev server quite slow, so I decided to take this route for now. I am really open to changing this setup anytime with alternative approaches.

sandrahoang686:
"simple structure" is opinion based as both are simple.. its just organization and points of reference in the code. When there is a directory we can follow that has all the USWDS or whatever DS elements with simple wrappers its easy to pinpoint, follow patterns, and quicker to decouple. While having it blended with other common component just doesn't provide that clear distinction and pattern standard. But most important pattern is there which is not using USWDS directly in our components 👍🏼 the organization and logic looks good to me!

Issue #840, opened 2024-02-13
[Data Catalog] Overview page redesign to accommodate many data types and interactions


faustoperez:
### Issue
To make it more flexible and future proof, the data catalog overview page should accommodate different data types and link to different tools in the platform + incorporate more advanced functionality.

#### User story
A researcher wants to get authoritative information about how greenhouse gas emissions from land evolved over the last decade, to get a first indication for a hypothesis they have.

They find a relevant dataset (or concept page / story).

They can see that there are 7 different options to browse, study, or analyze the dataset(s):

- A quick browser for the specific layer (e.g. current Exploration interface, integrated with catalog or so)
- A raster and zonal statistics time series explorer (E&A)
- A link into viewing this data in QGIS
- A link into loading this data in Python/R in JupyterHub
- A link to an external tool from EPA (or other agency) that presents the same or similar data
- An interface for downloading specific files directly
- API docs for loading the data from somewhere else

The user can see that the tools differ in complexity and kind of study they support. After a quick browse in the Exploration interface, they decide to open the E&A interface.

They take note that there are many other options and the GHG Center promotes access to the data directly in the cloud and web browser, which may become relevant as they deepen their research.

### Acceptance criteria

- [x] List the different data types it should support (i.e. raster time series, plume-type, 3D, tabular, etc)
- [x] Gather feature requirements

aboydnw:
There are a few different things to consider for this redesign. Some of this is already in the original comment, just jotting down my thoughts here.

### Data Types
There will be lots of different types of data, that might require different tools for visualization and analysis. These include:
- L3 (and below) raster and time-series-based data
- L2 (and up) raw imagery and swaths of images
- Sub-daily model outputs, such as weather or fire data
- "Instances" or "events" such as methane plumes, earthquakes, or volcanic eruptions

### Connected Tools
- Map visualization and zonal statistics
- QGIS or ArcGIS 
- VEDA's connected JupyterHub environment
- Dataset specific tools already created by the data providers
- Related data stories in VEDA

### Additional Information/Resources for Consideration
- API documentation
- Downloadable files
- Simple visualization for that specific dataset
- Description of the dataset, including examples and references
- Dataset metadata

aboydnw:
@j08lue when you get the chance, could you review the list above and double check if we are missing anything?

[...]

PR #1032, opened 2024-07-03
Ab/titiler cmr 2

abarciauskas-bgse:
**Related Ticket:** #1000 and #1001 

### Description of Changes
The goal of this PR is to deduplicate common functionality for raster layers. This was inspired by @hanbyul-here comment on #1000 

>We now have more types of layers than we initially planned. They have different ways of getting the data they need, but eventually, most of them are displayed as a raster layer in Mapbox. I was wondering if it would be better to separate out the layer for drawing so it can be used throughout the layers and make each different layer pass the information that the raster layer needs (tile URL and parameters).

The current raster data type is using a more complex RasterTimeseries component than the other data types, but we are able, in this PR to change it enough so the other layer types can use it. With the addition of titiler-cmr, it makes sense to consolidate the code for adding a raster layer into one component (which is currently still `RasterTimeSeries`.

This PR:
* Creates a hook for each dataset type to return a tileUrls object which includes, optionally, a tileJsonUrl (although this should be required), a wmtsTileUrl and tileServerUrl. The latter 2 URLs are being used in the metadata for raster layers. I am returning the entire tileJsonUrl with parameters from the `useDatasetTiles` hooks because the tileJsonUrl endpoint for the mosaic is generated via POST register mosaic request and the tile parameters formulation is also different than the other layers.
* Also creates a hook for the stacCollection generation, since this code is also only for the raster/pgstac-supported layers.
* Moves the changeStatus function to the layer component, so it can be used in the hooks.
* Modifies the conditional component logic in layer.tsx so all layer types apart from vector use the RasterTimeseries component.

* Just generally, does this approach make sense.
* The `raster` layer type now seems like a too generic name as there are 4 types which qualify as raster layer types. Could we call this the `mosaic` (or `raster-mosaic`) layer type? Another characteristic of this type is that we are using the VEDA endpoints so these datasets are presumably in the database and thus can be used by titiler-pgstac.
* I removed the functionality to support these layers in the existing `data-catalog/{collection.id}/explore` pages to reduce effort and changes because I got the impression from @sandrahoang686 that those pages would be removed in favor of the /exploration interface. Should I add them back?

### Validation / Testing
* Tested all layer types locally
* Was not able to get the facebook population density layer to work, but no2 and geoglam layers worked, along with the zarr, cmr-stac and titiler-cmr layers
* Will test deploy preview when it's ready


sandrahoang686:
> I removed the functionality to support these layers in the existing data-catalog/{collection.id}/explore pages to reduce effort and changes because I got the impression from @sandrahoang686 that those pages would be removed in favor of the /exploration interface. Should I add them back?

There were talks and designs about combining the datasets overview and /explore page to a single view so separate from E&A (ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/840)). I'm unsure when that work would start though as I dont think we scoped it for any time soon

abarciauskas-bgse:
🤔 @sandrahoang686 👌🏽 if it's not scoped to happen soon I should probably try and re-integrate these dataset types into the `data-catalog/{collection.id}/explore` pages

abarciauskas-bgse:
But before I add back support for those dataset types in those collection-specific pages, I want validate this approach is a good one, hoping someone more familiar with react and the veda-ui codebase can help me with that.

app/scripts/components/common/map/style-generators/hooks.ts
```diff
@@ -106,8 +204,187 @@ export function useCMR({ id, stacCol, stacApiEndpointToUse, date, assetUrlReplac
     return () => {
       controller.abort();
     };
-  }, [id, stacCol, stacApiEndpointToUse, date, assetUrlReplacements, stacApiEndpoint, onStatusChange]);
+  }, [id, stacCol, tileApiEndpointToUse, stacApiEndpointToUse, date, onStatusChange, sourceParams]);
+
+  return tileUrls;
+};
+
+export function useMosaic(options: UseDatasetTilesOptions) {
```
sandrahoang686, line 212: Nice! 👏🏼 Great idea on moving `mosaic` and `stacCollection` logic out of raster-timeseries file and into their own hooks here to be reused. Also cleans up that file a ton...

sandrahoang686 (commented):
Hey Aimee, nice job on this PR! In my opinion moving some of the logic that originally lived in the `raster-timeseries` component out into their own hooks and making `changeStatus` a callback function prop being passed in gives control/flexibility to the component using the hook was a good move. @hanbyul-here may have other thoughts as well.

Created a PR to this branch to help clean up to some of the comments I made which is hopefully helpful https://github.com/NASA-IMPACT/veda-ui/pull/1035

hanbyul-here:
Thanks for clarifying the priority @abarciauskas-bgse  the workplan call sounds good. I will set up something for the next week 🙇

Issue #1038, opened 2024-07-09
[Bug] Sorting should be done alphabetically by default in the Catalog Content


dzole0311:
### Context

URL: 

Blocking a release? (y/n): n

Regression issue? (y/n): 
[If yes, add Regression label]

Priority: 

### Steps to reproduce the issue

1. Visit the Data Catalog
2. See that the catalog items are not sorted by the name field

PR #1039, opened 2024-07-09
Sort catalog items alphabetically in the E&A modal and the data catalog

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1038

### Description of Changes
Add fix that sorts catalog content alphabetically by default for the Data Catalog view and the E&A modal. This behavior was there for the Data Catalog before the last release, but not for the old E&A modal (due to it using horizontal cards).

### Validation / Testing
1. The filtering, sorting and search work as before for the Data Catalog and the E&A modal

### DIFF
--- a/app/scripts/components/common/catalog/catalog-content.tsx
+++ b/app/scripts/components/common/catalog/catalog-content.tsx
@@ -37,6 +37,7 @@ export interface CatalogContentProps {
 }
 
 const DEFAULT_SORT_OPTION = 'asc';
+const DEFAULT_SORT_FIELD = 'name';
 
 export const findParentDataset = (layerId: string, datasets) => {
   const parentDataset: DatasetData | undefined = Object.values(datasets).find((dataset: DatasetData) =>
@@ -82,7 +83,7 @@ function CatalogContent({
     prepareDatasets(allDatasetsWithEnhancedLayers, {
     search,
     taxonomies,
-    sortField: DEFAULT_SORT_OPTION,
+    sortField: DEFAULT_SORT_FIELD,
     sortDir: DEFAULT_SORT_OPTION,
     filterLayers: filterLayers ?? false
   }));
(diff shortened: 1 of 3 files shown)


app/scripts/components/common/catalog/catalog-content.tsx
```diff
@@ -82,7 +83,7 @@ function CatalogContent({
     prepareDatasets(allDatasetsWithEnhancedLayers, {
     search,
     taxonomies,
-    sortField: DEFAULT_SORT_OPTION,
+    sortField: DEFAULT_SORT_FIELD,
     sortDir: DEFAULT_SORT_OPTION,
     filterLayers: filterLayers ?? false
   }));
```
sandrahoang686, line 89: Lets type sortDir in prepareDatasets to be `'asc' | 'desc'` and if you know the options sortField could be too, would be great to type those. I know we could do by name and pubDate

app/scripts/components/common/catalog/prepare-datasets.ts
```diff
@@ -80,6 +80,15 @@ export function prepareDatasets (
...
     });
 
+  if (filterLayers && sortField) {
+    filtered = filtered.map((d) => ({
+      ...d,
+      layers:
+        isDatasetData(d) &&
+        d.layers.sort((a, b) => a[sortField]?.localeCompare(b[sortField]) || 0)
+    })) as DatasetData[];
+  }
```
sandrahoang686, line 90: I think https://github.com/NASA-IMPACT/veda-ui/blob/03d0488ce02a0fc474a5deb16df93828056f6e6e/app/scripts/components/common/catalog/prepare-datasets.ts#L75-L81 should already work?

dzole0311: The problem is that this will only work for the Data Catalog, where we show the datasets. But in the E&A Modal we show the layers within those datasets, so we have to go one level down (to the layers prop) to sort them.

PR #1040, opened 2024-07-09
Fix sortField bug

sandrahoang686:
**Related Ticket:** [_{link related ticket here}_
](https://github.com/NASA-IMPACT/veda-ui/issues/1038)

https://github.com/NASA-IMPACT/veda-ui/issues/1038

### Description of Changes
set `sortField` to default to `'name'`

### Validation / Testing

### DIFF
--- a/app/scripts/components/common/catalog/prepare-datasets.ts
+++ b/app/scripts/components/common/catalog/prepare-datasets.ts
@@ -7,11 +7,13 @@ const isDatasetData = (data: DatasetData | StoryData): data is DatasetData => {
   return 'layers' in data;
 };
 
+type SortOptions = 'asc' | 'desc';
+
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/catalog/prepare-datasets.ts
```diff
@@ -21,7 +23,7 @@ export function prepareDatasets (
   data: DatasetData[] | StoryData[],
   options: FilterOptionsType
 ) {
-  const { sortField, sortDir, search, taxonomies, filterLayers } = options;
+  const { sortField = 'name', sortDir, search, taxonomies, filterLayers } = options;
```
sandrahoang686, line 26: we should really type `sortField` but unsure of all strings it could be, should default to 'name' though

Issue #990, opened 2024-06-04
[E & A] Date picker aligns with temporal density


aboydnw:
## Context

Deborah received some feedback from scientists while they were playing with the new exploration and analysis page. One comment they had was, when they are viewing an annual dataset, the date shouldn't be specific to the day. "We're not viewing October 8th, we're viewing 2021." This is potentially confusing.

My assumption is the data picker should default to the lowest common denominator, since we can show datasets of multiple temporal densities on the page. So, I wrote the AC below to match that. I intentionally left out sub-daily temporal densities, we can adress that in a different ticket.

## Acceptance Criteria
- [ ] Given I am only viewing an annual dataset, then the date picker only allows for selecting of years
- [ ] Given I am viewing layers with temporal densities no finer than monthly, then the date picker only allows for selecting of months
- [ ] Given I am viewing layers with daily temporal densities, then the date picker allows for selecting of days
- [ ] I can only select dates in the date picker that are included in the temporal range(s) of the dataset(s) selected

aboydnw:
Potentially relates to #979 if we are assessing the new design system, could check for feasibility of this ticket

faustoperez:
> My assumption is the data picker should default to the lowest common denominator, since we can show datasets of multiple temporal densities on the page

Agree. Self-assigning this.

j08lue:
You cannot avoid that the "master playhead" sits at an arbitrary point of a **period** (year, month, day), instead of in the center, when there is another layer with lower time resolution.

But maybe we can get a better indication, **which period is currently active**, even if the playhead sits somewhere in between?

Maybe something like highlighting the active period or so? Just an idea...

![image](https://github.com/NASA-IMPACT/veda-ui/assets/3404817/83902a02-27c6-400b-8584-0c0395166765)

faustoperez:
> But maybe we can get a better indication, which period is currently active, even if the playhead sits somewhere in between?

I like that idea Jonas!

j08lue:
I see this ticket was about the date picker and not the playhead or the boxes on the timeline.

Did you consider, though, also implementing some visual cue for the user to see which period is active, as I suggested above? Or did you drop / postpone that? Just curious...

Issue #979, opened 2024-05-28
[E & A] Update Date Picker


aboydnw:
## Context

This is the implementation ticket of #833  It could also be a good ticket to test out the new design system.

## Acceptance Criteria

- [x] Date picker uses new design system
- [ ] Timelines sliders and playheads function to match designs
- [ ] Users are able to compare different dates in the exploration view

PR #1044, opened 2024-07-10
E&A Date Picker (Part One) - Integration and usage within the E&A Timeline header 

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/979

### Description of Changes
The PR integrates the USWDS date picker into the E&A timeline. To reduce code review complexity, the changes here are:

1. Integration of the react-uswds DatePicker in the veda-ui
2. Replacement of the current timeline DatePicker with the new one
3. Layout changes of the timeline header to match the designs (ie. placement of the datepickers)

Next:

1. Changes for the timeline playheads + fix for the date picker to expand up/down based on the screen space
2. Date picker aligns with the temporal density of the dataset(s): https://github.com/NASA-IMPACT/veda-ui/issues/990

I discovered a few limitations of the DatePicker from react-uswds while working on this. For example, if we want to toggle the Calendar via a custom trigger or set a different Calendar mode (day, month, year) based on the temporal density of the datasets, we will have to do it programmatically. The library currently doesn't expose ref, nor props / methods to have more control over the component.  

### Validation / Testing


dzole0311:
> can we align the selectable dates with the ones available on the active datasets:

@faustoperez Yeah, I will work on that as we discussed but via this issue: https://github.com/NASA-IMPACT/veda-ui/issues/990. Tried to do it here, but I realized that I might introduce scope creep if I pull it in this sprint (and will make it hard to review the logic in one big PR).

app/scripts/components/common/datepicker/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const SimpleDatePicker = ({ disabled, tipContent, onConfirm, id, triggerHeadReference, selectedDay, renderTriggerElement }: SimpleDatePickerProps) => {
+    const [showCalendar, setShowCalendar] = useState(true);
+    const datePickerRef = useRef<HTMLDivElement>(null);
+
+    const toggleCalendar = () => {
+      // We have no reference to the internals of the react-uswds DatePicker, so we have to query for it's trigger.
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      if (datePickerRef.current) {
+        const button = datePickerRef.current.querySelector('button.usa-date-picker__button')! as HTMLButtonElement;
```
hanbyul-here, line 31: Ah I see. This is a sticky problem 😬 . Can we control the visibility of the calendar in this way? If we do a conditional render like `{showCalendar && <DatePicker...>}`, we will lose the access to this button. (Calendar is not toggling right now, it just re-renders when renderTriggerElement is pressed I think. )

hanbyul-here:
Thanks for making the PR in phases. I think this is a lesson for us that it is not very easy to pick only speicific parts of the components from USWDS. 

I wonder if what we need is just a [Calendar component](https://github.com/trussworks/react-uswds/blob/main/src/components/forms/DatePicker/Calendar.tsx) that the DatePicker component uses.  This component is not being exported from the USWDS React library, but it might be a simpler feature request for their end. I can't think of any better approach than how you are doing to control the DatePicker now, but I left a comment on the code about visibility control. 

There are some items that I noticed: 

1. When I turn on the compare mode, and then open the datepicker, the layout collapses: 
Screenshot 2024-07-11 at 2 32 10 PM

2. Do we need `moment` dependency? We are already using `date-fns` (https://date-fns.org/) throughout the app and I wonder if we need another date management library.

app/scripts/components/common/datepicker/index.tsx
dzole0311: We can't unfortunately, since the `<Calendar />` comes coupled with the `<DatePicker />` and we can't conditionally render only that component. So the only thing I could think of is programmatically toggling it via the `usa-date-picker__button`. 

This gets trickier when we will have to show the `MONTH_PICKER` or the `YEAR_PICKER` based on the dataset time density, because these calendar modes exist as a local state in the Calendar. So the only solution I can think of is to click the picker button to open the calendar, then query and click either `usa-date-picker__calendar__month-selection` or `usa-date-picker__calendar__year-selection` to toggle the corresponding calendar mode depending on the time density 😰 

I opened two issues on `react-uswds` side:

- https://github.com/trussworks/react-uswds/issues/2972
- https://github.com/trussworks/react-uswds/issues/2973

dzole0311:
> I wonder if what we need is just a [Calendar component](https://github.com/trussworks/react-uswds/blob/main/src/components/forms/DatePicker/Calendar.tsx) that the DatePicker component uses

@hanbyul-here Yeah, I think so as well.

> When I turn on the compare mode, and then open the datepicker, the layout collapses:

Thanks for spotting this, will fix.

> Do we need moment dependency

Replaced with `date-fns`

app/scripts/components/common/datepicker/index.tsx
hanbyul-here: Throwing some ideas here. Can it be something that we can solve by overwriting the style? I feel like they (calendar trigger element for both uswds date picker and our date picker) are eventually doing the same thing, showing the date, and then a button to trigger the calendar. (I understand our original design wants the whole thing to be a button to trigger - including the text- but maybe we can adjust the design a bit to accommodate USWDS system?)

ex. overwriting the text input style, calendar icon (to show the caret) here? 

Screenshot 2024-07-12 at 10 57 42 AM



Screenshot 2024-07-12 at 10 57 48 AM

dzole0311: Thanks for the idea @hanbyul-here. As you pointed out, we will need to do a lot of manual triggering if we want to sync the calendar with the time density of the dataset, which I think is very error prone. 

> is this a good component to write React components for our own just using uswds style

Yes, I think so. We need a more robust approach but at the same time not to reinvent the wheel (e.g. implementing basic date picker functionality from scratch). So I tried using `react-calendar` (https://www.npmjs.com/package/react-calendar)

Here's a PR with it: https://github.com/NASA-IMPACT/veda-ui/pull/1048 and [here's the deployment](https://deploy-preview-1048--veda-ui.netlify.app/exploration?datasets=%5B%7B%22id%22%3A%22GPM_3IMERGDF.v07%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2024-07-14T22%3A00%3A00.000Z): 

We could override some of the styles using uswds mixins / functions: https://github.com/NASA-IMPACT/veda-ui/pull/1048/files#diff-88b2701907396a34bbe38997d6970f2f6824389dbefe1dcc112a5fdfaa35d811R24

Also:

- It works well with Tippy
- It has out-of-the-box config for showing only the month or year calendar view (it's the `maxDetail`), so it will save us a lot of time

What do you think about this?

hanbyul-here: Yeh, I Agree that we should not invent the wheel. I will vote for using ReactCalendar for now 👍

dzole0311:
Closing this one as work continued here: https://github.com/NASA-IMPACT/veda-ui/pull/1048

Issue #1002, opened 2024-06-13
Implement Data Exposure Store in nextJs Prototype and for veda-ui MapBlock component


sandrahoang686:
As a result from spike ticket https://github.com/NASA-IMPACT/veda-ui/issues/978. This is the implementation ticket to implement the data exposure store in the nextJs prototype and showcase the pattern with what it looks like to replace the veda faux modules starting with the MapBlock mdx component.

**Acceptance Criteria:**
- [ ] Implement custom context provider store
- [ ] Get MapBlock component working in nextJs component, we would need static content that uses the MapBlock component (do this first and get it working as is)
- [ ] Update MapBlock component in veda-ui to take in datasets as a prop and remove faux modules as dependency from the component itself, create a wrapper container like done for data-catalog

**Bonus:**
- [ ] Get most if not all MDX components working when rendering custom mdx from mdx files, there seems to be an error with the Block component currently.. This way we can see the prototype rendering content correctly

aboydnw:
Dropping a loom video from @sandrahoang686 here

https://www.loom.com/share/c177d7c437094b828a5cc2efe37b6150?sid=2cd102d8-50e5-4bfe-8eac-c68319899d84

PR #1046, opened 2024-07-11
Abstract Away Faux Modules from MapBlock Component and make renderable for NextJs Instance Prototype

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1002

**These changes are RELATED TO THIS PR:** https://github.com/developmentseed/next-veda-ui/pull/1

### Description of Changes
This makes the MapBlock component renderable by the NextJs instance by  abstracting away the veda faux modules dependencies. This also exports the themeProvider and reactQueryProvider to be used by the nextJs instance.

I've tried to keep these changes quite isolated to the problem we are trying to solve so its not too big of a PR.

* Should NOT BE BREAKING CHANGES. Should be able to merge without care of order of NextJs PR
* See my comments in the code

### Validation / Testing
* Regular build should work
* Library build should work
* The stories and mapblocks in the stories should render as usual in the deploy preview

### DIFF
--- a/app/scripts/components/analysis/results/timeseries-data.ts
+++ b/app/scripts/components/analysis/results/timeseries-data.ts
@@ -1,4 +1,3 @@
-import * as process from 'process';
 import axios, { AxiosRequestConfig } from 'axios';
 import { QueryClient } from '@tanstack/react-query';
 import { FeatureCollection, Polygon } from 'geojson';
--- a/app/scripts/components/analysis/results/timeseries-data.ts
+++ b/app/scripts/components/analysis/results/timeseries-data.ts
@@ -1,12 +1,12 @@
+import * as process from 'process';
 import axios, { AxiosRequestConfig } from 'axios';
 import { QueryClient } from '@tanstack/react-query';
 import { FeatureCollection, Polygon } from 'geojson';
-import { DatasetLayer } from 'veda';
-
 import { MAX_QUERY_NUM } from '../constants';
 import { getFilterPayload, combineFeatureCollection } from '../utils';
-import EventEmitter from './mini-events';
 import { ConcurrencyManager, ConcurrencyManagerInstance } from './concurrency';
+import EventEmitter from './mini-events';
+import { DatasetLayer } from '$types/veda';
 import { TimeDensity } from '$context/layer-data';
 
 export const TIMESERIES_DATA_BASE_ID = 'analysis';
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -1,7 +1,8 @@
 import React, { useMemo, useState, useEffect } from 'react';
 import styled from 'styled-components';
 // @NOTE: This should be replaced by types/veda once the changes are consolidated
-import { ProjectionOptions } from 'veda';
+// import { ProjectionOptions } from 'veda';
+import { ProjectionOptions } from '$types/veda';
 import { MapboxOptions } from 'mapbox-gl';
 import {
   convertProjectionToMapbox,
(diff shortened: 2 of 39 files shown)


app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -361,10 +373,18 @@ function MapBlock(props: MapBlockProps) {
...
-        )}
+        )} */}
       </Map>
     </Carto>
   );
 }
 
+export function MapBlockWithProvider(props: MapBlockProps) {
+  return (
+    <ReactQueryProvider>
```
hanbyul-here, line 384: I know you're actively working on this branch right now, but I wanted to share some thoughts in case they might be helpful.


If I understand React Query correctly, the provider allows us to cache data so we don't have to fetch the same information repeatedly. If each map is wrapped with its own separate Provider, however, this cached data won't be shared across maps. 

One idea could be to export all the providers that UI requires from index.ts, and provide guidance for users on how to wrap their root layout with the necessary Providers. We already do something similar for ThemeProvider in our UI library (see [link](https://github.com/NASA-IMPACT/veda-ui/pull/1046/files#diff-df6782dfe8efb20003aa046424999792c0fdfa4472bc6b8f86a487b344ee0f6cR18)).

hanbyul-here:
Thanks for putting various works on making the component work with the NEXT instance. I could see how it was unavoidable to make multiple changes when making changes for the component library. Once we are clear on what needs to be done, can we make the changes more gradually? I see there are several kinds of changes here. 
Some changes I could spot are

1. Make the types come from non-faux module
2. Make the utility functions not to depend on faux-module (ex. additional parameters added for utility functions to accept the list of the datasets)
3. Make the change/adjustments for the components needed to be used in the library context.

How does this sound?

app/scripts/components/common/map/utils.ts
```diff
@@ -1,17 +1,17 @@
 import axios, { Method } from 'axios';
-import { format } from 'date-fns';
+import format from 'date-fns/format';
 import { Map as MapboxMap } from 'mapbox-gl';
 import { MapRef } from 'react-map-gl';
-import { endOfDay, startOfDay } from 'date-fns';
+import startOfDay from 'date-fns/startOfDay';
+import endOfDay from 'date-fns/endOfDay';
```
sandrahoang686, line 6: Parcel build for library was unable to find there date-fns if it wasn't imported like this... So had to switch over to doing this. Seems to be an open issue with Parcel
https://github.com/parcel-bundler/parcel/issues/9676

package.json
```diff
@@ -14,7 +14,6 @@
   "source": "./app/scripts/index.ts",
   "main": "lib/main.js",
   "module": "lib/module.js",
-  "types": "lib/index.d.ts",
```
sandrahoang686, line 17: I had to get rid of this.. otherwise the build for the library would keep failing with some `@parcel-transformer-typescript-types` error on very random types. This was a headache. This is probably so parcel can compile the types to export by extracting the library's types and generating a `.d.ts` file but I think we can do without it. Any concerns or unforeseen risks if we do this?

dzole0311: I might be wrong, but the first concern that comes to my mind is not having any TS declarations. So any project that would use our veda-ui will not have type information and will treat all imports from our package as having an `any` type.

I've also seen some TS errors on the main branch when running the `buildlib`, but the build succeeds there at the end. Maybe we could run the lib build command between the `main` branch and this one to narrow down which error fails the build?

hanbyul-here:
👋 

I could narrow down what component leads to a build failure by commenting out the components we are exporting gradually. I could check that the build only fails when `MapBlock` component is used. (Parcel still has some complaints about ts but still builds the library.)

When we look at the error log with Mapblock, there are several kinds of errors. To point out the most suspicious one:

`@parcel/transformer-typescript-types: Cannot find module 'veda' or its corresponding type declarations.` - Is there any utility function that is still using 'veda' module? We unfortunately check the dependency of the dependency recursively. ex. BlockMap is using types from `exploration/types.d.ts.` : https://github.com/NASA-IMPACT/veda-ui/pull/1046/files#diff-b93e783f2167705031fb25f815ba97972896110f71e42d83e32fd6e557d75bccR34 and that files still has veda faux module as a dependency: https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/types.d.ts.ts#L1

I suggest reviewing BlockMap's dependencies to ensure it is completely independent from the faux module. Then, we can see if other errors still interrupt the library build. 

Also, it might be worth merging the main branch since some major updates of Parcel went into the main recently!

app/scripts/components/analysis/results/timeseries-data.ts
```diff
@@ -1,12 +1,12 @@
+import * as process from 'process';
```
sandrahoang686, line 1: having to import this to get rid of parcel-typescript-types error

hanbyul-here (approved):
Yay, thanks for the work 🙇  

I could also export `CatalogView` component without problem on this branch. I am excited for the next step!

PR #1048, opened 2024-07-15
E&A Date Picker (Part Two) - react-calendar with USWDS tokens 

dzole0311:

### Description of Changes
We're introducing a more flexible calendar component which will use some USWDS design tokens and icons. The change is made based on this discussion: https://github.com/NASA-IMPACT/veda-ui/pull/1044#discussion_r1677783700

### Validation / Testing


app/scripts/components/common/datepicker/index.scss
```diff
@@ -0,0 +1,52 @@
...
+    background-color: utils.color('primary-vivid') !important;
+    color: utils.color('base-lightest');
+  }
+
+  &__month-view__days__day--weekend {
+    color: utils.color('gray-90') !important;
+  }
+}
+
+.tippy-box {
```
hanbyul-here, line 28: (I know this pr is just poc, but for my curiosity) - Would this style be applied globally?

app/scripts/styles/theme.ts
```diff
@@ -79,6 +80,7 @@ export const GlobalStyles = createGlobalStyle`
 
   :root {
     --base-space-multiplier: ${themeVal('layout.glspMultiplier.xsmall')};
+    --base-font-family: ${themeVal('type.base.fontFamily')};
```
hanbyul-here, line 83: Is there any way to set this up through uswds scss configuration?

app/scripts/components/common/datepicker/index.scss
dzole0311: Thanks for spotting this, you're right it will be applied globally. Fixed to apply to the scope only.

app/scripts/components/common/datepicker/index.scss
```diff
@@ -0,0 +1,77 @@
+// @NOTE: Optimize the import paths to the file
```
dzole0311, line 1: To optimize the import paths, I tried adding them in the sassrc, but for some reason the mixins were not found. I'll need to have another look in the next PRs.

app/scripts/styles/theme.ts
dzole0311: Thanks, I moved it to the USWDS config. We should however think about if (and how) to give the veda-ui lib users ability to override the base fonts which is the reason why I added it as a root variable initially. We don't have to do it here though as it's out of scope.

hanbyul-here: This is such a good point. And I think this might be something that we need to resolve before merge - because each instance has its font configuration - 
The typeface is Open Sans by default, and GHG is overwriting it as Inter - https://github.com/US-GHG-Center/veda-config-ghg/blob/develop/veda.config.js#L61

Moreover, when I check the style, the calendar is rendered with `Source Sans pro` - which makes me believe that we need to handle the type overwriting of the component.

sandrahoang686:
Still need to actually code review but wondering if we wanted to update the date pick in the `/analysis` route as well? EIC still uses that page/route
Screenshot 2024-07-16 at 10 29 47 AM

app/scripts/styles/theme.ts
dzole0311: Okay, I think I came up with a working solution and pushed the change. Thanks for having another look 🙇 

As we discussed yesterday, we will override the default typeface used by USWDS with the root base font used in our instances.

Test PR to showcase that USWDS will use the base font root variable set by the theme on the SiteAlert: https://deploy-preview-1051--veda-ui.netlify.app/

Screenshot 2024-07-17 at 14 29 28

Also going forward I think we should import all the packages we would like to use (e.g "usa-alert", "usa-button") in one SCSS entry file (in our case the `styles.scss`) [like this](https://github.com/NASA-IMPACT/veda-ui/pull/1051/files#diff-8890a38014798c861bfe6ddab91bf46ec3c56d330831731a47453f7a1e7bce72), and only use our custom scss files per component if we would like to override some specific uswds component stylings. Otherwise, the overriding of the typeface won't work. I did a short test and this doesn't seem to affect the Parcel build time nor the CSS size.

dzole0311:
> Still need to actually code review but wondering if we wanted to update the date picker in the `/analysis` route as well? EIC still uses that page/route 

Thanks for raising this point @sandrahoang686 ! As we discussed, we will only use the new DatePicker in the E&A.

app/scripts/components/common/datepicker/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+}: SimpleDatePickerProps) => {
+  const [isCalendarOpen, setIsCalendarOpen] = useState(false);
+
+  const handleDateChange = (date: Date) => {
+    onConfirm(date);
+    setIsCalendarOpen(false);
+  };
+
+  const handleTriggerClick = () => {
+    setIsCalendarOpen(!isCalendarOpen);
```
hanbyul-here, line 40: Toggling doesn't seem to work - the function gets called twice in a row.
I think it is because `onClickoutside` function also gets called when the triggerElement is clicked. can you take a look at this ?

app/scripts/components/common/datepicker/index.scss
```diff
@@ -0,0 +1,86 @@
...
+
+    &[data-placement^='right'] > .tippy-arrow::before {
+      border-right-color: transparent !important;
+    }
+
+    .tippy-content {
+      padding: 0 !important;
+    }
+  }
+}
```
sandrahoang686, line 86: ❓ : We are using a ton of `!important`s, are we trying to override some styles?

app/scripts/components/exploration/components/timeline/timeline-controls.tsx
```diff
@@ -102,149 +102,142 @@ export function TimelineControls(props: TimelineControlsProps) {
...
+                      const [, max] = xScaled.range();
+
+                      const currentX = xScaled(selectedDay);
+                      const nextX = currentX + DAY_SIZE_MAX;
+                      const newDate = xScaled.invert(
+                        nextX > max ? currentX - DAY_SIZE_MAX : nextX
+                      );
+
+                      setSelectedCompareDay(newDate);
+                    }}
```
sandrahoang686, line 132: Possible to separate move this into a useCallback fn? Dependencies of `xScaled`

app/scripts/components/common/datepicker/index.tsx
dzole0311: IMO I think we should keep the default uswds behavior (ie. clicking on the trigger should toggle the calendar)

app/scripts/components/common/datepicker/index.scss
dzole0311: Yes, overriding styles. An alternative would be to use more tags / classes for specificity.

app/scripts/styles/theme.ts
hanbyul-here: 🥳 Wooh amazing, thanks for making it work @dzole0311 
I agree that the packages need to be in `styles.scss`. What do you think about mixins and utils? Would it be better to have a centralized point to expose those? or is it better to import them on the component levels when needed.

aboydnw:
I tested this a bit too, looks good. Only thing to note is that I don't think we should allow people to select a date outside of the temporal ranges of the selected datasets. But it wasn't in the AC and I see we don't have this rule on production so I dropped that AC in this ticket and we can discuss later: https://github.com/NASA-IMPACT/veda-ui/issues/990

sandrahoang686:
Manually validated, similar to Anthony's concern, i'm able to select dates in the future. This matches behavior in prod though. Also, when expanding my screen the date picker moves up and blocks the date. It doesn't do this when it sits below the date ranges though, not sure if this is okay?
Screenshot 2024-07-17 at 12 37 23 PM

hanbyul-here (approved):
Thanks for working on this (and figuring out uswds related questions on the way)

Issue #1052, opened 2024-07-17
[E & A] Update the Timeline Playheads


dzole0311:
## Context

This is the second part of https://github.com/NASA-IMPACT/veda-ui/issues/833, and includes the updating of the Timeline Playheads.

## Acceptance Criteria

- [ ] Timelines sliders and playheads function to match designs
- [ ] When the date picker floats above the date, it is spaced to not cover up the date (description in [this comment](https://github.com/NASA-IMPACT/veda-ui/pull/1048#issuecomment-2233743446))

## Related Tickets

https://github.com/NASA-IMPACT/veda-ui/issues/979

aboydnw:
@faustoperez @dzole0311 posting a video on some thoughts relating to this ticket from testing out our recent earth.gov release. I'm not sure it's worth iterating on this at the moment but curious what you would think of this functionality

https://www.loom.com/share/f4c944de004e447baaf6c9e7f4ddd965?sid=b7997d67-7863-4d32-af87-bc36fdde481c

faustoperez:
Good catch @aboydnw! 

I think that a simple fix would be that when users change a date using the datepicker the timeline re-centers around that date, as they're being intentional with a selection. So:

- The datepicker is used to jump between dates
- The playheads are used to indicate the date active selection and fine-tune it by manually sliding them.

The same behaviour should happen with the analysis playheads, the timeline should re-center depending on the selection.

PR #1067, opened 2024-07-24
[E&A] Add logic for tracking out-of-view Timeline playheads

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1052

### Description of Changes

This PR continues the E&A timeline changes with the following updates:

1. Logic to track which timeline playheads are out-of-view and display static (non-clickable) indicators for them
2. Revamped playheads to simplify style control and changed their content to show the corresponding date and labels

Designs: https://www.figma.com/design/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?node-id=1128-2306&t=XQg28Qtma2dgRiS5-4

There are a couple of details from the designs that I did not include in this PR due to some questions I have. I would appreciate your feedback, especially since @faustoperez is unavailable:

1. The designs show the timeline header dates vertically centered, as shown below:

Screenshot 2024-07-24 at 15 08 39

Currently, our header dates are displayed on two lines, and we sometimes show longer text like "Sun 05 Apr 2009":

Screenshot 2024-07-24 at 15 16 37

Changing these to be vertically centered and in one line can sometimes result in a cluttered view when zooming, as shown below:

Screenshot 2024-07-24 at 13 48 52

2. Clicking on a playhead toggles the corresponding date picker. If you intend to drag the playhead and accidentally click on it, the calendar toggles, which can be a bit inconvenient. I plan to open a separate PR for this for some team feedback

### Validation / Testing

The validation / testing cases are outlined in the designs: https://www.figma.com/design/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?node-id=1128-2306&t=XQg28Qtma2dgRiS5-4


aboydnw (approved):
I clicked around for a while and couldn't find a way to break it. Looks good to me. And I think your 2 differences from the designs are fine for this iteration.

app/scripts/components/exploration/components/timeline/timeline-controls.tsx
```diff
@@ -87,8 +90,132 @@ export function TimelineDateAxis(props: Omit<TimelineControlsProps, "onZoom">) {
...
+
+const TimelineHeadLeftIndicators = styled.div`
+  position: absolute;
+  bottom: 0;
+  left: -8px;
+  display: flex;
+  gap: 10px;
+`;
+
+const getIndicators = (outOfViewHeads: TimelineHead[]) => {
```
sandrahoang686, line 175: 🤔 Is there a reason you are creating this like a hook fn instead of a functional component? its returning a JSX.Element and can accept `outOfViewHeads` as a prop. If we do it like this doesn't it make sense to place this inside the functional component and wrap it with a useCallback with `outOfViewHeads` as dep to prevent unnecessary recreation? I think with how its done right now, its going to be unnessarily called? correct me if i'm wrong

app/scripts/components/exploration/components/timeline/timeline-head.tsx
```diff
@@ -32,25 +20,136 @@ interface TimelineHeadBaseProps {
...
+const TimelinePlayheadLeft = styled(TimelinePlayheadBase)`
+  background-color: ${TIMELINE_PLAYHEAD_COLOR_PRIMARY};
+  border-top-left-radius: 4px;
+  border-top-right-radius: 4px;
+  border-bottom-left-radius: 4px;
+  border-bottom-right-radius: 0;
+  position: absolute;
+  transform: translateX(-100%);
+  white-space: nowrap;
+  left: 17px;
```
hanbyul-here, line 71: I am a bit concerned about this magic number. Why should this be 17px?

app/scripts/components/exploration/components/timeline/timeline-head.tsx
```diff
@@ -32,25 +20,136 @@ interface TimelineHeadBaseProps {
...
+    &::after {
+      border-top: 8px solid ${TIMELINE_PLAYHEAD_COLOR_PRIMARY};
+    }
+  }
+`;
+
+const TimelinePlayheadBase = styled.div`
+  background-color: ${TIMELINE_PLAYHEAD_COLOR_SECONDARY};
+  color: ${TIMELINE_PLAYHEAD_COLOR_TEXT};
+  padding: 3px;
```
hanbyul-here, line 56: Can we suse spacing functions (`glsp`) for a coherent look?

hanbyul-here (commented):
👏 Thanks for the amazing work! I don't think I have anything to comment about the logic. 
I am pretty sure a lot of hard-coded styles were inevitable for this type of component. but it would be nice to make the style less hardcoded, and coherent with the current dashboard. When you have a chance, do you mind going over the styles to check 1. if some of the magic numbers were needed 2.if any values (spacings, radius numbers..) can be replaced with the utility function from theme provider? - https://ui.ds.io/latest/?path=/docs/components-theme-provider--theme-explorer)

app/scripts/components/exploration/components/timeline/timeline-head.tsx
```diff
@@ -32,25 +20,136 @@ interface TimelineHeadBaseProps {
...
+  border-bottom-right-radius: 0;
+  position: absolute;
+  transform: translateX(-100%);
+  white-space: nowrap;
+  left: 17px;
+  min-width: 115px;
+  text-align: center;
+`;
+
+const TimelinePlayheadRight = styled(TimelinePlayheadBase)`
```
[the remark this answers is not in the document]
dzole0311: Okay, I think it should be better now...

app/scripts/components/exploration/components/timeline/timeline-head.tsx
dzole0311: I went through the styles and changed up the ones I could. Some magic numbers are there as we're implementing the playheads in the context of an svg canvas, part of which is not as responsive.

Some styles (e.g. the pseudoselector styling for :after, :before for the indicator triangles) I kept in 'px' units as I think it's easier to control them from a dev perspective, but let me know wdyt.

app/scripts/components/exploration/components/timeline/timeline-controls.tsx
dzole0311: Aside from the lowerCase naming, this is already a functional component:P But you're completely right we should make sure it doesn't re-render, so I re-factored it to make use of [memo](https://react.dev/reference/react/memo). Let me know what do you think.

hanbyul-here (approved):
I have one tiny suggestion to give a subtle box-shadow for playheads to increase the readability (I gave 1px 1px 1px rgba(255,255,255,0.3) as a value for the screenshot), but it looks pretty good as it is now so I will leave this up to you. Thanks again for the amazing work! 

Screenshot 2024-07-26 at 9 39 30 AM

dzole0311:
> I have one tiny suggestion to give a subtle box-shadow for playheads to increase the readability (I gave 1px 1px 1px rgba(255,255,255,0.3) as a value for the screenshot)

That's a good idea - thaks! I applied the change. In either case I'll run the whole implementation by @faustoperez once he's back, and we could make more adjustments as needed.

app/scripts/components/exploration/components/timeline/timeline-controls.tsx
sandrahoang686: hm technically it is not a functional component. Its just a function that returns JSX - `getIndicators`. Difference is functional components are something called like `<Indictors {...props} />` while function that returns JSX is called like `<>{getIndictors()}</>`

faustoperez:
Hi @dzole0311 👋 happy to review together

PR #1072, opened 2024-07-25
Exportable nav

hanbyul-here:
**Related Ticket:** #1071 

### Description of Changes
Exporting nav from UI side so it can be used to override its menu from GHG side. 
The code changes look like a lot, but most of them are just pieces moving around and making the components accept props.

This change checks the veda.config to see if there are any defined `navItems` and use it if there is one instead of the default one. There can be 4 types of nav item, which are `internalLink`, `externalLink`, `dropdown`, `modal`.  The details of each type will be documented. 

The current version does not mark the dropdown item as 'active' when its child is selected. 

### Validation / Testing

Instance pr: https://github.com/US-GHG-Center/veda-config-ghg/pull/457

- [x] remove test config from veda.config
- [ ] Add documentation : will do it after style polish


app/scripts/components/common/page-header/index.tsx
```diff
@@ -0,0 +1,347 @@
...
+    align-items: center;
+    gap: ${glsp(1.5)};
+  `}
+`;
+
+
+export const MODAL_TYPE = 'modal';
+export const INTERNAL_LINK_TYPE = 'internalLink';
+export const EXTERNAL_LINK_TYPE = 'externalLink';
+export const DROPDOWN_TYPE = 'dropdown';
```
dzole0311, line 233: An enum might be better for these in terms of type safety?

```
enum HeaderLinkTypes {
  MODAL = 'modal',
  INTERNAL_LINK = 'internalLink',
  EXTERNAL_LINK = 'externalLink',
  DROPDOWN = 'dropdown'
}
```

app/scripts/components/common/page-header/nav-menu-item.tsx
```diff
@@ -0,0 +1,154 @@
...
+  `}
+`;
+
+export const MODAL_TYPE = 'modal';
+export const INTERNAL_LINK_TYPE = 'internalLink';
+export const EXTERNAL_LINK_TYPE = 'externalLink';
+export const DROPDOWN_TYPE = 'dropdown';
+
+
+function ChildItem({ child, onClick }: { child: NavLinkItem, onClick?:() => void}) {
```
dzole0311, line 67: To make the component purpose clearer, how about renaming `ChildItem` to `NavSubItem` or something similar? It might better reflect its role and context in the navigation.

`function NavSubItem({ item, onClick }: { item: NavLinkItem, onClick?:() => void }) {`

app/scripts/components/common/page-header/nav-menu-item.tsx
```diff
@@ -0,0 +1,154 @@
...
+      color: inherit;
+      opacity: 0.64;
+    }
+  `}
+`;
+
+export const MODAL_TYPE = 'modal';
+export const INTERNAL_LINK_TYPE = 'internalLink';
+export const EXTERNAL_LINK_TYPE = 'externalLink';
+export const DROPDOWN_TYPE = 'dropdown';
```
dzole0311, line 64: We could reuse the ones from the `page-header/index.tsx`?

app/scripts/components/common/page-header/nav-menu-item.tsx
```diff
@@ -0,0 +1,154 @@
...
+        href={(rest as ExternalNavLink).href}
+      >
+        {title}
+      </GlobalMenuLink>
+      </li>
+
+    );
+  } else if (type === MODAL_TYPE) {
+    return (<li><GoogleForm title={title} src={(item as ModalNavLink).src} /></li>);
+  } else {// if (item.type === DROPDOWN_TYPE
```
dzole0311, line 120: Leftover comment

dzole0311 (commented):
Had some time and had a first look into the PR. Apologies if there are things you're still planning to change or modify.

app/scripts/components/common/page-header/index.tsx
hanbyul-here: I had a confusing time figuring out the type. Let me write down my thoughts. 

1. I want to have an explicit string literal value per each navlinkitem type like: https://github.com/NASA-IMPACT/veda-ui/pull/1072/files#diff-d3839ae268fbc273c2df81af1f599ed221b1456f9979d2861004161fa1dc63abR8 Because each type requires a clearly different object. 

2. I can't seem to use an enum value as a type. ex. I can't do 
export interface InternalNavLink {
  title: string;
  to: string;
  type: ENUM.internalLink // <= This will give an error.
}
I am guessing it is because enum object gets available at runtime? (https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime) 

3. I am not sure enum does the job better or not - when we need a clear string literal value per each type of nav item. 🤔 

Any thoughts?

dzole0311: Yeah, you're right it's a runtime object. But when I suggested the enum I thought about using it like `type: HeaderLinkTypes.INTERNAL_LINK` which gets compiled to the string literal 'internalLink'. It's generally good for IDEs in terms of autocompletion and finding references so I don't have to know all the header link types upfront, and it groups together the pool of strings.

dzole0311:
I noticed a strange vertical line on the right of the dropdown menu + the border-radius does not seem applied on the top right and bottom right corners of the dropdown. Is this as expected?

Screenshot 2024-08-08 at 15 05 14

hanbyul-here:
It is not really expected, but it is a style problem that we could not resolve from the past tech-debt: https://github.com/NASA-IMPACT/veda-ui/pull/691 😅 I will try to handle it in the next PR.

dzole0311:
> It is not really expected, but it is a style problem that we could not resolve from the past tech-debt: #691 😅 I will try to handle it in the next PR.

Ah, I think it's just the scrollbar showing if I'm not mistaken. Adding `overflow-y: auto;` on the `DropContent` would probably fix it, but let's address & test it in the next PR to avoid regressions in other places where it's used.

dzole0311 (approved):

Issue #1070, opened 2024-07-25
Update Raster API endpoints as part of new version breaking change


sandrahoang686:
Titiler-pgstac is being upgraded from v0.8 => v1.3 which includes breaking changes to two endpoints. We need to update these endpoints in veda-ui to the new changes. This should be merged around the same time that veda-backend merges their changes to production. (data-services [develop PR for reference](https://github.com/NASA-IMPACT/veda-backend/pull/398))

**Endpoints to update:**
* `/mosaic/{searchid}` => `/searches/{search_id}`
   * Any requests to this endpoint that previous referenced `searchid` should now be `id` (see [here](https://stac-utils.github.io/titiler-pgstac/latest/migrations/v1_migration/#searchid-id))

* `stac/` => `/collections/{collection_id}/items/{item_id}`

**Where:**
* raster-timeseries files

PR #1075, opened 2024-07-30
Update Raster API endpoints as part of new version breaking changes

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1070

⚠️ **This needs to be coordinated with veda BE team on when this can be merged. BE team needs to merge their changes into their prod before this can merge.**

### Description of Changes
Updating raster api (titiler) endpoints as part of new version breaking changes. This does not worry about integrating new endpoints.

* [VEDA BACKEND PR](https://github.com/NASA-IMPACT/veda-backend/pull/398) for ref
* [Titiler version migration guide](https://stac-utils.github.io/titiler-pgstac/latest/migrations/v1_migration/#searchid-id) for ref

### Relevant API Changes
[**yes**] rename mosaic/ endpoint to searches/
[**no**] for `/register` endpoints, update `searchId` to `id`
[**no**] a dd new /collections/{collection_id} endpoint
[**no**] add new /colorMaps endpoint
[**no**] rename stac/ endpoint to /collections/{collection_id}/items/{item_id}
[**no**] rename stac-alt/ endpoint to /alt/collections/{collection_id}/items/{item_id}

### Validation / Testing

### DIFF
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -274,7 +274,7 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
         /* eslint-enable no-console */
 
         const responseData = await requestQuickCache<any>({
-          url: `${tileApiEndpointToUse}/mosaic/register`,
+          url: `${tileApiEndpointToUse}/searches/register`,
           payload,
           controller
         });
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -274,13 +274,15 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
         /* eslint-enable no-console */
         
         let responseData;
-        let fallbackToOld = false;
+
         try {
           responseData = await requestQuickCache<any>({
             url: `${tileApiEndpointToUse}/searches/register`,
             payload,
             controller
           });
+          const mosaicUrl = responseData.links[1].href;
+          setMosaicUrl(mosaicUrl.replace('/{tileMatrixSetId}', '/WebMercatorQuad'));
         } catch (e) {
           // @NOTE: conditional logic TO BE REMOVED once new BE endpoints have moved to prod... Fallback on old request url if new endpoints error with nonexistance... 
           if (e.request) {
@@ -290,16 +292,14 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
               payload,
               controller
             });
-            fallbackToOld = true;
+      
+            const mosaicUrl = responseData.links[1].href;
+            setMosaicUrl(mosaicUrl);
+          } else {
+              LOG && 
+                /* eslint-disable-next-line no-console */
+                console.log('Titiler /register endpoint error', 'color: red;', e);
           }
-        }
-        let mosaicUrl = responseData.links[1].href;
-        
-        if (fallbackToOld) { // @NOTE:  conditional logic TO BE REMOVED once new BE endpoints have moved to prod...
-          setMosaicUrl(mosaicUrl);
-        } else {
-          mosaicUrl = mosaicUrl.replace('/{tileMatrixSetId}', '');
-          setMosaicUrl(mosaicUrl.replace('/{tileMatrixSetId}', ''));
         }
 
         /* eslint-disable no-console */
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -283,9 +283,9 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
           });
           const mosaicUrl = responseData.links[1].href;
           setMosaicUrl(mosaicUrl.replace('/{tileMatrixSetId}', '/WebMercatorQuad'));
-        } catch (e) {
+        } catch (error) {
           // @NOTE: conditional logic TO BE REMOVED once new BE endpoints have moved to prod... Fallback on old request url if new endpoints error with nonexistance... 
-          if (e.request) {
+          if (error.request) {
             // The request was made but no response was received
             responseData = await requestQuickCache<any>({
               url: `${tileApiEndpointToUse}/mosaic/register`, // @NOTE: This will fail anyways with "staging-raster.delta-backend.com" because its already deprecated...
@@ -298,8 +298,8 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
           } else {
               LOG && 
               /* eslint-disable-next-line no-console */
-              console.log('Titiler /register endpoint error', 'color: red;', e);
-              changeStatus({ status: S_FAILED, context: STATUS_KEY.Layer });
+              console.log('Titiler /register endpoint error', 'color: red;', error);
+              throw error;
           }
         }
 
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -297,8 +297,9 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
             setMosaicUrl(mosaicUrl);
           } else {
               LOG && 
-                /* eslint-disable-next-line no-console */
-                console.log('Titiler /register endpoint error', 'color: red;', e);
+              /* eslint-disable-next-line no-console */
+              console.log('Titiler /register endpoint error', 'color: red;', e);
+              changeStatus({ status: S_FAILED, context: STATUS_KEY.Layer });
           }
         }
 
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -274,7 +274,7 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
         /* eslint-enable no-console */
         
         let responseData;
-
+        let fallbackToOld = false;
         try {
           responseData = await requestQuickCache<any>({
             url: `${tileApiEndpointToUse}/searches/register`,
@@ -290,10 +290,17 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
               payload,
               controller
             });
+            fallbackToOld = true;
           }
         }
-
-        setMosaicUrl(responseData.links[1].href);
+        let mosaicUrl = responseData.links[1].href;
+        
+        if (fallbackToOld) { // @NOTE:  conditional logic TO BE REMOVED once new BE endpoints have moved to prod...
+          setMosaicUrl(mosaicUrl);
+        } else {
+          mosaicUrl = mosaicUrl.replace('/{tileMatrixSetId}', '');
+          setMosaicUrl(mosaicUrl.replace('/{tileMatrixSetId}', ''));
+        }
 
         /* eslint-disable no-console */
         LOG &&
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -298,7 +298,7 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
           } else {
               LOG && 
               /* eslint-disable-next-line no-console */
-              console.log('Titiler /register endpoint error', 'color: red;', error);
+              console.log('Titiler /register %cEndpoint error', 'color: red;', error);
               throw error;
           }
         }
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -272,12 +272,26 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
         LOG && console.log('Payload', payload);
         LOG && console.groupEnd();
         /* eslint-enable no-console */
+        
+        let responseData;
 
-        const responseData = await requestQuickCache<any>({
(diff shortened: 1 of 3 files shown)


app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -274,7 +274,7 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
         /* eslint-enable no-console */
 
         const responseData = await requestQuickCache<any>({
-          url: `${tileApiEndpointToUse}/mosaic/register`,
+          url: `${tileApiEndpointToUse}/searches/register`,
```
hanbyul-here, line 277: 👋 It will be awesome if we can make this code work regardless of the backend api so we don't have to worry about when the backend code will be shipped. 

For example, do you think it would be possible to check one of the new end point that is included in the new titiler update and let the code to fallback to the old behavior when the new api endpoint doesn't exist? (This is just an idea for an example, it woudl be even better if there is a simpler way to do so!)

sandrahoang686: we could do that but I think we can just feature flag then? Once BE ships to prod, we can turn on flag, if all is good, go in with a follow up to remove flag.

sandrahoang686: But it might be useless to have a fallback because the old endpoints wont work anyways once these changes move to prod on the BE. I think it makes more sense to just have careful coordination with BE? Unless i'm misunderstanding you..

hanbyul-here: I can't think of how to make a feature flag work without stressing over the timing of the backend deployment. Will the flag be set as an environment variable? If so, should the change in this environment variable still be coordinated with the backend deployment timing? I might be missing something, so if you have a concrete idea, I'd love to see it! 

So ideal sequence will be 

- ship dashboard code that works for both old and new endpoints
- release the dashboard with the code from the previous step
- ship backend
- remove the code for old endpoint

sandrahoang686: yep, FF would be instead of checking for a condition, we manually pull a trigger without touching code at point in time but used to doing that with LaunchDarkly which we haven't been doing. Will add this condition so that this change can merge without waiting. We will need the instances to have this change before BE moves to prod 👍🏼 thanks Hanbyul!

anayeaye:
The other breaking change in the titiler-pgstac v1 upgrade is related to  **`{tileMatrixSetId}`** for all **`tiles`** requests. The issue is actually in using the template urls from the searches/register response. I hadn't thought of this until now:

**Migration docs (background)**
https://stac-utils.github.io/titiler-pgstac/latest/endpoints/searches_endpoints/#tiles
`:endpoint:/searches/{search_id}/tiles[/{TileMatrixSetId}]/{z}/{x}/{y}[@{scale}x][.{format}]`

**TileMatrixSetId:** TileMatrixSet name, default is WebMercatorQuad. OPTIONAL

**Staging examples (copy paste these long URLs to a browser)**
TileMatrixSetId in path works
`https://staging.openveda.cloud/api/raster/searches/6b9ce6a9b0323759c79045ac074f54b9/WebMercadorQuad/tilejson.json?assets=cog_default&resampling_method=bilinear&bidx=1&color_formula=gamma r 1.05&colormap_name=coolwarm&rescale=0,15000000000000000`

and TileMatrixSetId NOT in path params works
`https://staging.openveda.cloud/api/raster/searches/6b9ce6a9b0323759c79045ac074f54b9/tilejson.json?assets=cog_default&resampling_method=bilinear&bidx=1&color_formula=gamma r 1.05&colormap_name=coolwarm&rescale=0,15000000000000000`

BUT
The register mosaic response template includes TileMatrixSetId
```
curl -X 'POST' \
  'https://staging.openveda.cloud/api/raster/searches/register' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{"filter-lang":"cql2-json","filter":{"op":"and","args":[{"op":">=","args":[{"property":"datetime"},"2023-12-01T00:00:00.000Z"]},{"op":"<=","args":[{"property":"datetime"},"2023-12-01T23:59:59.999Z"]},{"op":"eq","args":[{"property":"collection"},"no2-monthly"]}]},"limit":500,"fields":{"include":["bbox"],"exclude":["collection","links"]}}' | jq

{
  "id": "110aafd44491eca143e849840129273c",
  "links": [
    {
      "href": "https://staging.openveda.cloud/api/raster/searches/110aafd44491eca143e849840129273c/info",
      "rel": "metadata",
      "title": "Mosaic metadata"
    },
    {
      "href": "https://staging.openveda.cloud/api/raster/searches/110aafd44491eca143e849840129273c/{tileMatrixSetId}/tilejson.json",
      "rel": "tilejson",
      "templated": true,
      "title": "Link for TileJSON (Template URL)"
    },
    {
      "href": "https://staging.openveda.cloud/api/raster/searches/110aafd44491eca143e849840129273c/{tileMatrixSetId}/WMTSCapabilities.xml",
      "rel": "wmts",
      "templated": true,
      "title": "Link for WMTS (Template URL)"
    }
  ]
}
```

So without a TileMatrixSetId defined we end up with bad urls because the path parameter placeholder is in the template:
`https://staging.openveda.cloud/api/raster/searches/6b9ce6a9b0323759c79045ac074f54b9/%7BtileMatrixSetId%7D/tilejson.json?assets=cog_default&resampling_method=bilinear&bidx=1&color_formula=gamma r 1.05&colormap_name=coolwarm&rescale=0,15000000000000000`

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
sandrahoang686: @hanbyul-here I made this update to fall back back on old urls, can you check to see if that is what you meant? If you know a better way to do this without having to make the request first, please suggest. Also a note.. old and new endpoints will fail with `https://staging-raster.delta-backend.com` anyways because its deprecated

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -272,14 +272,35 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
...
+            });
+            fallbackToOld = true;
+          }
+        }
+        let mosaicUrl = responseData.links[1].href;
+        
+        if (fallbackToOld) { // @NOTE:  conditional logic TO BE REMOVED once new BE endpoints have moved to prod...
+          setMosaicUrl(mosaicUrl);
+        } else {
+          mosaicUrl = mosaicUrl.replace('/{tileMatrixSetId}', '');
```
hanbyul-here, line 301: Should we just pass `WebMercatorQuad` as TileMatrixSetId to be more explicit?

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -272,14 +272,35 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
...
+              url: `${tileApiEndpointToUse}/mosaic/register`, // @NOTE: This will fail anyways with "staging-raster.delta-backend.com" because its already deprecated...
+              payload,
+              controller
+            });
+            fallbackToOld = true;
+          }
+        }
+        let mosaicUrl = responseData.links[1].href;
+        
+        if (fallbackToOld) { // @NOTE:  conditional logic TO BE REMOVED once new BE endpoints have moved to prod...
```
hanbyul-here, line 298: I think we can move all the logics for the new endpoint in try block from line 278 instead of depending on the flag! How do you think?

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -272,14 +272,35 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
...
+          if (e.request) {
+            // The request was made but no response was received
+            responseData = await requestQuickCache<any>({
+              url: `${tileApiEndpointToUse}/mosaic/register`, // @NOTE: This will fail anyways with "staging-raster.delta-backend.com" because its already deprecated...
+              payload,
+              controller
+            });
+            fallbackToOld = true;
+          }
+        }
```
hanbyul-here, line 295: Should we have an else block to throw an error in case the error is not from the new/old API endpoint problem?

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
sandrahoang686: oh true

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
sandrahoang686: I added an else block with a `LOG` following the patterns in this file, do you think this is sufficient enough?

dzole0311 (changes requested):
Looks good, added only one comment.

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -273,13 +273,34 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
...
+              payload,
+              controller
+            });
+            
+            const mosaicUrl = responseData.links[1].href;
+            setMosaicUrl(mosaicUrl);
+          } else {
+              LOG && 
+              /* eslint-disable-next-line no-console */
+              console.log('Titiler /register endpoint error', 'color: red;', e);
```
dzole0311, line 301: Should we change the status here aside from the logging? Otherwise it might not propagate the error to the UI in case something goes wrong.

`changeStatus({ status: S_FAILED, context: STATUS_KEY.Layer })`

app/scripts/components/common/mapbox/layers/raster-timeseries.tsx
```diff
@@ -273,13 +273,34 @@ export function MapLayerRasterTimeseries(props: MapLayerRasterTimeseriesProps) {
...
+        let responseData;
+
+        try {
+          responseData = await requestQuickCache({
+            url: `${tileApiEndpointToUse}/searches/register`,
+            payload,
+            controller
+          });
+          const mosaicUrl = responseData.links[1].href;
+          setMosaicUrl(mosaicUrl.replace('/{tileMatrixSetId}', '/WebMercatorQuad'));
```
anayeaye, line 285: probably nice to have tileMatrixSetId as a variable but this change works with the upgraded titiler-pgstac version ✅

anayeaye (approved):

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -272,14 +272,36 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
         LOG && console.log('Payload', payload);
         LOG && console.groupEnd();
         /* eslint-enable no-console */
+        
+        let responseData;
```
hanbyul-here, line 276: since the data is not needed outside of the scope, I think we can scope this into each block, (then removing the old code bits will be even easier I think.)

sandrahoang686: responseData is still being referenced here though https://github.com/NASA-IMPACT/veda-ui/blob/80687e78127a3f57a998cd98646d4a6b591ad655/app/scripts/components/common/map/style-generators/raster-timeseries.tsx#L314-L315

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -272,14 +272,36 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
...
+            payload,
+            controller
+          });
+          const mosaicUrl = responseData.links[1].href;
+          setMosaicUrl(mosaicUrl.replace('/{tileMatrixSetId}', '/WebMercatorQuad'));
+        } catch (e) {
+          // @NOTE: conditional logic TO BE REMOVED once new BE endpoints have moved to prod... Fallback on old request url if new endpoints error with nonexistance... 
+          if (e.request) {
+            // The request was made but no response was received
+            responseData = await requestQuickCache<any>({
```
hanbyul-here, line 290: Ah, I realized that we might need a nested try- because we cannot catch the error from this request. But then I realized that this try is already nested 😅(https://github.com/NASA-IMPACT/veda-ui/blob/80687e78127a3f57a998cd98646d4a6b591ad655/app/scripts/components/common/map/style-generators/raster-timeseries.tsx#L153-L210)  maybe we can move this to the first-level try-catch? https://github.com/NASA-IMPACT/veda-ui/blob/80687e78127a3f57a998cd98646d4a6b591ad655/app/scripts/components/common/map/style-generators/raster-timeseries.tsx#L195

sandrahoang686: hah also just saw this.. i ended up throwing the error in nested try/catch so error can bubble up to the first try/catch.  I think this is fine, i'm wrapping the nested try catch directly around the request and catching and checking for a specific error right away => that errors either leads to old request or throws error back up. It will be easy to identify and clean up later too.

hanbyul-here: oh yeah, I like this approach (https://github.com/NASA-IMPACT/veda-ui/pull/1075/files#diff-fc839f8286ca1783e679e0fcddd8a3da452570ebbe9f433e16fda7003b91687eR302) better.

As far as I can see, I don't think the error for this specific request will be caught because this request is already in 'catch' block unless we are wrapping it with try-catch again?

hanbyul-here: Oh wow, I mistook how this would work; I guess the parental level of try-catch catches the error in the nested catch? 🤔 Thanks for giving it a try!

hanbyul-here (approved):
Thanks for working on this 🙇

PR #1086, opened 2024-08-05
[E & A] Date picker aligns with temporal density + simplify timeline header

dzole0311:

### Description of Changes

### Validation / Testing

### DIFF
--- a/app/scripts/components/exploration/components/timeline/date-axis.tsx
+++ b/app/scripts/components/exploration/components/timeline/date-axis.tsx
@@ -154,7 +154,7 @@ export function DateAxis(props: DateAxisProps) {
             key={`minor-${d.getTime()}`}
             x1={xPos}
             x2={xPos}
-            y1={24}
+            y1={26}
             y2={32}
           />
         );
--- a/app/scripts/components/exploration/components/timeline/date-axis.tsx
+++ b/app/scripts/components/exploration/components/timeline/date-axis.tsx
@@ -1,11 +1,8 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import styled from 'styled-components';
 import { ScaleTime } from 'd3';
 import {
-  format,
-  isSameMonth,
-  isSameYear,
-  startOfMonth,
+  format, startOfMonth,
   startOfYear
 } from 'date-fns';
 import { themeVal } from '@devseed-ui/theme-provider';
@@ -18,15 +15,11 @@ const GridLine = styled.line`
 `;
 
 const DateAxisSVG = styled.svg`
-  border-top: 1.5px solid #dde0e3;
+  border-top: 1px solid ${themeVal('color.base-200')};
 
   text {
     font-size: 0.75rem;
     fill: ${themeVal('color.base')};
-
-    &.parent {
-      font-size: 0.625rem;
-    }
   }
 `;
 
@@ -47,14 +40,14 @@ function getTimeDensity(domain) {
   }
 }
 
-function timeDensityFormat(date: Date, timeDensity: TimeDensity) {
+function dateAxisLabelFormat(date: Date, timeDensity: TimeDensity) {
   switch (timeDensity) {
     case TimeDensity.YEAR:
       return format(date, 'yyyy');
     case TimeDensity.MONTH:
-      return format(date, 'MMM dd');
+      return format(date, 'MMM');
     case TimeDensity.DAY:
-      return format(date, 'iii dd');
+      return format(date, 'dd');
   }
 }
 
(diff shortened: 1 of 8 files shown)


faustoperez (approved):
The date selection is much clearer now, great job @dzole0311

faustoperez:
I found that the horizontal scrolling of the timeline isn't working in this deploy preview. I approved the changes as I wasn't I wasn’t sure the bug was related to this particular fix.

image

dzole0311:
Thanks @faustoperez!  Based on the back-end response, the temporal extent for those datasets is `[["2003-01-01T00:00:00+00:00", "2017-12-31T00:00:00+00:00"]]` which is where the timeline currently stops scrolling.



So I'm a bit confused though, on where you saw January 2000 to December 2021 as a temporal extent?

Also, I realized we should set the min and max dates of the date picker according to the min/max dates of the selected timeline datasets. I'll go ahead and fix that!

aboydnw (approved):
Looks good to me, I wasn't able to find any issues. I wasn't able to test an annual dataset, because I didn't see one available, but I assume it will function similar to the monthly dataset.

I also think the fact that a user can select a date outside the temporal extent is more of a minimal issue now because of the updates to the playhead and timeline, the user can more clearly see that they picked a date outside the temporal extent. So, if it's more than a trivial change, maybe we leave it off for now unless you all disagree.

dzole0311:
Thanks for testing @aboydnw! The date picker should now also be limited to the selected datasets min & max domain.

hanbyul-here:
👋 The date selection does look much better!!Thanks for working on it. While clicking around, I found some unexpected stles. 
This color scheme doesn't look right, and I can only find this color combination on Fire datasets? (I am not even sure what yellow is supposed to mean)  Also, shouldn't 'selected date' have the primary color? @faustoperez ?

Screenshot 2024-08-06 at 9 14 07 AM

dzole0311:
> This color scheme doesn't look right, and I can only find this color combination on Fire datasets? (I am not even sure what yellow is supposed to mean)

Thanks for testing @hanbyul-here ! The yellow comes from react-calendar and shows the 'now' month, not the selected (in your case Aug is 'now', July is the 'blue' selected). The styles I added for overriding the 'yellow' seemed to have been wiped by purgeCSS (locally it works fine).

Also, the EIS Fire dataset has a small temporal extent (
`["2024-07-16T12:00:00+00:00", "2024-08-05T12:00:00+00:00"]`), thus only two months are available for selecting. But I should have set the default view of the calendar to a month view instead of a year view, so you could select days there -> fixed!

app/scripts/components/exploration/components/timeline/date-axis.tsx
```diff
@@ -75,6 +68,71 @@ function getTicks(scale: ScaleTime<number, number>) {
...
+  const minorTickInterval = (majorTicks[1].getTime() - majorTicks[0].getTime()) / segments;
+
+  // Initialize minorTicks as an array of Date objects
+  let minorTicks: Date[] = [];
+
+  // Add minor ticks before the first major tick (left hand side of
+  // the timeline) until reaching the view start
+  let tickTime = majorTicks[0].getTime() - minorTickInterval;
+
+  while (tickTime > viewStart.getTime()) {
```
hanbyul-here, line 108: aha this should be done in three parts like this, such a detailed touch!

hanbyul-here (changes requested):
Most of the comments are minor, but we probably should not put the new utility in the deprecated file. Thanks for the great work!

app/scripts/components/common/map/controls/aoi/preset-selector.tsx
```diff
@@ -117,7 +117,7 @@ const OptionValueDisplay = styled.div`
 const SelectorSubAction = css`
   position: absolute;
   top: 0;
-  right: 10px;
+  right: ${glsp(1.3)};
```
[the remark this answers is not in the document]
hanbyul-here: + (Maybe 1.25 instead of 1.3 so we can get 20px instead of 20.8px? 😬 )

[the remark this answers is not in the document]
dzole0311: This came out of a chat with @faustoperez, and was a low hanging 🥝 to fix, so we sneaked it in here 😃

app/scripts/components/exploration/components/timeline/date-axis.tsx
dzole0311: yeah, I couldn't figure out any other way on how to show the minor vertical ticks on the edges (after the last major ticks), without them completely disappearing until another major tick pops up when a user swipes left / right.

app/scripts/components/exploration/data-utils.ts
```diff
@@ -227,3 +228,34 @@ export class ExtendedError extends Error {
...
+    const currentDensity = obj.data.timeDensity;
+    if (
+      TIME_DENSITY_ORDER[currentDensity] <
+      TIME_DENSITY_ORDER[lowestCommonDensity]
+    ) {
+      lowestCommonDensity = currentDensity;
+    }
+  }
+
+  return lowestCommonDensity;
```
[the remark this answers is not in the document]
dzole0311: Thanks for the suggestion! `Math.min` probably won't work here as `e.data.timeDensity` is not number, but I simplified it with a reduce function (lmk wdyt!)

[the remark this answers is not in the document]
hanbyul-here: oh, right! Thanks for taking another look :)

hanbyul-here:
Interseting.. the 'now' color doesn't work only on a super wide screen! (~ > 2500px) When I make my screen smaller, the color goes away.
Screenshot 2024-08-08 at 9 03 32 AM

app/scripts/components/exploration/components/timeline/date-axis.tsx
```diff
@@ -75,6 +68,71 @@ function getTicks(scale: ScaleTime<number, number>) {
...
+  // Number of segments to divide the interval between major ticks
+  // Set to 10 to for a balance between the level of detail, but also
+  // it offers a better alignment of the timeline playhead over the
+  // minor ticks.
+  const segments = 10;
+
+  // Ensure there are at least two major ticks to define an interval
+  if (majorTicks.length < 2) {
+    return [];
+  }
```
sandrahoang686, line 96: ⛏️ : can combine with first condition and return right away if so

app/scripts/components/common/datepicker/index.scss
```diff
@@ -38,6 +38,16 @@
         background-color: utils.color('gray-10') !important;
       }
     }
+
+    &--now:disabled {
```
hanbyul-here, line 42: I can't see the button getting disabled. (Therefore, this style isn't applied.) Am I missing anytyhing? 
 
Screenshot 2024-08-09 at 11 18 06 AM

app/scripts/components/exploration/components/timeline/timeline-controls.tsx
```diff
@@ -228,8 +236,27 @@ export const TimelineHeadIndicators = memo(({ outOfViewHeads }: { outOfViewHeads
...
+/**
+ * Determines the appropriate calendar view based on the time density.
+ *
+ * The TimeDensity enumeration is mapped to the corresponding calendar view:
+ * - MONTH: Displays the calendar in 'year' view, showing all months in a year.
+ * - YEAR: Displays the calendar in 'decade' view, showing multiple years in a decade.
+ * - Default: Displays the calendar in 'month' view, showing all days of the current month.
+ */
+const getCalendarView = (timeDensity: TimeDensity): View | undefined => {
+  switch (timeDensity) {
```
sandrahoang686, line 248: ⛏️ : can get rid of `undefined` return type? Because it looks like it would default to `month` string

sandrahoang686 (commented):
Screenshot 2024-08-09 at 11 18 34 AM
Small nit pick comments and then had a question about whether this ^ was okay? I would expect day to show up but if not, then cool, wanted to bring up

app/scripts/components/common/datepicker/index.scss
dzole0311: I might be misunderstanding your question, so let's catch up over a call if I do. But I think this is an edge case, which is difficult to reproduce now when we added a limit to the calendar min and max dates.

Here's a simple scenario: you have a dataset with temporal extent from Aug 5, 2024- Aug 10, 2024. If you open the calendar picker, you would see the `now` date (e.g Aug 12 which is today) and it will be disabled due to the temporal extent of the dataset. It will also be colored yellow due to `react-calendar` and our style aims to override that (to align with the USWDS behavior).

dzole0311:
> I would expect day to show up

@sandrahoang686 thanks for catching this, it was not as expected

app/scripts/components/common/datepicker/index.scss
hanbyul-here: Ah I see ! I somehow thought we want to disable 'now' (marked with a yellow background) totally. Thanks for walking through the case 🙇

hanbyul-here (approved):
Thanks for the awesome work 🙇 

The name and the comment of the hook `useTimelineDatasetsDomain` does not seem to reflect what it does now, so it might be worth noting that the domain set changes based on timeline width in the comment? But I will leave it up to you

Issue #1088, opened 2024-08-05
Hiding certain data stories from catalog


aboydnw:
### Context

For GHGC R3, they will be using the data stories construct to create custom pages for and Urban landing page (and others)

This will help us meet the needs for R3, but we do not want those pages to appear in the stories catalog. So, we need to make sure some stories are hidden. This can be manually defined.

Sounds like this is already possible, just needs some custom work. We should work with the UAH team to get this done.

### Acceptance Criteria

- [ ] Stories manually identified by the GHG team do not appear on the stories catalog view, but can still be viewed by navigating directly to them (or via top nav, other means)

slesaad:
@hanbyul-here @sandrahoang686 can you add details here for what custom work needs to be done? I wasn't aware it was already possible 😄

dzole0311:
@slesaad I opened up a draft to check if this is what we want: https://github.com/NASA-IMPACT/veda-ui/pull/1090

Basically, I introduced a `isHidden` prop which could be optionally added to the story in the .mdx. This story will then not be shown in the stories catalog, but it will still be known to the veda-ui and could be navigated to via routing. 

Example with "This is the life of water":

Stories catalog (hidden): https://deploy-preview-1090--veda-ui.netlify.app/stories
Story page: https://deploy-preview-1090--veda-ui.netlify.app/stories/life-of-water

However the story will still appear as a featured story unless `isFeatured` is also set to `false` in the .mdx, so it's pretty configurable.

@slesaad is this the behavior that we want for GHG?

slesaad:
@dzole0311 that's exactly what we wanted! great that you also added it to datasets, I'm sure that's gonna be really useful too. thanks for doing this so quickly!

slesaad:
@dzole0311 just noticed that the count in the stories page still counts the hidden stories - example here https://deploy-preview-451--ghg-demo.netlify.app/stories. That might be confusing to the users.

aboydnw:
I don't think this was ever done, and probably isn't valuable for GHGC anymore. But, I wonder if it could be valuable for other instances like Disasters and earth.gov? Imagining a theme page with a list of stories pages under it. Right now in earth.gov we have to manually manage the story cards on a themes page. If that list grows at all, it might be easier to just place a catalog that shows the linked stories. Similar for disasters, if they want to show any story related to an event, or hazard type.

I'll move to #1920 then but feel free to close if you'd rather have a clean backlog and wait for a specific stakeholder request for this.

PR #1090, opened 2024-08-06
Add isHidden prop for optional hiding of stories

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1088

### Description of Changes
This ticket introduces a `isHidden` prop for both stories and datasets, so that they could be hidden from the appropriate catalogs. If a story has this new prop, then it will not be shown in the stories catalog, but it will still be known to the veda-ui and could be navigated to via routing.

Example with "This is the life of water":

Stories catalog (hidden): https://deploy-preview-1090--veda-ui.netlify.app/stories
Story page: https://deploy-preview-1090--veda-ui.netlify.app/stories/life-of-water

Note: the story will still appear as a featured story unless `isFeatured` is also set to false in the .mdx, so it's pretty configurable.

### Validation / Testing

### DIFF
--- a/app/scripts/components/common/catalog/prepare-datasets.ts
+++ b/app/scripts/components/common/catalog/prepare-datasets.ts
@@ -15,20 +15,29 @@ interface FilterOptionsType {
   filterLayers?: boolean | null;
 }
 
-export function prepareDatasets(data: DatasetData[], options: FilterOptionsType): DatasetData[];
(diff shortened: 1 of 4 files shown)


hanbyul-here (approved):
🙇 Thanks for working on this (and fixing the lints :] )!

Issue #1085, opened 2024-08-05
[VEDA2 Refactor] Data Catalog Component


aboydnw:
### Context

This ticket seeks to break out the Data Stories component into a view that renders the type of look the current Data Catalog page 

Notes:
- Mark new files for VEDA2 refactor with @VEDA2-REFACTOR-WORK

### Acceptance Criteria

- [ ]  Create a view component that can be used and imported directly to render the view of the current data stories hub
- [ ]  Reusability => modularize stateless/presentational components to be as reuseable as possible within the core feature view containers
- [ ]  Feature Component should be entirely data agnostic => Separate out the data logic from this feature component into the data layer. Note: Data will now have to be passed in as props so we will need to update the page views to supply this
- [ ]  Feature Components should have stricter typed interfaces
- [ ] This component should still support current instances and be backwards compatible

Related to: https://github.com/NASA-IMPACT/veda-ui/issues/801

Resources:

ADR: https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/adr/002-application-architecture-for-configurability.md#decision-outcome
[Miro Board outline Core Features](https://miro.com/app/board/uXjVNlIZVJI=/)

PR #1096, opened 2024-08-07
Expose Data Catalog and update child components to pass in routing/nav

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1085

### Description of Changes
* Exposes Data Catalog component
* Updates child components to pass in routing/nav and link component

* Manually test that Data Catalog filters are working as is in current architecture

### Validation / Testing
- [ ] Make sure Data Catalog works as expected including filtering and all cards
- [ ] Make sure Data Layer Selector Modal cards work as expected
- [ ] Make sure stories hub cards works as expected

### DIFF
--- a/app/scripts/components/common/card/index.tsx
+++ b/app/scripts/components/common/card/index.tsx
@@ -262,14 +262,14 @@ function CardComponent(props: CardComponentProps) {
     linkProperties
   } = props;
 
-  const isExternalLink = /^https?:\/\//.test(linkProperties?.linkTo);
+  const isExternalLink = /^https?:\/\//.test(linkProperties.linkTo);
 
   return (
     <ElementInteractive
       linkProps={{
-        as: linkProperties?.LinkElement,
-        [linkProperties?.pathAttributeKeyName]: linkProperties?.linkTo,
-        onLinkClick: linkProperties?.onLinkClick,
+        as: linkProperties.LinkElement,
+        [linkProperties.pathAttributeKeyName]: linkProperties.linkTo,
+        onLinkClick: linkProperties.onLinkClick,
       }}
       as={CardItem}
       cardType={cardType}
--- a/app/scripts/components/common/card/index.tsx
+++ b/app/scripts/components/common/card/index.tsx
@@ -1,4 +1,5 @@
 import React, { MouseEventHandler } from 'react';
+import { Location } from 'react-router';
 import styled, { css } from 'styled-components';
 import { Link } from 'react-router-dom';
 import { format } from 'date-fns';
@@ -239,7 +240,8 @@ export interface CardComponentProps {
   footerContent?: JSX.Element;
   onCardClickCapture?: MouseEventHandler;
   onLinkClick?: MouseEventHandler;
-  CardInteractiveElement?: any;
+  OverrideLinkElement?: JSX.Element;
+  location?: Location | string;
 }
 
 function CardComponent(props: CardComponentProps) {
@@ -259,17 +261,19 @@ function CardComponent(props: CardComponentProps) {
     footerContent,
     onCardClickCapture,
     onLinkClick,
-    CardInteractiveElement,
+    OverrideLinkElement,
+    location,
   } = props;
 
   const isExternalLink = /^https?:\/\//.test(linkTo);
-
+  
   return (
     <ElementInteractive
       linkProps={{
         as: SmartLink,
         to: linkTo,
-        onLinkClick
+        onLinkClick,
+        ...(OverrideLinkElement && {OverrideTag: OverrideLinkElement})
       }}
       as={CardItem}
       cardType={cardType}
@@ -278,7 +282,8 @@ function CardComponent(props: CardComponentProps) {
       linkTo={linkTo}
       onLinkClick={onLinkClick}
       onClickCapture={onCardClickCapture}
-      CardInteractiveElement={CardInteractiveElement}
+      OverrideLinkElement={OverrideLinkElement}
+      location={location}
     >
       {
         cardType !== 'horizontal-info' && (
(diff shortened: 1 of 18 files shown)


app/scripts/components/data-catalog/container.tsx
```diff
@@ -15,15 +18,32 @@ import { veda_faux_module_datasets } from '$data-layer/datasets';
  * veda2 instances can just use the direct component, 'DataCatalog', and manage data directly in their page views
  */
 
+const InteractiveLink = styled.a`
```
hanbyul-here, line 21: I see no `CardInteractiveElement` is passed from Next instance from [this pr from Next instance](https://github.com/developmentseed/next-veda-ui/pull/3/files#diff-6578b281e50705562cdf8ba73c73ae2bb1df366e18e91e9595106b09cfbf3f89). Based on [this line](https://github.com/NASA-IMPACT/veda-ui/pull/1096/files#diff-c618b2febe87cadd3d3931c1498d8b40c67f5793d0235e3777dd79a202183f34R78), it will be subbed as `a` tag. But `InteractiveLink` here is more than an `a` tag. It is a styled component with a style to expand its target area. I can't seem to click on the card on this page : https://deploy-preview-3--veda-ui-next-test.netlify.app/datasets .

Also, I think the interactive link could be rendered with either `a` tag or `Link` because styled component accepts `as` as a polymorphic prop to decide which component to render (https://styled-components.com/docs/api#as-polymorphic-prop)  - if you look at the rendered html from next instance, you can see that `a` tag as a pure html doesn't know what to do with `as`.
Screenshot 2024-08-28 at 3 31 03 PM


This is related to why I opened #1121 . We can leave the styled component as it is but pass the component only in charge of routing as a prop. 

What do you think?

app/scripts/components/common/catalog/catalog-content.tsx
```diff
@@ -34,6 +35,9 @@ export interface CatalogContentProps {
   search: string;
   taxonomies: Record<string, string[]>;
   onAction: (action: FilterActions, value?: any) => void;
+  linkProperties: LinkProperties;
+  location?: Location | string;
+  rootPath?: string;
```
sandrahoang686, line 40: Asking for both `location` and `rootPath` separately because location is used filterCards component and is expected from nextJs instance and container. While `rootPath` is used to decide the navigation path when clicking on a card. Container doesn't need to pass this in for now because it is defaulted but nextJs instance does

app/scripts/components/data-catalog/container.tsx
sandrahoang686: addressed

app/scripts/components/common/catalog/controls/hooks/use-filters-with-query.ts
```diff
@@ -32,11 +32,23 @@ export function useFiltersWithURLAtom(): UseFiltersWithQueryResult {
...
+    push=false
+  }: {
+    navigate: any,
+    push?: boolean,
+  }): UseFiltersWithQueryResult {
+    
+    let navCommit = navigate;
+
+    if (push) {
+      navCommit = ({ search }) => navigate.push(`?${search}`);
```
hanbyul-here, line 46: I think there is a potential to make it even more flexible if we just let the instance to pass this whole thing as navigate props. ex. Replace this line: https://github.com/developmentseed/next-veda-ui/pull/3/files#diff-6578b281e50705562cdf8ba73c73ae2bb1df366e18e91e9595106b09cfbf3f89R14 
```
 const controlVars = useFiltersWithQS({navigate: ({ search }) => router.push(`?${search}`)});
```

then let's say if the instance wants to replace the url instead of push, it can handle it on its level. (and no additional push props is needed) What do you think?

sandrahoang686: I think that also makes sense but I think we should work on closing out the PR if this is not blocking we can address this in the routing ticket / make note of this

hanbyul-here (approved):
Thanks for working on it 🙇 It will be great to get one more testing (clicking around and making sure everything is ok) but here is my approval for a swift merge.

dzole0311 (changes requested):
I think I spotted a bug which I can't see in the rest of the deployments:

1. Go to Data Catalog
2. Click on "CASA-GFED3 Land Carbon Flux"
3. Click on "Explore Data"
4. I get an error page with `Cannot destructure property 'pathname' of 'e' as it is undefined`

The button should lead to the E&A page with the dataset filtered.

app/scripts/components/common/card/index.tsx
```diff
@@ -257,24 +259,22 @@ function CardComponent(props: CardComponentProps) {
...
-  const isExternalLink = /^https?:\/\//.test(linkTo);
+  const isExternalLink = /^https?:\/\//.test(linkProperties?.linkTo);
 
   return (
     <ElementInteractive
       linkProps={{
-        as: SmartLink,
-        to: linkTo,
-        onLinkClick
+        as: linkProperties?.LinkElement,
```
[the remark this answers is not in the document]
sandrahoang686: removed ✅

app/scripts/components/common/catalog/catalog-card.tsx
```diff
@@ -108,6 +111,9 @@ export const CatalogCard = (props: CatalogCardProps) => {
     }
   };
 
+  // Either type Location for return type of react-router-dom "useLocation" or string from nextJs "usePathname"
+  const linkTo = typeof location === 'string' ? getDatasetPath(dataset, location) : getDatasetPath(dataset);
```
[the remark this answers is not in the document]
sandrahoang686: addressed, we only needed pathname from location object. Changed `location` prop to `pathname` and always expect string ✅

sandrahoang686:
@dzole0311 thanks for catching, fixed ✅

aboydnw:
Looks good to me! I clicked around the data catalog/overview, stories, and E&A and didn't. notice anything off

Issue #1098, opened 2024-08-07
Implement Carousel


aboydnw:
### Context

For GHGC R3, We have landed on a carousel design that GHG stakeholders agreed with and we feel is relatively accessible. This is intended to be part of R3. Ultimately we need some version of this in production by the end of August, but the earlier we can have something up and running, the faster we can get feedback, configure content, etc. 

It does not need to be a core component right now, although we expect that we'll want to make it a core component in the future.

Design ticket: #1084 

### Acceptance Criteria
- [ ] To the best of our ability, the carousel matches the designs
- [ ] Carousel auto progresses through each panel at a pace of 5 seconds per panel
- [ ] There is a "loading" animation to depict the current panel and progress of the autoprogression
- [ ] There is an animation (TBD) when the panel autoprogresses
- [ ] (ideally) When configuring, we can add more than 1 call to action link at the bottom of the content
- [ ] (ideally) When configuring, we can have 2, 3, or 4 panels

slesaad:
Some new info on the carousel:

> Please go ahead with the new carousel design, we’re good with three features. These will link to:
> 
> 1. Urban landing page
> 2. GRA2PES story page
> 3. NYC story page
> 
> Just to reiterate my gratitude to the team for coming up with an elegant and functional solution here, we really appreciate your patience and ingenuity!
> 
> I will owe you supporting text for the carousel, among other pages, and that’s on my list to do next week. Will need to decide images as well, but if it’s not too much trouble to use some generic ones for now just to have something in staging:
> 
> 1. Busy city sidewalk (maybe one like this, I’d just like to have some sort of human element in one of the images: > https://www.istockphoto.com/photo/crowd-of-people-on-new-york-city-sidewalk-gm896586898-247564664)
> 2. GRA2PES screenshot
> 3. NYC skyline

PR #1104, opened 2024-08-13
Adding styles & Configurations for R3

hanbyul-here:
**Related Ticket:** #1098 
+ We are encouraging IMPACT team to use USWDS. This new style configuration that uses grid-layout will be useful for IMPACT team. (I also need this style for Carousel )

### Description of Changes
- Added grid-layout style from uswds design system
- Edited purgecss configuration to include `instance/overrides` folder as content
- Added CSS Variables for VEDA customization colors. 

I almost gave up getting purgecss to work, but @dzole0311 helped me figure out what was wrong. (The default extractor doesn't allow special characters like `:`, which we need to target specific screen size like `tablet: ` ex. https://designsystem.digital.gov/utilities/layout-grid/ Thank you so much 🙇 @dzole0311 !

@slesaad , you don't need to review this PR, but I think it would be beneficial to check how we are importing the styles and purging the styles through `postcss.config`.

### Validation / Testing
This PR doesn't bring any visual change to UI, but this GHG pr uses this branch: https://github.com/US-GHG-Center/veda-config-ghg/pull/482

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -7,7 +7,8 @@ const isVedaInstance = fs.existsSync(vedaPath);
 
 let basePath = CWD;
 let vedaUIPath = CWD;
-let contentPaths = [`${vedaUIPath}/app/**/*.{js,jsx,ts,tsx}`];
+let contentPaths = [`${vedaUIPath}/app/**/*.{js,jsx,ts,tsx}`
+];
 
 let uswdsPath = path.resolve(
   basePath,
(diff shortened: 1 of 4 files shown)


hanbyul-here:
purgecss configuration : https://github.com/FullHuman/purgecss/issues/86#issuecomment-391279202

dzole0311 (approved):
Thank you @hanbyul-here! Looking forward to any future feedback on USWDS usage from @slesaad as well as we continue to transition towards USWDS.

In addition to the docs mentioned in the description, we also have on our disposal the `<Grid />` React component from the react implementation of the USWDS library for layout purposes (depends on preference of the developer). Link to the relevant docs: https://trussworks.github.io/react-uswds/?path=/docs/components-grid--docs.

hanbyul-here:
Oh!! Thanks a lot for pointing the grid component from react-uswds @dzole0311 ! I didn't even think about it 😓 We probably need to list grid-related classes as a whitelist to use it, though, I guess?

postcss.config.js
```diff
@@ -16,32 +18,35 @@ let reactCalendarPath = path.resolve(
...
   );
   reactCalendarPath = path.resolve(
-    basePath,
+    vedaUIPath,
     'node_modules/react-calendar/dist/**/*.css'
   );
+
+  contentPaths = [
+    `${vedaUIPath}/app/**/*.{js,jsx,ts,tsx, html}`,
+    `${basePath}/overrides/**/*.{js,jsx,ts,tsx}`
```
slesaad, line 33: should this also include `custom-pages`? (sorry if I'm misunderstanding this :D )

hanbyul-here: Yes, I forgot that there are also custom pages; I will add them. Thanks for catching it!

Issue #1055, opened 2024-07-17
Create a raster paint layer component


abarciauskas-bgse:
## Context

To support https://github.com/NASA-IMPACT/veda-ui/issues/1054 where we need to integrate a titiler-cmr layer, we will first separate a raster paint layer that will be used by the raster layers (cmr-stac, zarr).

## Acceptance Criteria

- [ ] zarr and cmr-stac are using the same raster paint layer component

abarciauskas-bgse:
This has been completed

PR #1105, opened 2024-08-13
Create new raster paint layer module and factor out BaseTimeseriesPropos

abarciauskas-bgse:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1055
### Description of Changes
Simply put, this PR is a small refactoring to create a basic `RasterPaintLayer` available, which both ZarrTimeseries and CmrTimeseries use. 
* Both CmrTimeseires and ZarrTimeseries were using `ZarrPaintLayer`. It seems appropriate then to rename `ZarrPaintLayer` to `RasterPaintLayer` and put it in a separate file.
* Merged common attributes of `Raster`, `Zarr`, and `CmrTimeseriesPropos` into `BaseTimeseriesProps

1. Should I create a new issue to remove the Zarr and Cmr layers from the app/scripts/common/components/mapbox/layers?
2. I noticed that a request is made for tiles even when the selected date falls outside of the datasets temporal range. Is this the expected behavior? I raised an issue in titiler-xarray (https://github.com/developmentseed/titiler-xarray/issues/62) that the API should not respond with tiles in this case, but also wondering if a request should be made at all 🤔 
    * _This is probably something we need to fix in useZarr and useCmr, so the question is, should we do that in this PR or a new PR? I would propose a new PR to keep changes as minimal as possible._

### Validation / Testing
I loaded the layers in the deploy preview, used the show/hide and time slider to make sure all that functionality was working as expected. I came across the issue described in (2) under Notes & Questions About Changes

### DIFF
--- a/app/scripts/components/common/map/style-generators/cmr-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/cmr-timeseries.tsx
@@ -1,27 +1,9 @@
 import React from 'react';
-import { BaseGeneratorParams } from '../types';
-import { ZarrPaintLayer } from './zarr-timeseries';
+import { CMRTimeseriesProps } from '../types';
+import { RasterPaintLayer } from './raster-paint-layer';
 import { useCMR } from './hooks';
-import { ActionStatus } from '$utils/status';
 
-interface AssetUrlReplacement {
-  from: string;
-  to: string;
-}
-
-export interface CMRTimeseriesProps extends BaseGeneratorParams {
-  id: string;
-  stacCol: string;
-  date?: Date;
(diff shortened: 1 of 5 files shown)


hanbyul-here (approved):
LGTM 👍!  
I would like to know more about the issues mentioned in the PR, that the request being made for tiles even when the selected date falls outside of the dataset temporal range.  But if we are handling that issue separately anyway, that should not be a blocker.

abarciauskas-bgse:
@hanbyul-here I am going to merge this for now so as not to fall behind main but will be writing up an issue soon on the temporal extent bug.

Issue #885, opened 2024-03-13
Load colormaps from TiTiler to ensure tiles and legend agree


j08lue:
Currently, we have to configure both the TiTiler dynamic tiling parameters for a colormap and the stops for the legend component. In principle, the stops could be generated dynamically, ideally by TiTiler, since it receives the rendering specs and we need to make sure legend and tiles agree.

I added a comment to this existing TiTiler feature request:

- https://github.com/developmentseed/titiler/discussions/509

## Acceptance criteria

- [ ] Colormaps are loaded from TiTiler

j08lue:
TiTiler implemented a method to get the full color table (usually 255 stops), which should work fine for our purpose as well.

- https://github.com/developmentseed/titiler/pull/796

Btw, other projects I learned about have been loading colormaps from [D3](https://d3js.org/d3-scale-chromatic/sequential), where they were identical with TiTiler's.

j08lue:
The`/colormaps` endpoint (docs [here](https://developmentseed.org/titiler/endpoints/colormaps/)) does not seem to be added to the GHG Center and VEDA APIs yet.

- https://staging-raster.delta-backend.com/docs
- https://ghg.center/api/raster/docs

 Might need a ticket in `veda-backend`.

slesaad:
It can be a part of this - https://github.com/NASA-IMPACT/veda-backend/issues/364.

j08lue:
## Context and Problem Statement

A TiTiler UI wants to offer to the user a **curated** list of colormaps to choose from, with visual previews or the colormaps.


## Decision Drivers

1. Agreement between legend and rendered colors. Small differences could compromise the information people derive from the maps - possibly unnoticed. Larger differences break the user experience.
2. Ease of implementation and maintenance
3. Decent performance
[...]

PR #1107, opened 2024-08-16
994 exploration colormap

dzole0311:

### Description of Changes

### Validation / Testing


j08lue:
I had a first look - awesome to see this feature materialize!

🎉  switching between colormaps and reversing them works - the tiles on the map change and use the right colormaps

🐞 - the legend graphics is off - the tiles are shaded the way I would expect (knowing Viridis or Plasma), but the graphics in the drop-down do not match.

image

dzole0311:
> the legend graphics is off

Yeah, I think this is because I tried to rescale the colors based on the rescale values. I've removed this now since it makes the legend unclear.

j08lue:
Ah, I see. I think we will always want to make use of the full color range. We only limit the data range.

So there should not be a situation where the colormap would be rescaled / cropped, as far as I can see.

app/scripts/components/exploration/components/datasets/colormap-options.tsx
```diff
@@ -0,0 +1,227 @@
...
+  border-bottom: 1px solid ${themeVal('color.base-200')};
+`;
+
+const ToggleLabel = styled.label``;
+
+const ToggleInput = styled.input.attrs({ type: 'checkbox' })`
+  display: none;
+`;
+
+export const SEQUENTIAL_COLORMAPS = [
```
hanbyul-here, line 86: 🤔 Should we have a step  to validate these colormaps are supported color maps?( from https://openveda.cloud/api/raster/colorMaps)? In reality, the supported colormap will change so rarely that I am not sure if we need to add this one more step, but then we will be safe from whichever update from the backend.

j08lue: Ah, I was hoping that we were doing that - fetching the supported colormaps from TiTiler instead of hard-coding in the client?

- https://github.com/NASA-IMPACT/veda-ui/issues/885

hanbyul-here: Ah, I see; I missed that there is even an endpoint for getting the colors of each colormap using each colormap ID - ex. https://openveda.cloud/api/raster/colorMaps/magma. Then it might make sense to have only IDs for colormaps on the client side and get the values from the endpoint.

dzole0311: I wasn't aware of the endpoint as well, thanks @j08lue!

@hanbyul-here Since we use d3-color-schematic already which supports the colorMaps by ID and to avoid sending requests on each selection of a colorMap to fetch the values, I don't think we need to hardcode them at all in the UI. In general, we could fetch them from titiler, list them in the UI and use d3-color-schematic to interpolate the values. What do you think about this?

hanbyul-here: Oh, don't we want to curate the colormaps (not all the colormap options from the colormap endpoint)? I meant the IDs for these curated colormaps can be on the front end. (Sorry for not being clear.) 

I think it is always best to have one source of truth, so we don't have to think about how the colors from Titiler don't seem to match 100% with what the legend shows. (If that ever happens -  So we probably should at least check where D3 colormap values are from and how the values match up with the titiler values.) I also understand your concerns about not making too many requests. Since this is the type of data that doesn't change very often, I wonder if we can make a request while building and save it as a local file? 🤔 (This sounds like a way to make this ticket go out of the scope 😅 )

dzole0311: So titiler uses colorMaps from rio-tiler. And rio-tiler uses colorMaps (at least the ones that are currently on the client-side) derived from Matplotlib: https://cogeotiff.github.io/rio-tiler/colormap/#colormaps

d3-scale-chromatic also [derives the ones we use from Matplotlib](https://www.npmjs.com/package/d3-scale-chromatic/v/1.5.0#interpolateMagma), which is why I think we will add unnecessary overhead if we send requests to fetch hexes per colorMap ID or save them in local files.

> Oh, don't we want to curate the colormaps (not all the colormap options from the colormap endpoint

Yes I think this was the idea, to curate a selection of most widely used colors instead of showing all the supported ones. Users could still pass in a valid colorMap name which is not in the list at the moment (e.g. PuRd) and it should show up fine as a default value.

j08lue: I see the pros and cons of all options. I would really like a solution that makes maximum use of the nice TiTiler built-ins. But the decision is up to you! 

I lined up the options here, for overview (using an [ADR template](https://github.com/adr/madr/blob/develop/template/adr-template.md?plain=1)):

- https://github.com/NASA-IMPACT/veda-ui/issues/885#issuecomment-2296121564

Please feel free to review and let me know if I should open a proper ADR PR.

PR #1112, opened 2024-08-19
Ditch PurgeCSS, configure uswds to use only tablet 

hanbyul-here:
### Description of Changes
- This PR ditches PurgeCSS - because purgecss cannot be configured in scale with react-uswds now. 
- This PR also configures USWDS design system to use only tablet breakpoints, so we don't end up with multiple screen helper classes


postcss.config.js
```diff
@@ -1,54 +1,4 @@
...
-  content: contentPaths,
-  defaultExtractor: (content) => content.match(/[A-z0-9-:\/@]+/g) || [],
-  safelist: {
-    deep: [/usa-banner$/],
-    greedy: [/^usa-banner/]
-  }
-});
-
-// Do not purge for local development.
-if (process.env.NODE_ENV !== 'development') plugins = [...plugins, purge];
```
dzole0311, line 51: 🫡

dzole0311 (approved):
Great idea about the breakpoints!

Issue #1113, opened 2024-08-19
[E&A] Annual datasets show as monthly in the date picker and timeline


aboydnw:
### Context

During testing on GHG Center, we can see that when an annual dataset is selected that the date picker correctly narrows down selection to the year level, but the timeline view and datepicker still display the month.

![Image](https://github.com/user-attachments/assets/f5a7e554-f4c2-4777-b39f-6fb943d9794e)

### Acceptance Criteria

- [ ] Only the year appears in the datepicker and timeilne view when only viewing annual datasets

PR #1114, opened 2024-08-20
Fix the label formatting for annual datasets

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1113

### Description of Changes
The change fixes the label formatting of the timeline view and the datepicker labels when selecting annual datasets. The labels should show only in `yyyy` instead of `MM yyyy` format when annual datasets are the lowest common denominator selected.

### Validation / Testing

- Add a dataset with annual time density and verify that the timeline view labels and the datepicker labels show only the year
- Add datasets with month time density, and verify that now the labels show the month and the year
- The rest should work as before (date picker shows the appropriate calendar view)


hanbyul-here (approved):
Thanks for handling this issue 🙇

Issue #1095, opened 2024-08-07
[Bug] `externalLinksInNewTab` in `veda.config.js` isn't respected anymore


slesaad:
### Context

URL: https://earth.gov/ghgcenter/data-catalog/eccodarwin-co2flux-monthgrid-v5

Blocking a release? (y/n): n

Regression issue? (y/n): y
[If yes, add Regression label]

Priority: It feels like a minor annoyance but if it's not too much work,  we would like to get it in for the next release

### Steps to reproduce the issue

1. Go to the URL ☝️ 
2. Click on one of the external links at the bottom of the page
3. It opens the link in the same tab when it should open in a new tab

### Other details
It looks like it was introduced in `veda-config-ghg@v2.0.9` which referenced `veda-ui@v5.3.0`

hanbyul-here:
👋 How important is to keep the configuration variable? Would it be okay to open the external link in a new tab (without any configuration)?

PR #1116, opened 2024-08-20
open an external link in a  new tab

hanbyul-here:
**Related Ticket:**  #1095 

Instead of using configuration, (I think it is a bit weird thing to configure on the instance level) I just made all the external links to be opened in the new tab.

### DIFF
--- a/app/scripts/utils/url.ts
+++ b/app/scripts/utils/url.ts
@@ -1,5 +1,4 @@
 import React, { MouseEventHandler } from 'react';
-// import { getBoolean } from 'veda';
 import { LinkProps } from 'react-router-dom';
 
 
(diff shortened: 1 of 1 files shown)


dzole0311 (approved):
I liked the fact that the link can be configured to open in a new tab, and USWDS supports it as a pattern: https://designsystem.digital.gov/components/link/

But this is fine as well and we could add a config once it's needed 👍

hanbyul-here:
Oh, thanks for pointing out the link style for external links in USWDS. I remember @faustoperez having a similar idea and design, but we never got to it (we should!). 
I will go ahead and merge this fix, but I have some thoughts. 


I decided to deprecate `externalLinksInNewTab` 
- I wanted to avoid  the prop drillings required to make it work especially when all the existing instances likely want the same behavior.
- I think UI should decide (design) how the component should behave 🤔 so I concluded that an instance should not decide the component level behavior. (This thought can be better shaped though - like what is the line between customization and behavior control?) 
- if we get to the routing problems raised by refactoring work,  the instance will pass the Link components that they want. (which might be soon.)

faustoperez:
I know that this is answering a request and I might have expressed other opinions in the past 😅 , but my preference (following @ricardoduplos steps) would be to have all links open in the same tab, more so now that we are marking all the external links with an icon. 

Opening a new tab is more cumbersome on mobile, and on desktop users can always right-click and open a new tab anyway if that's their preference 🤷‍♂️

PR #1120, opened 2024-08-22
Make focus style more subtle

hanbyul-here:
This PR makes the focus style from uswds more subtle. This PR also adds a mixin `only-focus-visible`, which gives focus style only for keyboard interactions.

### DIFF
--- a/app/scripts/styles/_mixins.scss
+++ b/app/scripts/styles/_mixins.scss
@@ -1,3 +1,5 @@
+@use "uswds-core/src/styles/functions/utilities" as utils;
+
 @mixin only-focus-visible {
   &:focus {
     outline: none;
--- a/app/scripts/styles/_mixins.scss
+++ b/app/scripts/styles/_mixins.scss
@@ -3,7 +3,7 @@
 @mixin only-focus-visible {
   &:focus {
     outline: none;
-    box-shadow: 1px #ccc;
+    box-shadow: 1px utils.color('gray-5');
   }
   &:focus-visible {
     outline: 1px solid;
--- a/app/scripts/styles/_mixins.scss
+++ b/app/scripts/styles/_mixins.scss
@@ -1,0 +1,16 @@
+@mixin only-focus-visible {
+  &:focus {
+    outline: none;
+    box-shadow: 1px #ccc;
+  }
+  &:focus-visible {
+    outline: 1px solid;
+  }
+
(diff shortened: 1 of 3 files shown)


app/scripts/styles/_mixins.scss
```diff
@@ -0,0 +1,16 @@
+@mixin only-focus-visible {
+  &:focus {
+    outline: none;
+    box-shadow: 1px #ccc;
```
hanbyul-here, line 4: I could not use the mixins from uswds here 🤔 (If I `@use "uswds-core/src/styles/functions/utilities" as utils` do this, scss compiler will complain on the instance level.) Do you know why? @dzole0311 ?

dzole0311: Can you paste the compiler error that you see? I added back `@use "uswds-core/src/styles/functions/utilities" as utils` and tried it locally on the GHG instance and pushed remotely [here](https://github.com/NASA-IMPACT/veda-config/pull/439), but I don't see any errors.

dzole0311 (approved):

app/scripts/styles/_mixins.scss
hanbyul-here: Yeah! sorry for not being clear. I just checked it again, and including the path itself is not a problem - but when I try to use the value (https://github.com/NASA-IMPACT/veda-ui/pull/1120/commits/9100e0da2441ea79ca4960b240364e51d5017269) - and run the application on the instance level, it throws an error that 
```
@parcel/transformer-sass: Can't find stylesheet to import.
  ╷
1 │ @use "uswds-core/src/styles/functions/utilities" as utils;
```

My hunch is - the app detects that it is running on the instance level so it finds uswds from `.veda/ui/node_modules....` but then the uswds imported from UI level gets confused because it doesn't have a path to `.veda/ui/...` ?

dzole0311: Ah, that might be the issue. Maybe we could create an alias for the `uswds` utils or in general for `uswds`-related imports (similar to $types, $utils etc.) in package.json and use the alias in our .scss files?

hanbyul-here:
First, This PR doesn't need to be merged. I worked around by putting the style on the instance side without using mixins. 

Second, I pushed the branch trying to use mixins : 
https://github.com/US-GHG-Center/veda-config-ghg/pull/506

PR #1121, opened 2024-08-23
Consolidate a place where to check linkProps

hanbyul-here:
**Related Ticket:** Very vaguely related to #946 

This PR makes the Card component delegate what link component to render to the `SmartLink` component. I initially started working on this PR to continue the work on https://github.com/NASA-IMPACT/veda-ui/pull/1096, more specifically, to make the link component come from props.  I thought this change was worth merging regardless of whether we work on the routing problem now or later.

### DIFF
--- a/app/scripts/components/common/element-interactive.js
+++ b/app/scripts/components/common/element-interactive.js
@@ -1,9 +1,8 @@
 import React, { useCallback, useState } from 'react';
 import T from 'prop-types';
 import styled from 'styled-components';
-import SmartLink from './smart-link';
 
-const InteractiveLink = styled(SmartLink)`
+const InteractiveLink = styled.a`
   position: absolute;
   inset: 0;
   z-index: -1;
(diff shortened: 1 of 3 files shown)


sandrahoang686:
This will cause conflicts with exposing Data Catalog PR [here](https://github.com/NASA-IMPACT/veda-ui/pull/1096/files#diff-c618b2febe87cadd3d3931c1498d8b40c67f5793d0235e3777dd79a202183f34). In that PR, I just have the parents pass in the the JSX element tag into the `ElementInteractive` component because nextJs was having issues with the `InteractiveElement` styled component but do you think this will solve for that? Do you think its best to resort to this instead of what I have currently in my PR?

app/scripts/components/common/element-interactive.js
```diff
@@ -109,6 +118,8 @@ export const ElementInteractive = React.forwardRef(
 );
 
 ElementInteractive.propTypes = {
+  linkTo: T.string.isRequired,
+  onLinkClick: T.func,
```
sandrahoang686, line 122: Is there a reason you decided to add this prop now? I see `getLinkProps` already accepted an `onClick` param but wondering if you had a specific reason?

hanbyul-here: ah, that was because of how I was passing the props for SmartLink before. But your comment made me think that I don't have to do that! Thanks!

sandrahoang686 (approved):

Issue #953, opened 2024-05-10
Cookie Consent Alert


aboydnw:
## Context

We use cookies to track usage with google analytics. Instances that use google analytics should have a cookie banner to give users the option to reject cookies.

## Acceptance Criteria

- [ ] Given that an instance has turned on the cookie banner, when a user visits that dashboard and has not previously provided an explicit response to the cookie alert, then they are presented with a cookie alert
- [ ] Given a user visits an instance with the cookie alert who has not previously accepted the use of cookies when they ACCEPT cookies then we continue to track using google analytics 
- [ ] Given a user visits and instance with the cookie alert who has not previously accepted the use of cookies when they REJECT the use of cookies then we don't track their usage with google analytics
- [ ] Any links in the alert content open in new tabs
- [ ] If the user elects to close the alert without making an explicit selection, we assume that as a yes and continue collecting GA data, but ask the user again the next time they visit the site

## Related Tickets

- https://github.com/US-GHG-Center/veda-config-ghg/issues/162

aboydnw:
@snmln I made some tweaks to the AC based on the convo in the PR, let me know if anything looks out of place

snmln:
Linking PR: https://github.com/NASA-IMPACT/veda-ui/pull/1127

PR #1127, opened 2024-08-26
Creating cookie consent component 

snmln:
**Related Ticket:**[ Cookie Consent Form](https://github.com/orgs/NASA-IMPACT/projects/17/views/6?pane=issue&itemId=62543604)

### Description of Changes
USWDS Alert component used as a cookie consent form. This form should render on any page of the site if a user has not consented to the use of cookies. If a user _Accepts Cookies_ or _Declines Cookies_ a cookie will be made to catalog the response by the user. The cookie will expire after 3 months and the form will re-render for the user.

Figma Designs: https://www.figma.com/design/xaZSp74DKFGYm0k2w2BbVb/Shared-VEDA-file?node-id=0-1&t=wGNpe3xrUyLTU1pB-1

CookieConsent form is stored behind a feature flag to allow for each instance to opt in to leveraging the component. To turn the CookieConsentForm **ON** in the `veda-ui/.env` set `COOKIE_CONSENT_FORM='TRUE'`

To change contents of the consent form change the contents of `cookieConsentForm` within the `veda-ui/mock/veda.config.js` to reflect desired content. To add link in the content, it must be added in the following format `[Link text](URL)`

**Example prop:** 'We use cookies to enhance your browsing experience and to help us understand how our website is used. These cookies allow us to collect data on site usage and improve our services based on your interactions. To learn more about it, see our `[Privacy Policy](https://www.nasa.gov/privacy/#cookies)`'

### Validation / Testing
Run unit test CookieConsent.spec.js. To test in browser, run locally navigate to browser cookies you should see the following values for specific interactions: 
- **Decline Cookie:** {"responded":true,"answer":false}
- **Accept Cookie:** {"responded":true,"answer":true}
- **Close out:** {"responded":false,"answer":false}


aboydnw:
I think this looks great. Seeing this for me raises two questions:
1. Should we also include a close button? 
2. After somebody takes any of these actions, what happens next time they visit the site? Would they see the alert again or do we have a way of recognizing that they already answered the form?

Also, I know it's just content and not what you asked for feedback on, but should the Privacy Policy link open in a new tab?

snmln:
> I think this looks great. Seeing this for me raises two questions:
> 
> 1. Should we also include a close button?
> 2. After somebody takes any of these actions, what happens next time they visit the site? Would they see the alert again or do we have a way of recognizing that they already answered the form?
> 
> Also, I know it's just content and not what you asked for feedback on, but should the Privacy Policy link open in a new tab?

We can definitely introduce a close out button in the interface for those that would rather not respond. It does then prompt the question, do we interpret a close out of the interface as a _decline_ or an _non-answer_ and continue to collect the cookies from their session. Based on the Nasa.gov data privacy site (which is somewhat vague) I would interpret it as a non-answer and continue collecting: 

> Each NASA site using persistent cookies identifies itself as doing so. If you do not wish to have session or persistent cookies stored on your machine, you can turn them off in your browser. However, this may affect the functioning of some NASA Web sites.

As for the next time the user comes to the site, it depends on what type of cookie we are using if they are persistent cookies the user is good to go, no reoccurring interface on their end unless they delete their browsing history. We can grab that information via their browser during each session. As for opening a link in a new tab, yes the privacy link should open in a new tab, that is something we can note in the acceptance criteria once we move to development.

aboydnw:
> Based on the Nasa.gov data privacy site (which is somewhat vague) I would interpret it as a non-answer and continue collecting

That seems reasonable to me 👍

app/scripts/components/common/cookie-consent/index.scss
```diff
@@ -0,0 +1,8 @@
+.modal {
+    z-index: 999;
+    position: fixed;
+    max-width: 725px;
+    border-radius: 35px;
```
dzole0311, line 5: Could we use some of the radius classes from USWDS? https://designsystem.digital.gov/utilities/border/#utility-radius

dzole0311 (commented):
The POC looks great @snmln, welcome to the project!

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -0,0 +1,127 @@
+import React, { useState, useEffect } from 'react';
+
+import {
+  Alert,
+  Button,
+  ButtonGroup,
+  Link,
+  Icon
+} from '@trussworks/react-uswds';
```
sandrahoang686, line 9: we have a `uswds` directory where instead of importing components directly from `uswds`, we create an abstraction layer and import them from this directory instead. This would help us visually identify right away what components we depend on from the DS. To decouple the DS component(s) quicker if need be. If any breaking changes is updated in the DS, we have the ability to just handle it in one place rather than multiple where direct imports are happening. Looks like we haven't been keeping to that standard though. Still early on, so i recommend we do.

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export const CookieConsent = ({
+  title,
+  copy,
+  onFormInteraction
+}: CookieConsentProps) => {
+  const [cookieConsentResponded, SetCookieConsentResponded] = useState(Boolean);
+  const [cookieConsentAnswer, SetCookieConsentAnswer] = useState(Boolean);
+  const [closeConsent, setCloseConsent] = useState(Boolean);
```
sandrahoang686, line 25: Interesting, never initialized state like this. Is there a benefit to doing it like this? Looked into it and looks like it'd just be more beneficial to be declarative. `const [cookieConsentResponded, SetCookieConsentResponded] = useState<boolean>(false);`

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+          </Link>
+        );
+        /* eslint-enable react/no-array-index-key */
+
+        return newItem;
+      }
+      return item;
+    });
+    return updatedContent;
+  };
```
sandrahoang686, line 63: could benefit from memoization here, dependencies on `copy`

parcel-resolver-veda/index.d.ts
```diff
@@ -334,6 +339,8 @@ declare module 'veda' {
   export const getBoolean: (variable: string) => boolean;
 
   export const getBannerFromVedaConfig: () => BannerData | undefined;
+  export const getCookieConsentFromVedaConfig: () => CookieConsentData;
```
hanbyul-here, line 342: Can the return value be undefined?

.env
```diff
@@ -13,4 +13,6 @@ API_STAC_ENDPOINT='https://staging.openveda.cloud/api/stac'
 GOOGLE_FORM = 'https://docs.google.com/forms/d/e/1FAIpQLSfGcd3FDsM3kQIOVKjzdPn4f88hX8RZ4Qef7qBsTtDqxjTSkg/viewform?embedded=true'
 
 FEATURE_NEW_EXPLORATION = 'TRUE'
-SHOW_CONFIGURABLE_COLOR_MAP = 'FALSE'
\ No newline at end of file
+COOKIE_CONSENT_FORM = 'TRUE'
```
hanbyul-here, line 16: If this is something that can be tweaked through `veda.config`,  another env flag here seems to be redundant. What do you think?

app/scripts/components/common/cookie-consent/index.scss
```diff
@@ -0,0 +1,24 @@
+.modal {
```
hanbyul-here, line 1: this style is breaking the style of other modals : ex. https://deploy-preview-1127--veda-ui.netlify.app/exploration?datasets=%5B%5D&taxonomy=%7B%7D

There are many ways to resolve this problem, but I wonder if this needs to be modal at all. (Considering modal usually includes the blocking of the rest of the screen.) What do you think? Was there any other component you considered using ?

app/scripts/components/common/cookie-consent/cookieConsent.spec.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
hanbyul-here, line 1: 👏 Amazing! Thanks for putting the first spec test!

hanbyul-here (changes requested):
Thanks for putting the work on it 🙇 The main problem I found is this style is breaking the other modal's style: https://deploy-preview-1127--veda-ui.netlify.app/exploration?datasets=%5B%5D&taxonomy=%7B%7D  

I left a comment, but I wonder if it is a modal at all 🤔 Have you considered not using Modal at all? Especially since the component was not even used at all!

app/scripts/components/common/cookie-consent/index.scss
snmln: Oooh good catch! No there is no need for this to be named `modal` any longer I think that was just a naming convention issue. I have gone ahead and swapped the implementation to `id=cookie-consent`. As for other component we did look at a modal but out of the box the modals are very disruptive to the user experience as a whole and Fausto and I decided against pursuing that direction.

.env
snmln: I believe so, I think there is some confusion on my end concerning the relationship between the `veda-ui` and `veda-config` codebases and the ultimate implementation of the feature flags. I could probably benefit from a quick connect to correct some assumptions on my end.

app/scripts/components/common/cookie-consent/index.scss
```diff
@@ -0,0 +1,25 @@
+#cookie-consent{
```
hanbyul-here, line 1: Have you had a chance to familiarize yourself with utility functions (such as https://designsystem.digital.gov/utilities/display/, https://designsystem.digital.gov/design-tokens/z-index/ ) in USWDS? I think a lot of this custom style can be replaced with the USWDS utility class - and this will also give us some coherence on the units.

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -35,17 +43,58 @@ const PageBody = styled.div`
 `;
 
 function LayoutRoot(props: { children?: ReactNode }) {
+  const useConsentForm = checkEnvFlag(process.env.COOKIE_CONSENT_FORM);
```
hanbyul-here, line 46: this is still using env variable. And I think your intention is to depend on veda config instead of env var?

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -35,17 +43,58 @@ const PageBody = styled.div`
...
+
+  const readCookie = (name) => {
```
hanbyul-here, line 48: Any reason that this function is repeated in two files? 
https://github.com/NASA-IMPACT/veda-ui/pull/1127/files#diff-e42ad7e47cbcdcd4c87a29eea61c27201f72b653eb6484df90d4666f41682658R1
Since there is util file already, wouldn't it be better to consolidate these functions in util file?

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -35,17 +43,58 @@ const PageBody = styled.div`
...
+      const cookie = readCookie('CookieConsent');
+      if (cookie != null) {
+        cookieContents = JSON.parse(cookie);
+
+        cookieContents.answer && setGoogleTagManager();
+
+        return cookieContents;
+      }
+    }
+    cookieContents = 'NO COOKIE';
```
hanbyul-here, line 72: I think this function can be written 

```

const getCookie = () => {
  const cookie = readCookie('CookieConsent');
  if (cookie) {
    const cookieContents = JSON.parse(cookie);
    if (cookieContents.answer)  setGoogleTagManager();
    return cookieContents;
  }
  return 'NO COOKIE';
};
```
(continuing in the comment below)

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -35,17 +43,58 @@ const PageBody = styled.div`
...
+  }
+
+  const cookieConsentContent = getCookieConsentFromVedaConfig();
+  const showForm = () => {
+    getCookie();
```
hanbyul-here, line 77: ☝️  so this function can use the return value from getCookie without going outside of the function scope. What do you think?

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -35,17 +43,58 @@ const PageBody = styled.div`
...
+
+    if (cookieContents === 'NO COOKIE') {
+      return true;
+    } else {
+      return !cookieContents.responded;
```
hanbyul-here, line 82: If the user says no to cookie usage, can we still save this information to cookie?

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  copy?: string | undefined;
+  onFormInteraction: () => void;
+}
+
+export const CookieConsent = ({
+  title,
+  copy,
+  onFormInteraction
+}: CookieConsentProps) => {
+  const [cookieConsentResponded, SetCookieConsentResponded] =
```
hanbyul-here, line 24: I would like to learn more about where to save cookie consent prefrence. We have been saving the user preference on local storage so far - can you leave some notes about why we are using cookie for this case? Moreover, when user says no to cookie usage, can we save this info to cookie at all?

hanbyul-here (commented):
Thanks for working on it, I left some comments that hopefully helps you. + It will be good if you can review the code and clean up a bit (remove unnecessary comments etc.)

app/scripts/components/common/layout-root/index.tsx
snmln: This would fall under a strictly necessary cookie that is accepted by the GDPR rules: https://gdpr.eu/cookies/#:~:text=Cookie%20compliance,cookies%20except%20strictly%20necessary%20cookies.

app/scripts/components/common/layout-root/index.tsx
snmln: This was a change that fell through the cracks while rebasing. There should only be 1 instance of this function.

app/scripts/components/common/layout-root/index.tsx
snmln: 100% agree!

app/scripts/components/common/cookie-consent/index.scss
snmln: I did, but wasn't familiar with all of the available tokens. I will get those adjusted.

app/scripts/components/common/cookie-consent/index.tsx
snmln: We can change this to local storage if that is the existing pattern across the system. 

Based on what I could find through my original research, the prevailing pattern is to store the consent in a cookie since they require an expiration associated with the cookies. Allowing for redundancy when a user clears their cache and cookies they will be re-prompted with cookie consent interface. Because of the temporary nature it appears to be the primary storage method (while contradictory to the implied practice).

hanbyul-here:
As I mentioned in stand-up let's merge this after v5.8.0 goes out. But there are several improvements that I'd like to note
(This PR is already pretty big, so these things can be followed up as separate PRs.)


1. Currently, the scripts that check cookies run every time a new page loads which I am not sure if it is necessary.

2. (might be related to 1.) I feel like Layout component is doing too much of cookie-related works, which should not be a part of layout's job. It might be slightly complicated in real life since it is layout's job to decide whether to render a specific component or not. But maybe there is something in the middle ground. 

3. (This is partially my fault that I rushed the change) Banner and CookieConsent expects different forms for its content, Banner expects html and CookieConsent expects markdown. I am fine with either of them, but we should consolidate them.

aboydnw:
@snmln can you create a ticket in veda-ui to represent this followup work? Might be good practice, and I can help you fill out any metadata on the ticket if you have questions

snmln:
> @snmln can you create a ticket in veda-ui to represent this followup work? Might be good practice, and I can help you fill out any metadata on the ticket if you have questions

Of course!

snmln:
> > @snmln can you create a ticket in veda-ui to represent this followup work? Might be good practice, and I can help you fill out any metadata on the ticket if you have questions
> 
> Of course!

@aboydnw  I have gone ahead and create 2 tickets to reflect the work: 
[Cookie Consent - fast follow](https://github.com/NASA-IMPACT/veda-ui/issues/1170)
[Cookie Consent - Clean up](https://github.com/orgs/NASA-IMPACT/projects/17/views/6?pane=issue&itemId=80952423)

Please let me know if I need to adjust any tagging or anything else for these tickets.

aboydnw:
Thanks @snmln . I wonder if the action of merging this PR is covered by #953 that is In Review right now, then we wouldn't need #1170 and can schedule #1171 for an upcoming sprint

What do you think?

snmln:
@aboydnw I agree with that plan! I will go ahead and delete https://github.com/NASA-IMPACT/veda-ui/issues/1170

hanbyul-here (approved):
I found an error that I didn't use the constant for setting cookie ! I pushed a fix for it: https://github.com/NASA-IMPACT/veda-ui/pull/1127/commits/9356b8750f8dbc2230b7bcb3d8bf91a8f2de35f4

I think this should be an ok status to merge, but I touched it too much at this point 😅 so please take time to test the functionality before merging.

Issue #994, opened 2024-06-07
[E & A] Implement colormap configurability


aboydnw:
### Context

An implementation ticket for the design ticket here: #929 

### Acceptance Criteria
- [ ] Users are able to change the color map to one of the options we provide them
- [ ] Color map defaults are able to be defined by data providers and identified in the app, even after changing
- [ ] Users can edit the rescale values to change the visualization of the data
- [ ] Use new titiler API, backend titiler update: https://github.com/NASA-IMPACT/veda-backend/pull/398

aboydnw:
Once we close this, we can close a related GHG ticket: https://github.com/US-GHG-Center/feature-requests/issues/3

faustoperez:
Hey @dzole0311 👋 

Based on the team conversations we had this week I updated the designs of the Colormap options selector:

- Changed the icon to one from the Collecticons library
- Added a Reset button. The Reset button should be enabled once the user has changed an option.

Link to Figma 👉 https://www.figma.com/design/9INQauBWhiRxvOWDGhRrxO/US-GHG-Center?node-id=1781-3165&t=8tmNwMDSmSA7hY3c-1


<img width="2880" alt="Desktop Flow" src="https://github.com/user-attachments/assets/d8a75dcd-c4c7-4a3a-b145-5f8695e3a01d">

dzole0311:
Thanks @faustoperez - should be easy to add the 'Reset' option. 

I couldn't find the icon in the [USWDS icon list](https://designsystem.digital.gov/components/icon/). Is it coming from there? If not, could we switch to one from the USWDS collection?

faustoperez:
> I couldn't find the icon in the [USWDS icon list](https://designsystem.digital.gov/components/icon/). Is it coming from there? If not, could we switch to one from the USWDS collection?

I didn't find any suitable icon on the USWDS list, that's why I proposed one from Collecticons. If this is a concern, let's skip the icon for now 🙂

j08lue:
@faustoperez - when reading the legend, I am unable to see at a glance, what the units are. They are only part of the list of properties under the dataset title. Would you consider to somehow associate them more closely with the legend?

Sorry this comment is coming so late...

faustoperez:
That's a good call Jonas, I did play with the position of the units on this card design. In the previous design sometimes the legend was truncated depending on the actual length of the colormap values. 

I wanted to leave space on the colormap legend in case we had long numbers, or divergent scales with a zero in the center.

[...]

PR #1128, opened 2024-08-27
Return default value of colormap along with other settings

hanbyul-here:
**Related Ticket:** #994 

### Description of Changes
I made the colormap default to be returned along with other settings values - I think this saves some redundancies of retrieving the value of colormap, and fits well with the existing pattern. 
I eliminated reconciling functions from data-util (and used the function from data-util-no-faux-module) so there is no confusion about which function is being called.


dzole0311 (approved):
Thanks for the change @hanbyul-here 🙇  I additionally removed the "Reset All" button and the "Default" tag.

Issue #1056, opened 2024-07-17
Replace cmr-stac with titiler-cmr


abarciauskas-bgse:
## Context

To support https://github.com/NASA-IMPACT/veda-ui/issues/1054 we will replace the cmr-stac layer with the titiler-cmr layer.

## Acceptance Criteria

- [ ] Given I add a titiler-cmr layer, I should be able to visualize it in the exploration page.

PR #1131, opened 2024-08-28
Feat/use titiler cmr

abarciauskas-bgse:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1056

### Description of Changes
This change was actually relatively straightforward. Titiler-cmr handles the business of searching CMR for which assets to tile, which means that logic no longer needs to exist in veda-ui. We can just pass the concept id and other parameters directly to titiler-cmr. The asset discovery logic _was_ happening in the useCmr hook which searched [CMR STAC endpoints](https://cmr.earthdata.nasa.gov/stac) (endpoints plural because there are different catalogues for different DAACs) for data. This was especially messy given the S3 urls are not in the CMR STAC item representations, so we had to replace an https protocol and host with the S3 protocol and bucket name. 

Itemized changes:
* Remove useCmr hook
* Change parameterization to RasterPaintLayer to make the url optional: titiler-xarray requires a url parameter, whereas titiler-cmr does not. 
* Removed TRMM layer as this dataset is not yet in VEDA STAC (necessary to be included in the dashboard)

### Notes

* dev-titiler-xarray.delta-backend.com uses `datetime` which is consistent with titiler-cmr, however this change has not yet been deployed to prod-titiler-xarray.delta-backend.com.

### Validation / Testing
[Loaded and toggled both layers in the deploy preview](https://deploy-preview-1131--veda-ui.netlify.app/exploration?datasets=%5B%7B%22id%22%3A%22GPM_3IMERGDF.v07%22%2C%22settings%22%3A%7B%22isVisible%22%3Afalse%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%2C%7B%22id%22%3A%22combined_CMIP6_daily_GISS-E2-1-G_tas_kerchunk_DEMO%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2006-12-22T08%3A00%3A00.000Z)


hanbyul-here:
Yay~ thanks for the great work. I have some questions about titiler parameters for these datasets! 

While resolving the conflict, I noticed that `rescale` values for both `zarr` and `cmr` are configured as string, not as an array. (ex. GPM IMERG  `rescale: 0,46`) . When I check `renders` attribute from the stac endpoint: https://staging.openveda.cloud/api/stac/collections/GPM_3IMERGDF it returns the scale value as an array of array. So, I tried to change the configuration to use an array for rescale values, but I can't seem to get the tiles loaded with an error that `  "detail": "could not convert string to float: '[[0'"
`. (https://dev-titiler-cmr.delta-backend.com/tiles/WebMercatorQuad/1/0/0@1x?datetime=2024-08-30T04%3A00%3A00.000Z&resampling=bilinear&variable=precipitation&colormap_name=gnbu&rescale=%5B%5B0,46%5D%5D&concept_id=C2723754864-GES_DISC&backend=xarray) 

We recently started using `renders` attributes from renders extensions and assumed that all the datasets would have rescale values as an array of arrays. Is the expectation different for Zarr and CMR datasets? We can work around it, but I would like to confirm the different expectations for different datasets!

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
```diff
@@ -43,11 +43,10 @@ export function RasterPaintLayer(props: RasterPaintLayerProps) {
 
   useEffect(
     () => {
-      if (!assetUrl) return;
-
+      // in the case of titiler-cmr, there is no asset url. The asset urls are determined internally by titiler-cmr.
       const tileParams = qs.stringify({
-        url: assetUrl,
-        time_slice: date,
+        ...(assetUrl && { url: assetUrl }), // Only include `url` if `assetUrl` is truthy (not null or undefined)
```
hanbyul-here, line 48: 🤔  I wonder if it is best if the raster-paint-layer doesn't need to care about the specifics of the layer- How about we move this logic to Zarr-time-series? 
I picture something like the below in zarr-timeseries layer
```
  const zarrSourceParams = { ...sourceParams, assetUrl};
 return <RasterPaintLayer {...props} sourceParams={zarrSourceParams} />;
```

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
```diff
@@ -29,7 +29,7 @@ export function RasterPaintLayer(props: RasterPaintLayerProps) {
 
   const { updateStyle } = useMapStyle();
   const [minZoom] = zoomExtent ?? [0, 20];
-  const generatorId = `zarr-timeseries-${id}`;
+  const generatorId = `raster-timeseries-${id}`;
```
hanbyul-here, line 32: a nitpick, but it might make sense to let each parent layer pass the generatorPrefix to keep the original way of signaling which layer is which through generator Id

abarciauskas-bgse:
@hanbyul-here thanks for reviewing!

In response to https://github.com/NASA-IMPACT/veda-ui/pull/1131#issuecomment-2321600823, you are right that in the dataset MDX files I am passing the rescale value for the rescale parameter as a string, not as an array. That is because the tile API, for titiler-xarray (zarr), titiler-cmr (cmr), and VEDA titiler accept the rescale parameter as a comma-delimited string, not an array of arrays. Looking at the [titiler changelog](https://developmentseed.org/titiler/release-notes/#032-2021-05-26) and the [documentation] it looks like the API expects multiple rescale parameters for multiple bands, e.g. `rescale=0,1000&rescale=0,1000&rescale=0,1000`

So I would expect, in the case that we are using data from the STAC renders extension to set tiling parameters, that we need to breakdown the array of arrays into separate rescale parameters. Can you point me to where the parameters are being set by the renders extension?

A valid question though is should my code also be using the renders extension (at least, when it exists and is not set in the MDX file which would override the defaults set in STAC metadata) to set these tiling parameters 🤔 . I think it should so I could open a new issue for that!

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
abarciauskas-bgse: yes I had it that way at one point actually so happy to make this change.

hanbyul-here:
Ah, I misunderstood in a way that renders extensions format === titiler inputs.  You are right that it was about formatting query aprameters - I opened a pr for it : https://github.com/NASA-IMPACT/veda-ui/pull/1140 Thanks for spotting !! 🙇 

re: using render extension - All the datasets are being reconciled with STAC data before reaching layers already! (The reconciled data includes render params - but we are giving priority to the input from configuration file. We fall back to render params only when the configuration for sourceParams is missing - we will gradually deprecate manual configuration ! )

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
abarciauskas-bgse: yes that makes sense and I made that change here: https://github.com/NASA-IMPACT/veda-ui/pull/1131/commits/aa6e9b882c66d0544bad9ccbd38a7c95aabf867e. I don't fully understand how the generatorId is used however so I can't really test it (other than it doesn't seem to break anything). Could you advise me on if there is a way to manually test that the generatorId is being set? And how is it being used out of curiosity?

abarciauskas-bgse:
@hanbyul-here I'm taking a look at #1140 I was confused at first because I didn't notice it was a PR to this branch _and_ includes a bit of history we don't have here. I'm not sure why yet but I'll look into it.

abarciauskas-bgse:
@hanbyul-here ok I believe the reason I was seeing so many commits was because you had merged main into feat/use-titiler-cmr and some conflicts needed to be resolved against my changes. I had to add the `arrayFormat: "comma"` argument to qs.stringify and now things are working nicely. Let me know what you think.

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
hanbyul-here: As far as I understand, generator Id is used to uniquely identify the layers for mapboxgl. (but the entire generator id is also suffixed by layer-id - which should be unique - so it will be only useful if there are layers with the same ID but with different data types) And it might come handy if I face a messy situation ex. I have to check  layer IDs on the dashboard? It is there for more of a practice rather than actual use at this point.

abarciauskas-bgse:
@hanbyul-here do you have any other feedback or is this good to merge?

hanbyul-here (approved):
Apologies that it took some time to review this PR. Thanks for working on it 🙇

Issue #1024, opened 2024-07-01
Spike: Automate the testing of one core feature in the veda-ui


dzole0311:
## Context

We have a simple and non-exhaustive [list](https://docs.google.com/document/d/1ig9qIv-73IL3TdnqLQ05RR35woxB_TouplDLVY37OXI/edit) of some core QA scenarios for the Dashboard. What would be useful is to try to automate one of them using Cypress or Playwright, so that it runs as part of a github action.

## Acceptance Criteria

- [ ] Implement a test scenario of a core feature in the veda-ui
- [ ] Run it as part of a github action

## Related Tickets

[If applicable, link any tickets that are related]

anayeaye:
Could there be any overlap and/or is this an opportunity to collaborate on [the playwright tests designed for testing out titiler and data catalog changes](https://github.com/NASA-IMPACT/veda-config/pull/363)? For context: the backend team is interested in integration testing to ensure stac-api and titiler-api changes support veda-ui. There are also plans to integrate some GHG test requirements that have been [previously implemented with selenium](https://github.com/US-GHG-Center/veda-config-ghg/tree/ui-test/tests).

cc: @stephenkilbourn @smohiudd

stephenkilbourn:
thanks for bringing it up @anayeaye. From an initial look, the Playwright tests I created in Veda-config cover the first 3 points and add in Page Objects to allow several of the other points.  I leveraged Playwright since it now is more widely used than Cypress and allows parallel, cross browser testing for free, but a similar test structure could be used for Cypress.

I'm happy to set up some time and talk through what is a good path for the UI so that we don't duplicate work.

dzole0311:
Thanks for the info @anayeaye @stephenkilbourn! 

I also had some concerns on how the architectural changes we're currently working on in the veda-ui repo, with the aim of it becoming a ui components library, and moving the hardcoded routes from the veda-ui repo to the veda-config instance for a more config flexibility would affect the overall testing strategy we choose now.

So I think a short call would be great, I'll set something up.

dzole0311:
Thanks all for the great call today! Summary of key points discussed (please correct me if I missed something)

1. Playwright vs Cypress: Playwright appears more powerful, offers parallel test execution, reduced need for chaining selectors
[...]

PR #1139, opened 2024-08-29
chore: add playwright

stephenkilbourn:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1024 https://github.com/NASA-IMPACT/veda-ui/issues/1197

### Description of Changes
This adds Playwright tests of some key user flows.  The tests are set to run on the github action. On failure, an html report along with traces containing screenshots and network calls will be attached to the action's step.

The tests will run only on release branches (branches starting with `release`).

### Validation / Testing
tests can be run locally with `yarn test:e2e` and they can also be seen in the github action.


stephenkilbourn:
It looks like I don't have access to view/change github secrets.  This is expecting a mapbox token stored as `MAPBOX_TOKEN` like in the veda-config.

dzole0311:
Thanks @stephenkilbourn for the great work and getting the tests green! I'm considering adjusting the github job to run on release branches only instead of each feature branch, to keep the fast development lifecycle. @sandrahoang686 @AliceR @hanbyul-here @snmln – any opinions before we merge these?

dzole0311 (approved):

sandrahoang686:
@stephenkilbourn thanks for setting this up, tests will be pretty crucial going forward and we could definitely spend more time here. Just a few thoughts -
* Do we want routing tests here? For those tests, we'd probably want at the instance level?
   * Path would eventually change to `await page.goto('/data-catalog');` for example this path would eventually come from instance
* At the instance level we orchesta e2d behavior so would we want to move more towards integration testing? 

cc @dzole0311 maybe you have more thought out thoughts here though

stephenkilbourn:
@sandrahoang686 @dzole0311 

We do have existing tests at the instance level in [veda-config](https://github.com/NASA-IMPACT/veda-config/pull/363) and [GHG.](https://github.com/US-GHG-Center/veda-config-ghg/tree/develop/e2e) My intent was to move the common ones down to the UI level to remove duplication. We can then have tests against anything unique to that instance live in the config repos.

I can set up the routes to read from an env if you think that makes sense.  I can set in the playwright config to read from the env and default to the route that is in the current UI. 
something along the lines of 
```
const catalogRoute = process.env.CATALOG_ROUTE ? process.env.CATALOG_ROUTE : 'data-catalog`
await page.goto(catalogRoute);
```

dzole0311:
@sandrahoang686 Yes I think we can move away from routing in the tests as we break out the components. But for consistency with our current architecture, I propose we keep it for now.

Since this PR provides a solid base for more extensive Playwright tests in the veda-ui, I’ll go ahead and merge it so we can start playing with them and using them into our dev workflow, and make modifications as needed. Thanks @stephenkilbourn for setting these up!

We have a follow-up here: https://github.com/NASA-IMPACT/veda-ui/issues/1200

PR #1145, opened 2024-09-04
FIX: update external link in top navigation target

stephenkilbourn:

### Description of Changes
this updates the `nav-menu-item` with external links to have  `target="_bank"`  instead of  `target="blank"`.  There's [a slight difference in the behavior](https://css-tricks.com/targetblank/).
and other external links have used the underscore.


hanbyul-here (approved):
Thanks for handling this 🙇

Issue #1142, opened 2024-09-03
What to do with configurable colormap when GHG instance doesn't want it for R3


aboydnw:
With timing of R3, GHG doesn't want to pull in colormap feature because they want to focus their efforts on finalizing R3 as opposed to testing. So, we want to make sure they can uptake the latest version of veda-ui for any updates that support R3, without needing to use colormap configurability.

Plan is to use a feature flag or just comment in the code, or a hardcoded variable to show colormap.

PR #1147, opened 2024-09-05
Add a flag not to show the colormap

hanbyul-here:
**Related Ticket:** #1142 

### Description of Changes
Turn off colormap configurability. It is default to off, but instances can turn it on by setting up environment variable `SHOW_CONFIGURABLE_COLOR_MAP`

While working on it, I noticed a black end of the gradient color bar, so I included the fix here, too. 

Screenshot 2024-09-05 at 4 50 11 PM


dzole0311 (approved):
Tested enabling/disabling the flag, works as expected. 

Thanks for fixing the black border as well! 🙏

PR #1148, opened 2024-09-11
Use sourceparams as it is when it is there

hanbyul-here:

### Description of Changes
While working on render extension talk, I realized that the authors don't necessarily use array format for rescale value (ex. rescale value for hls-ndvi layer in this file: https://raw.githubusercontent.com/NASA-IMPACT/veda-config/develop/datasets/hls-ndvi-ian.data.mdx is string `-1,1`) To support what we have now, I think we should return the source parameter as it is. 

I also simplified the logic a little bit, please check if I am missing anything!

### DIFF
--- a/app/scripts/components/exploration/data-utils-no-faux-module.ts
+++ b/app/scripts/components/exploration/data-utils-no-faux-module.ts
@@ -181,11 +181,11 @@ export function resolveRenderParams(
   }
 
   // Check for the dashboard render configuration in queryData
-  if (!queryDataRenders) throw new Error ('No proper render parameter exists.');
+  if (!queryDataRenders) throw new Error ('No render parameter exists from stac endpoint.');
 
   // Check the namespace from render extension
   const renderKey = queryDataRenders.dashboard? 'dashboard' : datasetSourceParams?.assets;
-  if (!queryDataRenders[renderKey]) throw new Error ('No proper render parameter exists.');
+  if (!queryDataRenders[renderKey]) throw new Error ('No proper render parameter for dashboard namespace exists.');
 
   // Return the render extension parameter
   if (queryDataRenders[renderKey] && hasValidSourceParams(queryDataRenders[renderKey])) {
--- a/app/scripts/components/exploration/data-utils-no-faux-module.ts
+++ b/app/scripts/components/exploration/data-utils-no-faux-module.ts
@@ -174,35 +174,22 @@ export function resolveRenderParams(
   datasetSourceParams: Record<string, any> | undefined,
   queryDataRenders: Record<string, any> | undefined
 ): Record<string, any> | undefined {
-  let renderParams: Record<string, any> | undefined;
   // Start with sourceParams from the dataset.
+  // Return the source param as it is if exists
   if (hasValidSourceParams(datasetSourceParams)) {
-    renderParams = datasetSourceParams;
+    return datasetSourceParams;
   }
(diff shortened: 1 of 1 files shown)


app/scripts/components/exploration/data-utils-no-faux-module.ts
```diff
@@ -174,35 +174,22 @@ export function resolveRenderParams(
...
-
-    if (renderParams?.rescale) {
-      renderParams.rescale = flattenAndCalculateMinMax(renderParams.rescale);
-    }
-  }
-
-  if (renderParams?.rescale) {
-    renderParams.rescale = flattenAndCalculateMinMax(renderParams.rescale);
+  const renderKey = queryDataRenders?.dashboard? 'dashboard' : datasetSourceParams?.assets;
+  if (!queryDataRenders?.[renderKey]) throw new Error ('No proper render parameter exists.');
```
hanbyul-here, line 185: should I return `undefined` instead of throwing an error here?

app/scripts/components/exploration/data-utils-no-faux-module.ts
```diff
@@ -174,35 +174,22 @@ export function resolveRenderParams(
...
+  if (queryDataRenders[renderKey] && hasValidSourceParams(queryDataRenders[renderKey])) {
+    const renderParams = queryDataRenders.dashboard;
```
dzole0311, line 187: Shouldn't this be `const renderParams = queryDataRenders[renderKey]`? Otherwise datasets like `Net Primary Production (NPP)` will error out.

dzole0311 (changes requested):
Thanks for simplifying this @hanbyul-here! 🙏 

> While working on render extension talk, I realized that the authors don't necessarily use array format for rescale value 

Just a note, the discussion we had here was the discreptancy between how Parcel outputs the rescale string. So the rescale from the mdx would be output as `[1, -1]` by Parcel, but titiler would return a nested array (`[[1, -1]]`), thus the flattening we're doing.

app/scripts/components/exploration/data-utils-no-faux-module.ts
hanbyul-here: You are right! thanks for catching it.

app/scripts/components/exploration/data-utils-no-faux-module.ts
hanbyul-here: Yeah I think missing render params need to throw an error than silently failing. thanks for your input!

app/scripts/components/exploration/data-utils-no-faux-module.ts
```diff
@@ -174,35 +174,27 @@ export function resolveRenderParams(
...
   if (hasValidSourceParams(datasetSourceParams)) {
-    renderParams = datasetSourceParams;
+    return datasetSourceParams;
   }
 
-  // Check for the dashboard render configuration in queryData if not defined.
-  if (!renderParams && queryDataRenders?.dashboard) {
-    renderParams = queryDataRenders.dashboard;
+  // Check for the dashboard render configuration in queryData
+  if (!queryDataRenders) throw new Error ('No proper render parameter exists.');
```
sandrahoang686, line 184: maybe different err messages for each spot to differentiate and find source of err right away - this line could be `No render data exists` ?

dzole0311 (approved):

PR #1149, opened 2024-09-11
Update PageHeader/Nav to not throw

sandrahoang686:
Was playing around a bit with nav header with next instance and noticed something I did not catch in the initial PR review. But we dont want to throw an error when type is incorrect. This will **break** the app because nothing is catching it. Instead we should still render the rest of the nav items that are correct.

I opted in for logging the error for incorrect types but if we prefer a different way like catching and bubbling up the error, lmk.

Updated behavior looks like:
Screenshot 2024-09-11 at 12 08 28 PM

### DIFF
--- a/app/scripts/components/common/page-header/nav-menu-item.tsx
+++ b/app/scripts/components/common/page-header/nav-menu-item.tsx
@@ -58,6 +58,7 @@ const DropMenuNavItem = styled(DropMenuItem)`
   `}
 `;
 
+const LOG = true;
 
 function LinkDropMenuNavItem({ child, onClick }: { child: NavLinkItem, onClick?:() => void}) {
   const { title, type, ...rest } = child;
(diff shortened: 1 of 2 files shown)


sandrahoang686:
This brings up a broader discussion of how do want to handle errors on UI side? We should have a standard of logging errors and then when we throw, we should make sure the app doesn't break. Instead these throws should bubble up and be caught at some point and component returns null instead of breaking... Food for thought.

dzole0311 (commented):
@sandrahoang686 Could you provide some steps for what to look for when testing the PR please?

sandrahoang686:
@dzole0311 with the preview, you'd just wanna make sure it works as expected. To test fail case, you can pull down code locally and update the navConfig to have an unexpected type. This would be a good test case scenerio 🤔

dzole0311 (approved):
Looks good to me, thanks for spotting this 👍

PR #1150, opened 2024-09-12
Allow `<Map>` component comparison to support configurable datasets/layers

ram-sharan25:
Previously, comparing datasets involved manually assigning the dataset to the properties of a dataset layer. With this change, you can now pass two additional parameters  `compareDatasetId` and `compareLayerId` to the `Map` block, allowing it to compare the specified dataset with the one defined by `datasetId` and `layerId`.

### DIFF
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -112,7 +112,7 @@ interface MapBlockProps {
   datasets: VedaDatum<DatasetData>;
   dateTime?: string;
   compareDateTime?: string;
-  compareDataSetId?: string;
+  compareDatasetId?: string;
   compareLayerId?: string;
   center?: [number, number];
   zoom?: number;
@@ -148,7 +148,7 @@ function MapBlock(props: MapBlockProps) {
     layerId,
     dateTime,
     compareDateTime,
-    compareDataSetId,
+    compareDatasetId,
     compareLayerId,
     compareLabel,
     center,
@@ -171,9 +171,9 @@ function MapBlock(props: MapBlockProps) {
     const [baseMapStaticData] = reconcileDatasets([layerId], datasetLayers, []);
     let totalLayers = [baseMapStaticData];
     const baseMapStaticCompareData =
-      !!compareDataSetId && !!compareLayerId
+      !!(compareDatasetId && compareLayerId)
         ? {
-            datasetId: compareDataSetId,
+            datasetId: compareDatasetId,
             layerId: compareLayerId
           }
         : baseMapStaticData.data.compare;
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -112,6 +112,8 @@ interface MapBlockProps {
   datasets: VedaDatum<DatasetData>;
   dateTime?: string;
   compareDateTime?: string;
+  compareDataSetId?: string;
+  compareLayerId?: string;
```
sandrahoang686, line 116: nit: `compareDatasetId`, no need to capitalize the `s` but do we need to pass that in? It wont be used it looks like. datasetId is never referenced from `baseMapStaticCompareData`

app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -166,7 +170,13 @@ function MapBlock(props: MapBlockProps) {
   const layersToFetch = useMemo(() => {
     const [baseMapStaticData] = reconcileDatasets([layerId], datasetLayers, []);
     let totalLayers = [baseMapStaticData];
-    const baseMapStaticCompareData = baseMapStaticData.data.compare;
+    const baseMapStaticCompareData =
+      !!compareDataSetId && !!compareLayerId
```
sandrahoang686, line 174: nit: `!!(compareDatasetId && compareLayerId)`

sandrahoang686 (changes requested):

app/scripts/components/common/blocks/block-map.tsx
ram-sharan25: It is optional. We have to pass tha in ,if only we want to compare the datasets.

hanbyul-here:
👋 This change needs a bit more thinking through regarding labeling. ex. We are depending on `compare` attribute of the base dataset to get the formatted label dynamically. If we are passing datasetId and layerId directly, we can't really use base dataset's label function. 
What is the expected label when compare datasetid and compare layer id are passed?

slesaad:
> 👋 This change needs a bit more thinking through regarding labeling. ex. We are depending on `compare` attribute of the base dataset to get the formatted label dynamically. If we are passing datasetId and layerId directly, we can't really use base dataset's compare label function. What is the expected label when compare datasetid and compare layer id are passed?

I'd say a combination of dataset name, layer name, and date? @hanbyul-here

aboydnw:
@slesaad is this needed for GHG in the near term, or is it more about pushing a useful feature that you all used on GHG back into veda-ui? 

If it's more about making a useful feature available in veda-ui, I'd advocate for making a ticket that can more holistically address the map comparison feature in stories. we haven't really touched on that in a while so it's probably due for an update soon. if it's more something that you need for GHG, maybe we push this forward as a quick n dirty solution, and then address more holistically later.

slesaad:
@aboydnw for sure, please create a ticket. there's no rush to get it in GHG right now, but it's something we will need soon enough. didn't have all the details when this PR was created :D

Issue #1134, opened 2024-08-28
[Bug] compareLabel on slider maps not populating


acblackford:
### Context

Any slider maps that specifically call compareLabel in the mdx do not display any text, but the maps that do not set this and let it handle on its own do display the date comparisons correctly. An example is in the urban heat data story, where the first slider does not call anything and shows the 2001 vs 2019 call but the next slider sets a compareLabel and does not show any text on the label.

URL:  https://www.earthdata.nasa.gov/dashboard/stories/urban-heating

Blocking a release? (y/n): n

Regression issue? (y/n): n
[If yes, add Regression label]

Priority: low (P3)

### Steps to reproduce the issue

1. [Step 1 to reproduce]
2. [Step 2 to reproduce]

PR #1153, opened 2024-09-13
Fix compare label on block map

hanbyul-here:
**Related Ticket:** Close https://github.com/NASA-IMPACT/veda-ui/issues/1134

### Description of Changes
This change needs to be included for R3. This branch needs to be used to cut the patch (So this doesn't need to be merged to main - but I will cut a release on this branch when approved.)  This branch is off from v5.7.0 not to include the latest changes on main

### Validation / Testing
Check the two compare maps on mdx page to see the fix https://deploy-preview-1153--veda-ui.netlify.app/sandbox/mdx-page

### DIFF
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -1,6 +1,7 @@
 import React, { useMemo, useState, useEffect } from 'react';
 import styled from 'styled-components';
 import { MapboxOptions } from 'mapbox-gl';
+import * as dateFns from 'date-fns';
 import {
   convertProjectionToMapbox,
   projectionDefault,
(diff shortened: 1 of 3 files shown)


slesaad (approved):
This looks good!

Although I noticed that in the image comparison, the labels get cut off by the slider - see this page - https://deploy-preview-1153--veda-ui.netlify.app/data-catalog/no2

Doesn't need to be fixed now, but just maybe needs to be tracked for later.

dzole0311:
> Although I noticed that in the image comparison, the labels get cut off by the slider

Thanks for pointing that @slesaad! This seems to be the implemented behavior of the `react-compare-image` npm package we use for the image comparison (similar example here: https://react-compare-image.yuuniworks.com/?path=/story/basic--show-labels). It looks like this behavior is common across other comparison libraries too like `react-compare-slider` (example: https://react-compare-slider.vercel.app/?path=/story/recipes--item-labels). 

We can keep track on it for later, though I'm unsure if there's a simple fix given how those libs work 🤔

dzole0311 (approved):

Issue #1144, opened 2024-09-03
Set up Build pipeline for USWDS font size


hanbyul-here:
USWDS builds out different font sizes for typefaces. Link: https://designsystem.digital.gov/utilities/font-size-and-family/ We decided that we don't need this feature. Investigate how we can consolidate the sizes for each type face.

dzole0311:
With @faustoperez and @hanbyul-here we agreed to skip removing the font-normalization for now, until we have a clearer image of the template instance. So I'm closing the related PR, however, we have a reference to it and the necessary working configuration if we decide to implement the feature.

PR #1157, opened 2024-09-17
Cap heights for the theme typeface in USWDS

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1144

### Description of Changes
USWDS normalizes font sizes across different typefaces to ensure optical consistency, but this feature isn't necessary for our current needs. Here we're trying to streamline the font sizes across all typefaces to ensure they have uniform sizing without normalization.

I found the [function where USWDS normalizes the typefaces](https://github.com/uswds/uswds/blob/e6a8bb2670e00cbd57177eb8c19d34313b7a53bb/packages/uswds-core/src/styles/functions/font/normalize-type-scale.scss#L15), which is based on the cap-height of each typeface. Since each typeface has a slightly different [cap-height setting](https://github.com/uswds/uswds/blob/e6a8bb2670e00cbd57177eb8c19d34313b7a53bb/packages/uswds-core/src/styles/tokens/font/typefaces.scss), I bypassed this by configuring the custom tokens that we need to use a fixed default cap-height of 364px.

To verify the changes, I tested with the following examples:

```
<h1 className='font-serif-3xs'>Lorem ipsum dolor sit amet, consectetur adipiscing elit </h1>
<h2 className='font-sans-3xs'>Lorem ipsum dolor sit amet, consectetur adipiscing elit </h2>
```

These previously would have been normalized differently, but now output the same `rem` value.

### Validation / Testing

The example from the description can be used for testing

### DIFF
--- a/app/scripts/styles/_uswds-theme.scss
+++ b/app/scripts/styles/_uswds-theme.scss
@@ -8,6 +8,36 @@
             "display-name": var(--base-font-family),
             "cap-height": 364px,
         ),
+        "system": (
+            "cap-height": 364px
+        ),
+        "georgia": (
+            "cap-height": 364px
+        ),
+        "helvetica": (
+            "cap-height": 364px
+        ),
+        "tahoma": (
+            "cap-height": 364px
+        ),
+        "verdana": (
+            "cap-height": 364px
+        ),
+        "open-sans": (
+            "cap-height": 364px
+        ),
+        "merriweather": (
+            "cap-height": 364px
+        ),
+        "roboto-mono": (
+            "cap-height": 364px
+        ),
+        "source-sans-pro": (
+            "cap-height": 364px
+        ),
+        "public-sans": (
+            "cap-height": 364px
+        )
     ),
     $theme-type-scale-md: 8,
     $theme-utility-breakpoints: (
--- a/app/scripts/styles/_uswds-theme.scss
+++ b/app/scripts/styles/_uswds-theme.scss
@@ -11,33 +11,6 @@
         "system": (
             "cap-height": 364px
         ),
-        "georgia": (
-            "cap-height": 364px
-        ),
-        "helvetica": (
-            "cap-height": 364px
-        ),
-        "tahoma": (
-            "cap-height": 364px
-        ),
-        "verdana": (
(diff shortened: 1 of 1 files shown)


app/scripts/styles/_uswds-theme.scss
```diff
@@ -8,12 +8,74 @@
             "display-name": var(--base-font-family),
             "cap-height": 364px,
         ),
+        "system": (
+            "cap-height": 364px
+        ),
+        "merriweather": (
+            display-name: "Merriweather Web",
+            cap-height: 364px,
```
dzole0311, line 16: I couldn't find a way to only set the cap-height without adding the rest of the values for the typeface, copied from the uswds config defaults.

app/scripts/styles/_uswds-theme.scss
```diff
@@ -8,12 +8,74 @@
...
+                ),
+            ),
+        ),
     ),
     $theme-type-scale-md: 8,
     $theme-utility-breakpoints: (
         "mobile-lg": false,
         "desktop": false
     ),
     $theme-font-type-sans: baseFontFamily,
```
dzole0311, line 80: @faustoperez I mapped sans to the base font family (e.g "Inter" in GHG Center's case). Could be adjusted as we need it.

dzole0311 (commented):
There should be no risk if we decide to merge these updated configs, as I think no visual changes will appear.

hanbyul-here:
Thanks a lot for working on it 🙇 It is quite amount of setups just to disable the normalization of the font - I end up having this question if we need to get rid of this normalization at all (I know, I am sorry 😓 )  I think we started this work because we don't need this normalization - but it might be ok to have the normalization of the font?  @faustoperez  - What do you think?

faustoperez:
@hanbyul-here sounds good! I think we should pause the USWDS detailed decisions work (ie should we normalize the font sizes) until we have more clarity on the vision/template instance design (quite likely at the end of this week)

dzole0311:
@faustoperez @hanbyul-here thanks for the feedback. I'll close this PR and the related issue now, since now we have a working implementation and reference if we decide to remove the normalization again.

PR #1159, opened 2024-09-17
Datacatalog followup

hanbyul-here:
**Related Ticket:**  Follow up of https://github.com/NASA-IMPACT/veda-ui/pull/1096 & https://github.com/NASA-IMPACT/veda-ui/commit/d84527bdf0eb2b4d7586646881796bb615b433b3#diff-7d1c949cac222e52ef45864275aff43e6fa94599bd4684d9611fa25b610e900b

### Description of Changes
- I accidentally removed the question mark operator while running a lint, I put the proper type so it doesn't happen with automatic lint
- I made the Card Component work backward compatible since the component is exposed to instance levels.

PR #1160, opened 2024-09-18
Lazyload SmartLink to avoid loading react-router-dom + use a proper linkProperties 

hanbyul-here:
**Related Ticket:** follow up of follow up -_- https://github.com/NASA-IMPACT/veda-ui/pull/1159

### Description of Changes
Urg I used wrong attributes, so the previous fix didn't really fix the problem. 
The previous approach broke the library build too by including SmartLink, which uses react-router-dom. I think I worked around by lazy loading the component in this PR. I tested it locally but if you can, please give it another test 🙇 

### Validation / Testing
ok, this time, I actually created a preview on GHG; this is the page that  uses Card component: https://deploy-preview-596--ghg-demo.netlify.app/learn

### DIFF
--- a/app/scripts/components/common/card/index.tsx
+++ b/app/scripts/components/common/card/index.tsx
@@ -1,9 +1,9 @@
-import React, { MouseEventHandler } from 'react';
+import React, { lazy, MouseEventHandler } from 'react';
 import styled, { css } from 'styled-components';
 import { Link } from 'react-router-dom';
 import { format } from 'date-fns';
 import { CollecticonExpandTopRight } from '@devseed-ui/collecticons';
-
+const SmartLink = lazy(() => import('../smart-link'));
 import {
   glsp,
   media,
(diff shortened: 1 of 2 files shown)


dzole0311 (approved):

app/scripts/components/common/card/index.tsx
```diff
@@ -287,10 +288,11 @@ function CardComponent(props: CardComponentPropsType) {
   } else {
     const { linkTo, onLinkClick,  } = props;
     linkProperties = {
-      to: linkTo,
+      linkTo,
       onLinkClick,
       pathAttributeKeyName: 'to',
-      linkComponent: SmartLink
+      // @ts-expect-error SmartLink needs to be lazily loaded temporarily to prevent ui-library from loading react-router-dom
+      LinkElement: SmartLink
```
sandrahoang686, line 295: Relooked at your [previous PR](https://github.com/NASA-IMPACT/veda-ui/pull/1159/files#diff-f936775422e987f9214c5bbdc155e139c0404494103a14f1da0792e2825beef3) - so looks like we have to default to SmartLink here because GHG is unable to pass this element into the card? We may just want to update this on GHG side to pass the LinkElement in instead of this non-ideal temporary solution? Unsure if thats too much of a lift right now though so this comment is non-blocking. But could we leave a `@TODO` or `@NOTE` too here with the same note you have?

sandrahoang686 (approved):

app/scripts/components/common/card/index.tsx
hanbyul-here: Yeah, it will be best to just change the code from the instance side. But I think it is easy to miss it with R3 and everything, so put the temporary measure here. I left a note.

PR #1163, opened 2024-09-20
Parse cookie consent form as html

hanbyul-here:
**Related Ticket:** [_{link related ticket here}_](https://github.com/NASA-IMPACT/veda-ui/pull/1127)

### Description of Changes
We already have a mechanism to parse markdown to html bts. It is highly likely that users will try to use different markdown syntax if they see the link syntax. So here, I made parcel resolver to compile the cookieconsent form to html beforehand. I only did it for copy, but some questions are 
- does title also accept markdown?
- should we have some kinds of prefixes to note that this is a markdown? (we use ::markdown prefix for mdx files, but this is configuration so it can be a bit different story.

### DIFF
--- a/parcel-resolver-veda/index.js
+++ b/parcel-resolver-veda/index.js
@@ -6,6 +6,7 @@ const dedent = require('dedent');
 const _ = require('lodash');
 const { Resolver } = require('@parcel/plugin');
 
+const markdownit = require('markdown-it');
 const stringifyYmlWithFns = require('./stringify-yml-func');
 const { loadVedaConfig } = require('./config');
 const { getFrontmatterData } = require('./frontmatter');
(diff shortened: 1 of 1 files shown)


snmln:
Thank you Hanbyul! This is far cleaner than what I previously had. To answer your questions: 

No I do not think we need to enable markdown in the title. To my knowledge there is no situation that we have discussed where we would need to accommodate that need. Additionally from a UX best practices, linking the title in this type of component would be a highly abnormal interaction pattern. I would recommend against it. 

Adding the prefix is a good idea. Since that is an input type that the user base is familiar with we should denote that in some fashion to avoid any confusion.

snmln (approved):

hanbyul-here:
I will merge this change now since some of the  UI changes would need to be made because of this change. let's keep the conversation about the prefix in the og pr.

snmln:
> I will merge this change now since some of the UI changes would need to be made because of this change. let's keep the conversation about the prefix in the og pr.

Sounds good!

PR #1169, opened 2024-09-24
Add docs for registry dev

hanbyul-here:
### Description of Changes
Adding a doc for VEDA-UI as a registry development. 

I am pretty sure there are lots of improvements that can be made on this doc. But I want to at least offer every essential thing that a dev needs to know re: veda-ui as a registry. Am I missing anything?

### DIFF
--- a/docs/development/REGISTRY.md
+++ b/docs/development/REGISTRY.md
@@ -1,0 +1,55 @@
+## Library build
+
+As VEDA-UI is trying to become a registry, there are several things that developers need to be aware of.
+
+### Local Development
+
+Build locally first with the command below.
+```
+yarn buildlib
+```
+
+After successfully building the library, link the project to your local environment with the following command, executed inside the VEDA-UI directory:
+
+```
+yarn link 
+```
+Then change your directory to the project that is using veda-ui, which has `@developmentseed/veda-ui` as a dependency. Create a local alias for VEDA-UI by running:
+```
+yarn link @developmentseed/veda-ui
+```
+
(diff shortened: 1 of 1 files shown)


docs/development/REGISTRY.md
```diff
@@ -0,0 +1,62 @@
...
+To check the latest version in the registry and ensure proper versioning, run:
+
+```
+npm view @developmentseed/veda-ui version
+```
+
+Adjust the version number as needed and use an appropriate pre-release tag if you're publishing incremental updates. You can version your package with pre-release tag with the command like below. 
+
+```
+npm version 5.7.0-alpha.3
```
sandrahoang686, line 58: When I did this, this also committed the version to git and i'm unsure if it creates a tag in git as well which is that something we want? Right now, we just need to publish for development purposes for the nextJs preview so I didn't think we needed to officially version and tag which is why I haven't been just changing it in the package.json directly to publish to keep versioning entirely separate but what do you think?

hanbyul-here: Ah yes! This is the documentation to document the current work process, so I will change it. But I'd like to suggest to start versioning with npm version (make commits and push it.) because there is no way to communicate what version was published or not through the code base currently.  I think this practice will be helpful as more developers touch it. And the version commit will help to see which features were published or not. But I also see the cons of it ex. People who need to update UI might be confused about pre-release tag. so I can go either way

sandrahoang686 (approved):

Issue #1155, opened 2024-09-17
[Refactor] VEDA Core Features - (E&A) Make Map Controls work


sandrahoang686:
This is a tech-debt ticket as a result of https://github.com/NASA-IMPACT/veda-ui/issues/902. 

**Description:**
Currently `AnalysisMessageControl` and `CustomAoIControl` are commented out from `/exploration/components/map` because of breaking errors in the NextJs instance. We need to get these controls working as usual in current build and working for nextJs instance. 

`AnalysisMessageControl` Error is related to `useThemedControl` which needs to be investigated more
`CustomAoIControl` Errors with...
```TypeError: Cannot read properties of undefined (reading 'getSelectedIds')
Source
app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx (339:28) @ getSelectedIds
337 |    } else {
 338 |      mbDraw.changeMode(SIMPLE_SELECT, {
> 339 |        featureIds: mbDraw.getSelectedIds()
     |                            ^
 340 |      });
 341 |    }
 342 |  }, [main, isDrawing]);
```

**Acceptance Criteria**:
Get AnalysisMessageControl and CustomAoIControl working

**Feature PR/Branch:** 
veda-ui: https://github.com/NASA-IMPACT/veda-ui/pull/1154
next-veda-ui: https://github.com/developmentseed/next-veda-ui/pull/4

PR #1172, opened 2024-09-25
Clean up the themed controls when unmounted

hanbyul-here:
### Description of Changes
Clean up ref when themed controls are unmounted

I kept seeing an error that 'cannot update unmounted root' from dev environment of Next instance. I think it is because our themed control doesn't clean up its ref when unmounted. (It happens when I revisit the page while navigating - does not happen on the first landing.)

Screenshot 2024-09-25 at 10 00 11 AM

But I do have some questions 🤔 
- It doesn't happen on production env: https://veda-ui-next-test.netlify.app/stories/lahina-fire
- It happens only on the first map element of the page. Why would that be?

### DIFF
--- a/app/scripts/components/common/map/controls/hooks/use-themed-control.tsx
+++ b/app/scripts/components/common/map/controls/hooks/use-themed-control.tsx
@@ -49,6 +49,10 @@ export default function useThemedControl(
         <ThemeProvider theme={theme}>{renderFn() as any}</ThemeProvider>
       );
     }
+    return () => {
+      rootRef.current?.unmount();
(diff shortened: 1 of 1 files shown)


hanbyul-here:
Oh interesting, this change breaks custom controls on E&A pages, with an error that ` Attempted to synchronously unmount a root while React was already rendering. React cannot finish unmounting the root until the current render has completed, which may lead to a race condition.` @dzole0311  just in case this is related to the ticket you assigned yourself (https://github.com/NASA-IMPACT/veda-ui/issues/1155

PR #1173, opened 2024-09-25
Integrate Geo-Copilot to VEDA UI Exploration Page

xhagrg:
### Description of Changes
This PR introduces components in the exploration page required for interaction with Geo-CoPilot backend being developed by our Microsoft Counterparts. Most of the interactions done in exploration page can now be done with the help of geo-copilot, mainly loading datasets, comparing dates for datasets, and running analysis. 

This PR mostly tackles the basic functionalities of the co-pilot. There are more interactions planned, and will be updated in the next PRs to come. 

### Validation / Testing
- The new components touch all of the existing features. However, it should not affect any of those functionality
- Interaction with Geo-Copilot should not result in un-restricted loading bar. (any interaction should either lead to error message or message from the co-pilot)
- Manual interaction should not affect the interaction with geo-copilot.


app/scripts/components/exploration/index.tsx
```diff
@@ -89,33 +111,93 @@ function ExplorationAndAnalysis(props: ExplorationAndAnalysisProps) {
     };
   }, [resetAnalysisController, setUrl]);
 
+  const mapPanelRef = useRef<any>(null);
```
hanbyul-here, line 114: Have you tried using utility methods that `react-resizable-panel expose` and using one ref? https://www.npmjs.com/package/react-resizable-panels#can-a-attach-a-ref-to-the-dom-elements

hanbyul-here (commented):
Thanks for working on it. It is such a fun feature and I like seeing the methods we built are being used by geo-copilots to dynamically add datasets and run analysis 🤩

I could test and validate the functionalities you described in the pr.  The main problem I can spot now is AOIs are behaving unexpectedly. ex. already-drawn AOIs are still there even after geo-copilots finished the task, and pre-defined AOI selector (the dropdown on the left-top with US States as options) doesn't get reset after AOI is gone. Of course this change was not introduced by your change, it is more like it highlights the problem we already had.  - We are currently working on the improvements now (https://github.com/NASA-IMPACT/veda-ui/issues/1207). I am not sure if that will fix all the issues, but I hope it helps future iterations.

app/scripts/components/exploration/index.tsx
xhagrg: I hadn't seen this. Thank you, @hanbyul-here . that definitely helps make the code a little cleaner.

xhagrg: @hanbyul-here I checked on this, and `getPanelElement` is not available in the current version of `react-resizable-panels`. will keep the changes for now, and will upgrade later.

hanbyul-here:
Closing it as it is stale. Let's discuss if we need to work further on this.

Issue #1182, opened 2024-10-01
Restrict dynamic colormaps to layer types that support it


j08lue:
Dynamic colormaps will only work for certain endpoints / services (e.g. TiTiler-based).

I guess we currently distinguish between three `layer` `type`s: `raster|vector|zarr`.

https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md

Will we need some more fine-grained distinction between server capabilities?

https://github.com/NASA-IMPACT/veda-config/blob/1250da71d5c09227938bcd614875ae85e54a1c58/datasets/cmip6-tas.data.mdx?plain=1#L30

_Originally posted by @j08lue in https://github.com/NASA-IMPACT/veda-ui/issues/1102#issuecomment-2291617986_

PR #1183, opened 2024-10-02
Restrict dynamic colormaps to applicable layer types

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1182

### Description of Changes
1. Added util function `isRenderParamsApplicable` to handle the exclusion of render parameters for vector datasets (for now, add more if applicable in the future)
2. Updated logic in `reconcileQueryDataWithDataset` to prevent errors when processing vector datasets that don't support colormaps or rescaling

Vector datasets should be excluded from applying dynamic colormaps.

### Validation / Testing

1. Verify that vector datasets no longer trigger errors in the exploration view
2. Test colormap functionality on raster datasets and make sure there are no regressions

### DIFF
--- a/app/scripts/components/exploration/data-utils-no-faux-module.ts
+++ b/app/scripts/components/exploration/data-utils-no-faux-module.ts
@@ -140,6 +140,18 @@ const hasValidSourceParams = (params) => {
 };
 
 /**
+ * Utility to check if render parameters are applicable based on dataset type.
+ *
+ * @param datasetType The type of the dataset (e.g., 'vector').
+ * @returns Boolean indicating if render parameters are applicable.
+ */
+export const isRenderParamsApplicable = (datasetType: string): boolean => {
+  const nonApplicableTypes = ['vector'];
+
+  return !nonApplicableTypes.includes(datasetType);
+};
+
+/**
  * Util to flatten and process rescale values,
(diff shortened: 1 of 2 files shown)


app/scripts/components/exploration/data-utils-no-faux-module.ts
```diff
@@ -139,6 +139,18 @@ const hasValidSourceParams = (params) => {
   return params && 'colormap_name' in params && 'rescale' in params;
 };
 
+/**
+ * Utility to check if render parameters are applicable based on dataset type.
+ *
+ * @param datasetType The type of the dataset (e.g., 'vector').
+ * @returns Boolean indicating if render parameters are applicable.
+ */
+export const isRenderParamsApplicable = (datasetType: string): boolean => {
```
sandrahoang686, line 148: can we possibly type this as `DatasetLayerType`

sandrahoang686 (approved):
thanks for looking into it, one comment, but not enough to block

app/scripts/components/exploration/data-utils-no-faux-module.ts
dzole0311: Thanks for spotting, updated!

PR #1184, opened 2024-10-07
get page header working with linkProperties

sandrahoang686:

### Description of Changes

### Validation / Testing

### DIFF
--- a/app/scripts/components/common/page-header/logo.tsx
+++ b/app/scripts/components/common/page-header/logo.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
 import { glsp, media, themeVal } from '@devseed-ui/theme-provider';
-import { Link } from 'react-router-dom';
 import NasaLogo from '../nasa-logo';
 import { Tip } from '../tip';
+import { LinkProperties } from '../card';
 import { ComponentOverride } from '$components/common/page-overrides';
 
 const appTitle = process.env.APP_TITLE;
@@ -74,7 +74,7 @@ const Brand = styled.div`
   }
 `;
 
-const PageTitleSecLink = styled(Link)`
+const PageTitleSecLink = styled.a`
   align-self: end;
   font-size: 0.75rem;
   font-weight: ${themeVal('type.base.bold')};
@@ -98,16 +98,21 @@ const PageTitleSecLink = styled(Link)`
   `}
 `;
 
-export default function Logo () {
+export default function Logo ({ linkProperties }: { linkProperties: LinkProperties }) {
+  const LinkElement = linkProperties.LinkElement; // @TODO-SANDRA: Revisit typing here...
+  const linkProps = {
+    [linkProperties.pathAttributeKeyName]: '/'
+  };
+
   return (
   <ComponentOverride with='headerBrand'>
     <Brand>
-      <Link to='/'>
+      <LinkElement {...linkProps}>
         <NasaLogo />
         <span>Earthdata</span> <span>{appTitle}</span>
-      </Link>
+      </LinkElement>
       <Tip content={`v${appVersion}`}>
-        <PageTitleSecLink to='/development'>Beta</PageTitleSecLink>
+        <PageTitleSecLink as={LinkElement} href='/development'>Beta</PageTitleSecLink>
       </Tip>
     </Brand>
   </ComponentOverride>);
--- a/app/scripts/components/common/layout-root/index.tsx
+++ b/app/scripts/components/common/layout-root/index.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode, useContext, useCallback, useEffect } from 'react';
+import { Link } from 'react-router-dom';
 import { useDeepCompareEffect } from 'use-deep-compare';
 import styled from 'styled-components';
 import { Outlet } from 'react-router';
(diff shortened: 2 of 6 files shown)


app/scripts/components/common/page-header/logo.tsx
```diff
@@ -74,7 +74,7 @@ const Brand = styled.div`
   }
 `;
 
-const PageTitleSecLink = styled(Link)`
+const PageTitleSecLink = styled.a`
```
hanbyul-here, line 77: This will make the links on the nav plain a tags, which we don't want. How about we take a similar approach that `ElementInteractive` component is taking,(https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/card/index.tsx#L307)  passing the linkElement as `as` so the styled component can be whatever props says.

app/scripts/components/common/page-header/logo.tsx
```diff
@@ -98,16 +98,18 @@ const PageTitleSecLink = styled(Link)`
   `}
 `;
 
-export default function Logo () {
+export default function Logo ({ linkProperties }: { linkProperties: LinkProperties }) {
+  const LinkElement: ComponentType<any> = linkProperties.LinkElement as ComponentType<any>;
+
   return (
   <ComponentOverride with='headerBrand'>
```
hanbyul-here, line 105: I am having trouble understanding how exporting Logo can work, when `ComponentOverride` component imports elements from the virtual module`veda` directly (https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/page-overrides.tsx#L3) Can you help me understand how this works?

hanbyul-here: All of the instsances except OG dashboard overrides this component with each instance's branding : ex. https://github.com/US-GHG-Center/veda-config-ghg/blob/16fe5b7c08e67f78e1a2a27727a82ff28034f4bc/overrides/components/header-brand/component.tsx#L4 🤔 which makes me wonder if this component is worth exporting at all. how do you think?

hanbyul-here (changes requested):
Thanks for working on it! I think navlink can be more worked on so they don't have to be a tags !
https://github.com/NASA-IMPACT/veda-ui/pull/1184/files#diff-6cccba12d155a579cba98d2e85e43fd88de10b0b6e83f44a261cceb1429e2695R77

app/scripts/components/common/page-header/logo.tsx
dzole0311: Probably not worth exporting, and we could anyway pass any ReactElement as a logo prop from Next.js: https://github.com/NASA-IMPACT/veda-ui/pull/1184/files#diff-86e372214b3b1dff36fcabf602d3ea911690d226a6e0bed48fee30b571864756R233

app/scripts/components/common/page-header/logo.tsx
sandrahoang686: oh, yep thats what I did actually => https://github.com/NASA-IMPACT/veda-ui/blob/06b3a016226b2f22a95fa9ab27d8a673b7525a65/app/scripts/components/common/page-header/logo.tsx#L112. I had to switch it to an `a` tag because we cant import `Link` from `react-router-dom` at this level so its moved up!

app/scripts/components/common/page-header/logo.tsx
sandrahoang686: very good point, thanks for making it 👍🏼 It might make sense in this case to separate the actual logo/svg from the styled container and have the logo/svg passed as a prop. I'll have to remove the veda virtual module dependency though so might have to recreate the same without the override 🤔, let me give it more thought!

app/scripts/components/common/page-header/logo.tsx
hanbyul-here: Oh I see, sorry for missing that part!

app/scripts/components/common/page-header/logo.tsx
sandrahoang686: Created a separate LogoContainer component that takes in the SVG logo as a prop along with other attributres like title. When working on the new page header design implementation, we might not have to expose this container though as its meant to just fit in the page header itself, might just have instance pass in only the SVG itself but this will works best for now for our current state of code.

app/scripts/components/common/page-header/logo-container.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const LinkElement: ComponentType<any> = linkProperties.LinkElement as ComponentType<any>;
+
+  return (
+    <Brand>
+      <LinkElement {...{[linkProperties.pathAttributeKeyName]: '/'}}>
+        {Logo}
+        <span>{title}</span> <span>{subTitle}</span>
+      </LinkElement>
+      <Tip content={`v${version}`}>
+      <PageTitleSecLink {...{as: linkProperties.LinkElement as ComponentType<any>, [linkProperties.pathAttributeKeyName]: '/development'}}>Beta</PageTitleSecLink>
```
hanbyul-here, line 26: I am  not sure if this component can be used by other instances when only logo svg is tweak-able - (ex. ghg overwrote the whole logo)

hanbyul-here (approved):
I left my concern about logo-container component that it might not be catered towards users' needs for logo. Then users don't need to use it - so it probably is not a blocker. Thanks for working on it!

PR #1190, opened 2024-10-10
Color maps range slider implementation

snmln:
**Related Ticket:** https://github.com/orgs/NASA-IMPACT/projects/17/views/6?pane=issue&itemId=74631613&issue=NASA-IMPACT%7Cveda-ui%7C1102

### Description of Changes
Creation and implementation of new range slider component. 

-There is a identified bug that while dragging the slider thumbs you can drag the entire panel up and down influencing the parent component position. 
-Additional functionality: If input field is left empty and clicked outside of the default _min_ or _max_ should auto populate the field. 

### Validation / Testing
Navigate to a data exploration and add a layer to the timeline. 
- **Layers:**
    - Confirm that **min** and **max** are accurate to value in the layer legend view. 
    - Switch to another layer with different value confirm that the min and max in the range slider are accurate. 
    - Select a layer with a min value less than 0 (net ecosystem exchange) you should see a middle marker that represents the center of the current gradient values. 
- **Slider:**
    - You should not be able to drag the min slider past the max slider, or the max slider past the min slider. 
- **Text Inputs:**
    - You should be able to type in a value greater or lower than the min max of the layer but then be shown with the error message _Please enter a value between {min} and {max}_ no change to the slider range or the map. 
    - After selecting a different max that is less than the layers absolute maximum with either slider or input, if you enter in a value in the min text input that is greater than the newly selected max you should be presented with an error message _Please enter a value less than {max}_ 
    - After selecting a different min that is larger than the layers absolute minimum with either slider or input, if you enter in a value in the max text input that is greater than the newly selected max you should be presented with an error message _Please enter a value larger than {min}_

### DIFF
--- a/app/scripts/components/exploration/components/datasets/color-range-slider.scss
+++ b/app/scripts/components/exploration/components/datasets/color-range-slider.scss
@@ -1,12 +1,3 @@
-
-.slider__track {
-  background-color: #dddfe2;
-}
-
-.slider__range {
-  background-color: #1565ef;
-}
-
 /* Removing the default appearance */
 .thumb,
 .thumb::-webkit-slider-thumb {
@@ -18,9 +9,6 @@
 
 .thumb {
   pointer-events: none;
-}
-.middle-marker {
-  background-color: #1b2631;
 }
 /* For Chrome browsers */
 .thumb::-webkit-slider-thumb {
--- a/app/scripts/components/exploration/components/datasets/color-range-slider.scss
+++ b/app/scripts/components/exploration/components/datasets/color-range-slider.scss
@@ -1,3 +1,4 @@
+
 .slider__track {
   background-color: #dddfe2;
 }
@@ -9,6 +10,8 @@
 /* Removing the default appearance */
 .thumb,
 .thumb::-webkit-slider-thumb {
+  touch-action: 'none';
+
   -webkit-appearance: none;
   -webkit-tap-highlight-color: transparent;
 }
@@ -16,8 +19,8 @@
 .thumb {
   pointer-events: none;
 }
-.middle_marker{
-  background-color: #1B2631;
+.middle-marker {
+  background-color: #1b2631;
 }
 /* For Chrome browsers */
 .thumb::-webkit-slider-thumb {
--- a/app/scripts/components/exploration/components/datasets/color-range-slider.scss
+++ b/app/scripts/components/exploration/components/datasets/color-range-slider.scss
@@ -1,0 +1,59 @@
+.slider__track {
+  background-color: #dddfe2;
+}
+
+.slider__range {
+  background-color: #1565ef;
+}
+
+/* Removing the default appearance */
+.thumb,
+.thumb::-webkit-slider-thumb {
+  -webkit-appearance: none;
+  -webkit-tap-highlight-color: transparent;
+}
+
+.thumb {
+  pointer-events: none;
+}
+.middle_marker{
+  background-color: #1B2631;
+}
+/* For Chrome browsers */
+.thumb::-webkit-slider-thumb {
+  -webkit-appearance: none;
+  pointer-events: all;
+  width: 20px;
+  height: 20px;
+  background-color: #fff;
+  border-radius: 50%;
+  border: 2px solid #1565ef;
+  border-width: 1px;
+  box-shadow: 0 0 0 1px #c6c6c6;
+  cursor: pointer;
+}
+
+/* For Firefox browsers */
+.thumb::-moz-range-thumb {
+  -webkit-appearance: none;
+  pointer-events: all;
+  width: 20px;
+  height: 20px;
+  background-color: #fff;
+  border-radius: 50%;
+  border: 2px solid #1565ef;
+  border-width: 1px;
+  box-shadow: 0 0 0 1px #c6c6c6;
+  cursor: pointer;
+}
+
+input::-webkit-outer-spin-button,
+input::-webkit-inner-spin-button {
+  -webkit-appearance: none;
+  margin: 0;
+}
+
+/* Firefox */
+input[type='number'] {
+  -moz-appearance: textfield;
+}
--- a/app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
+++ b/app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
@@ -25,8 +25,12 @@ export function ColorRangeSlider({
   colorMapScale,
   setColorMapScale
 }: ColorrangeRangeSlideProps) {
-  const [minVal, setMinVal] = useState(min);
-  const [maxVal, setMaxVal] = useState(max);
+  const [minVal, setMinVal] = useState(
+    colorMapScale?.min ? colorMapScale.min : min
+  );
+  const [maxVal, setMaxVal] = useState(
+    colorMapScale?.min ? colorMapScale.max : max
+  );
   const [inputError, setInputError] = useState({
     min: false,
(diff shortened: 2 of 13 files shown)


app/scripts/components/exploration/components/datasets/color-range-slider.scss
```diff
@@ -0,0 +1,62 @@
+
+.slider__track {
+  background-color: #dddfe2;
+}
+
+.slider__range {
+  background-color: #1565ef;
```
dzole0311, line 7: From a recent discussion with @faustoperez I think we would be fine if we use USWDS color utilities that approximate these hardcoded hexes, even if our designs don't strictly follow the USWDS color tokens yet (at least features that were designed a while ago).

I had a similar situation with the `react-calendar`, where the Figma designs used custom hex values, but we wanted to lean towards and try using their approximate USWDS counterparts. For this particular hex here (`#1565ef`) I suggest we use the `primary-vivid` token which comes pretty close (e.g. usage [here](https://github.com/NASA-IMPACT/veda-ui/blob/46e1c5df246aafec25b760e231938607cd1cd41e/app/scripts/components/common/datepicker/index.scss#L27). Sigh, I opened an issue to optimize the paths of these imports as well: https://github.com/NASA-IMPACT/veda-ui/issues/1196).


List of USWDS colors and tokens: https://designsystem.digital.gov/utilities/color/

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
```diff
@@ -0,0 +1,270 @@
+import React, { useCallback, useEffect, useState, useRef } from 'react';
+
+import './color-range-slider.scss';
+import { colorMapScale } from '$components/exploration/types.d.ts';
+
+import { USWDSTextInput } from '$components/common/uswds';
+
+interface ColorrangeRangeSlideProps {
```
dzole0311, line 8: `ColorRangeSliderProps` might be a better name?

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
```diff
@@ -0,0 +1,270 @@
...
+              ref={range}
+              className='slider__range display-flex flex-align-center flex-justify-center radius-md position-absolute height-05 z-200'
+            >
+              {/* Show divergent map middle point */}
+              {min < 0 ? (
+                <div
+                  ref={middleMarker}
+                  className='position-relative height-2 width-1px middle-marker'
+                />
+              ) : null}
```
dzole0311, line 217: How tricky would it be to have the `middleMarker` positioned in the center, at all times, even when dragging the thumbs? 

For example, I can see that the `middleMarker` is centered correctly within it's container, but in some cases (like in the screenshot) it looks misaligned because the left end of it's container is dynamically changed / set. Maybe tweaking the `left` and `width` values of the range ref could help? 

Screenshot 2024-10-17 at 12 20 28

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
```diff
@@ -0,0 +1,270 @@
...
+            className={`thumb ${thumbPosition} z-400`}
+            style={{
+              zIndex: minVal <= max - 10 * minMaxBuffer ? '500' : '400'
+            }}
+          />
+          <div className='slider width-card position-relative height-3 display-flex flex-align-center'>
+            <div className='slider__track radius-md position-absolute height-05 z-100 width-full display-flex flex-align-center' />
+            <div
+              ref={range}
+              className='slider__range display-flex flex-align-center flex-justify-center radius-md position-absolute height-05 z-200'
```
dzole0311, line 209: Found a small bug, steps to reproduce:

1. Add the Net Primary Production layer
2. Open the rescale pop up and reduce the max from 0.3 to somewhere in the middle (e.g 0.15)
3. Reload the page
4. Open the rescale pop up again
5. Notice that the range track is correctly set, but not the thumb for the max value


Screenshot 2024-10-17 at 12 28 47

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
```diff
@@ -0,0 +1,270 @@
...
+      !colorMapScale ||
+      (colorMapScale.max == max && colorMapScale.min == min)
+    ) {
+      setColorMapScale({ min: minVal, max: maxVal });
+    } else
+      setColorMapScale({
+        min: Number(minValRef.current),
+        max: Number(maxValRef.current)
+      });
+  }, [maxVal, minVal, getPercent, setColorMapScale, inputError, max, min]);
```
dzole0311, line 125: For this dependency array, I get a `React Hook useEffect has a missing dependency: 'colorMapScale'. Either include it or remove the dependency array.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)`

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
```diff
@@ -0,0 +1,270 @@
...
+  );
+  const range = useRef<HTMLDivElement>(null);
+  const middleMarker = useRef<HTMLDivElement>(null);
+
+  // Convert to percentage
+  const getPercent = useCallback(
+    (value) => Math.round(((value - min) / (max - min)) * 100),
+    [min, max]
+  );
+  //Calculate the range
```
dzole0311, line 55: Comments like this are a bit obsolete, especially when the utility function you wrote looks pretty clean and self-explanatory ;) Same for the `getPercent` above.

dzole0311 (changes requested):
Nice work @snmln! I left a couple of comments below

dzole0311:
@snmln Sorry, one more small nit: if we remove the min val from the input and then click outside of the input, could we revert back / show the default value instead of leaving the min input empty?

snmln:
> @snmln Sorry, one more small nit: if we remove the min val from the input and then click outside of the input, could we revert back / show the default value instead of leaving the min input empty?

As of right now that is the functionality on refresh. That shouldn't be hard to implement when the interface closes.

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
snmln: Good catch, there was a null check that was checking against the wrong prop and reseting the value for the range input.

app/scripts/components/exploration/components/datasets/colorRangeSlider.tsx
```diff
@@ -0,0 +1,272 @@
...
+    (value) => ((value - min) * 100) / (max - min),
+    [min, max]
+  );
+  //Calculate the range
+  const rangeCalculation = (maxPercent, minPercent) => {
+    const thumbWidth = 20;
+    if (range.current) {
+      range.current.style.width = `calc(${
+        maxPercent - minPercent >= 100 ? 100 : maxPercent - minPercent
+      }% - ${(thumbWidth - minPercent * 0.2) * (maxPercent / 100)}px )`;
```
snmln, line 55: There was a little bit of magic number math needed to get this working. Based off what I was able to locate in this link: https://css-tricks.com/value-bubbles-for-range-inputs/

Because the `left` and `width` are percentage we need to add a buffer of half the button width to both ends of the range. Doing so will ensure that the marker is centered when moving the left slider thumb. For the right slider thumb we needed to do the same but adjust the buffer value dynamically reduce the buffer on the right based on the max percent of the slider.

dzole0311 (approved):

PR #1199, opened 2024-10-18
1171 cookie consent clean up

snmln:
**Related Ticket:** https://github.com/orgs/NASA-IMPACT/projects/17/views/1?pane=issue&itemId=80952423&issue=NASA-IMPACT%7Cveda-ui%7C1171

### Description of Changes

- Relocating cookie retrieval, validation, and creation functionality from `layoutRoot` to `cookieConsentForm`
- Implementing `session item` to track when a session has begun for each individual user and only run validation and retrieval functionality on session start and when a response has not been recorded. 
- Adjusting Banner content to accept markdown in config props.  

### Validation / Testing
Clear out all cookies in your local machine to do so: open `dev tools` -> navigate to `application tab`-> navigate to `cookies` in the left panel -> right click `clear`
- After clearing your existing site cookies, navigate to any page. Click on **Decline**. Navigate to another page and you should not see the cookie consent form component. 
- Clearing your existing site cookies again, navigate to any page. Click on **Accept**. Navigate to another page and you should not see the cookie consent form component. 
- Clearing your existing site cookies again, navigate to any page. Click on **X**. Navigate to another page and you should see the cookie consent form component rerendered.

### DIFF
--- a/app/scripts/components/common/cookie-consent/cookieConsent.spec.js
+++ b/app/scripts/components/common/cookie-consent/cookieConsent.spec.js
@@ -3,10 +3,9 @@ import '@testing-library/jest-dom';
 
 import { render, screen, fireEvent } from '@testing-library/react';
 import { MemoryRouter } from 'react-router-dom'; // For testing
-
 import { createMemoryHistory } from 'history';
+import * as utils from './utils';
 
-import { COOKIE_CONSENT_KEY, SESSION_KEY } from './utils';
 import CookieConsent from './index';
 
 describe('Cookie consent form should render with correct content.', () => {
@@ -25,16 +24,36 @@ describe('Cookie consent form should render with correct content.', () => {
       pathname: 'localhost:3000/example/path'
     })
   }));
-
-  beforeEach(() => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  it('Check that session item is non existant prior to cookie consent render. Then confirm that cookie consent creates session item.', () => {
+    expect(sessionStorage.getItem(utils.SESSION_KEY)).toBeNull();
     render(
       <MemoryRouter history={history}>
         <CookieConsent {...cookieData} onFormInteraction={onFormInteraction} />
       </MemoryRouter>
     );
+    expect(sessionStorage.getItem(utils.SESSION_KEY)).toBe(`true`);
+  });
+  it('Check that getcookie is only called once on render and session item is true', () => {
+    const spy = jest.spyOn(utils, 'getCookie');
+
+    render(
+      <MemoryRouter history={history}>
+        <CookieConsent {...cookieData} onFormInteraction={onFormInteraction} />
+      </MemoryRouter>
+    );
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(sessionStorage.getItem(utils.SESSION_KEY)).toBe(`true`);
   });
 
   it('Renders correct content', () => {
+    render(
+      <MemoryRouter history={history}>
+        <CookieConsent {...cookieData} onFormInteraction={onFormInteraction} />
+      </MemoryRouter>
+    );
     expect(
       screen.getByRole('link', { name: 'Privacy Policy' })
     ).toHaveAttribute('href', 'https://www.nasa.gov/privacy/#cookies');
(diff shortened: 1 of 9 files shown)


dzole0311:
I noticed an issue when navigating between routes: the cookie consent temporarily disappears and then again reappears. Is this related to the relocation of the cookie functionality? I've recorded a video with the behavior below:

https://github.com/user-attachments/assets/fcb79f70-f740-4948-87b0-06b995cf5478

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -43,43 +49,16 @@ const PageBody = styled.div`
...
-    } else {
-      return !cookieContents.responded;
-    }
-  };
+  const bannerContent = getBannerFromVedaConfig();
   const { children } = props;
-
-
+  const [sessionStart, setSesstionStart] = useState<string | undefined>();
+  const sessionItem = window.sessionStorage.getItem(SESSION_KEY);
```
dzole0311, line 55: `window` is a client side object, so this will probably fail in our Next.js instance. Let's check if `window` is defined before trying to access it.

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -12,23 +14,50 @@ import './index.scss';
...
+  const readCookie = (name) => {
+    const nameEQ = name + '=';
+    const attribute = document.cookie.split(';');
+    for (let i = 0; i < attribute.length; i++) {
+      let c = attribute[i];
+      while (c.charAt(0) == ' ') c = c.substring(1, c.length);
+      if (c.indexOf(nameEQ) == 0) return c.substring(nameEQ.length, c.length);
+    }
+    return null;
+  };
```
dzole0311, line 40:
```suggestion
const readCookie = (name) => {
  const nameEQ = name + '=';
  const attribute = document.cookie.split(';');
  const cookie = attribute.find(cookie => cookie.trim().startsWith(nameEQ));
  return cookie ? cookie.substring(nameEQ.length).trim() : null;
};
```

dzole0311 (changes requested):

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -1,14 +1,20 @@
...
 import { useDeepCompareEffect } from 'use-deep-compare';
 import styled from 'styled-components';
 import { Outlet } from 'react-router';
 import { reveal } from '@devseed-ui/animation';
-import { getCookieConsentFromVedaConfig } from 'veda';
+import { getBannerFromVedaConfig, getCookieConsentFromVedaConfig } from 'veda';
 import MetaTags from '../meta-tags';
 import PageFooter from '../page-footer';
 import Banner from '../banner';
 import { CookieConsent } from '../cookie-consent';
```
hanbyul-here, line 16: Since we are on it, let's lazyload Banner and CookieConsent (the components that will render only when the configs are there)!

app/scripts/components/common/layout-root/index.tsx
hanbyul-here: Was this concern addressed?

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -42,67 +66,104 @@ export const CookieConsent = ({
...
+  const setSessionData = () => {
+    if (typeof window !== 'undefined') {
+      const checkForSessionDate = window.sessionStorage.getItem(SESSION_KEY);
+      if (!checkForSessionDate) {
+        window.sessionStorage.setItem(SESSION_KEY, 'true');
+      }
+    }
+  };
+  useEffect(() => {
+    if (sessionStart !== 'true' && !cookieConsentResponded) {
```
hanbyul-here, line 78: I wonder if there is any way to simplify the logic so the flow can be more clear. 
ex. is there any case that sessionStart is not true (which I think it means that this component was never rendered before? - please correct me if I am wrong) but `cookieConsentReponded` is true?

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { useState, useEffect } from 'react';
 import { Icon } from '@trussworks/react-uswds';
-import { COOKIE_CONSENT_KEY } from './utils';
+import { useLocation } from 'react-router-dom';
```
hanbyul-here, line 3: react-router-dom will be a trouble when exported as a library (and used in Next instance.) :[  What will be the best way to replace this logic? -

app/scripts/components/common/cookie-consent/cookieConsent.spec.js
```diff
@@ -36,11 +54,13 @@ describe('Cookie consent form should render with correct content.', () => {
 
   it('Check correct cookie initialization', () => {
     const resultCookie = document.cookie;
-
     expect(resultCookie).toBe(
       `${COOKIE_CONSENT_KEY}={"responded":false,"answer":false}`
     );
   });
+  it('Check for session initialization', () => {
```
hanbyul-here, line 61: I feel like this session value deserves more testing (it would have helped me understand what this session value is trying to do!) - Can we elaborate it?

app/scripts/components/common/cookie-consent/index.tsx
```diff
@@ -42,67 +66,104 @@ export const CookieConsent = ({
...
+      const checkForSessionDate = window.sessionStorage.getItem(SESSION_KEY);
+      if (!checkForSessionDate) {
+        window.sessionStorage.setItem(SESSION_KEY, 'true');
+      }
+    }
+  };
+  useEffect(() => {
+    if (sessionStart !== 'true' && !cookieConsentResponded) {
+      setSessionData();
+      getCookie();
```
hanbyul-here, line 80: nipick - the name of function can be clearer imo. getCookieis really not working as a getter, would there be any better name to reflect what it does?

app/scripts/components/common/cookie-consent/index.tsx
snmln: That is correct the only instance where `sessionStart` would be `false` is if the component was never rendered. Leading to a session token never being set. There is the potential that the `cookieConsentReponded` is true based on previous input in the form. But if they have already responded the component should not display in the browser.

hanbyul-here:
👋 Apologies for the review being late. I am somehow having trouble running this branch locally. I doubt that it is something this pr iontroduced since our dev environment has been unstable time to time. But in case, can you confirm me that you have no problem running this branch on local ?

snmln:
Hey @hanbyul-here no problem, odd I was able to run it previously but now the system is getting hung up on the bundling step locally.

hanbyul-here:
@snmln I did rebase main, but I still have a problem running a local instance. trying to figure out what is going on now.

hanbyul-here (approved):
🙇 thanks for working on it!

dzole0311 (approved):
Re-tested the cookie consent across routes and page reloads: it renders as expected.

PR #1205, opened 2024-10-21
Fix map controls and resolve Analyses Message issue

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1155 https://github.com/NASA-IMPACT/veda-ui/pull/1172

Related Next.js PR: https://github.com/developmentseed/next-veda-ui/pull/7

### Description of Changes

- Fixed the map controls so that they work in the Next.js instance
- Resolve issue with the Analyses Message display
- Attempted to address the React mount error mentioned in https://github.com/NASA-IMPACT/veda-ui/pull/1172

### Validation / Testing
1. Verify that the AOI controls and Analyses Message work as expected in both veda-ui and the Next.js instance
2. Verify that adding, deleting a single drawn area, or deleting all drawn areas works without issues
3. Ensure that analyses execute as expected

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
@@ -325,33 +325,20 @@ export default function CustomAoIControl({
   const { main } = useMaps();
   const { isDrawing } = useAois();
 
-  // Start/stop the drawing.
+  // Start the drawing mode when isDrawing is true
+  // There's no need to switch back to 'simple_select' mode when !isDrawing
+  // as Mapbox Draw handles this internally when the drawing is completed
   useEffect(() => {
     if (!main) return;
 
-    const changeDrawMode = () => {
-      const mbDraw = main._drawControl;
-      if (!mbDraw) return;
+    const mbDraw = main._drawControl;
 
-      if (isDrawing) {
-        mbDraw.changeMode(DRAW_POLYGON);
-      } else {
-        mbDraw.changeMode(SIMPLE_SELECT, {
-          featureIds: mbDraw.getSelectedIds()
-        });
-      }
-    };
+    if (!mbDraw) return;
 
-    if (main.loaded()) {
-      changeDrawMode();
-    } else {
-      main.on('load', changeDrawMode);
-      return () => {
-        main.off('load', changeDrawMode);
-      };
+    if (isDrawing) {
+      mbDraw.changeMode(DRAW_POLYGON);
     }
   }, [main, isDrawing]);
-
 
   useThemedControl(
     () => <CustomAoI map={main} disableReason={disableReason} />,
--- a/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
@@ -323,23 +323,35 @@ export default function CustomAoIControl({
   disableReason?: React.ReactNode;
 }) {
   const { main } = useMaps();
-
   const { isDrawing } = useAois();
 
   // Start/stop the drawing.
   useEffect(() => {
-    // Property was added to access draw control.
-    const mbDraw = main?._drawControl;
-    if (!mbDraw) return;
+    if (!main) return;
 
-    if (isDrawing) {
-      mbDraw.changeMode(DRAW_POLYGON);
+    const changeDrawMode = () => {
+      const mbDraw = main._drawControl;
+      if (!mbDraw) return;
+
+      if (isDrawing) {
+        mbDraw.changeMode(DRAW_POLYGON);
+      } else {
+        mbDraw.changeMode(SIMPLE_SELECT, {
+          featureIds: mbDraw.getSelectedIds()
+        });
+      }
+    };
+
+    if (main.loaded()) {
+      changeDrawMode();
     } else {
-      mbDraw.changeMode(SIMPLE_SELECT, {
-        featureIds: mbDraw.getSelectedIds()
-      });
+      main.on('load', changeDrawMode);
+      return () => {
+        main.off('load', changeDrawMode);
+      };
     }
   }, [main, isDrawing]);
+
 
   useThemedControl(
     () => <CustomAoI map={main} disableReason={disableReason} />,
@@ -347,5 +359,7 @@ export default function CustomAoIControl({
       position: 'top-left'
     }
   );
+
   return null;
 }
+
--- a/app/scripts/components/common/map/controls/aoi/index.tsx
+++ b/app/scripts/components/common/map/controls/aoi/index.tsx
@@ -2,13 +2,18 @@ import MapboxDraw from '@mapbox/mapbox-gl-draw';
 import StaticMode from '@mapbox/mapbox-gl-draw-static-mode';
 import { useTheme } from 'styled-components';
 import { useAtomValue } from 'jotai';
-import { useEffect } from 'react';
+import { useRef } from 'react';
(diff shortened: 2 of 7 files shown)


app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -323,29 +323,43 @@ export default function CustomAoIControl({
...
+      } else {
+        mbDraw.changeMode(SIMPLE_SELECT, {
+          featureIds: mbDraw.getSelectedIds()
+        });
+      }
+    };
 
-    if (isDrawing) {
-      mbDraw.changeMode(DRAW_POLYGON);
+    if (main.loaded()) {
```
hanbyul-here, line 345: Ah, I see; so the problem was that we couldn't guarantee that the map was loaded before registering mapbox-draw-related events?

app/scripts/components/common/map/controls/aoi/index.tsx
```diff
@@ -75,5 +67,14 @@ export default function DrawControl(props: DrawControlProps) {
     }
   );
 
+  useEffect(() => {
+    if (drawControl?.map) {
```
hanbyul-here, line 71: I am having hard time understanding what this does. What is the expected value of `drawControl.map` and when is the case that drawControl has `map` attribute?

dzole0311: Sorry about the confusion and I should have double-checked before pushing. I’ve cleaned up the code here and it should make more sense hopefully.

I was looking in the `useControl` [hook](https://github.com/visgl/react-map-gl/blob/7.1-release/src/components/use-control.ts#L28) from the `react-map-gl` repo and saw that `IControl` (the type for `DrawControl`) has `export interface IControl<MapT extends MapInstance = MapInstance>`. And I misread it assuming that `drawControl` had a direct map prop which isn’t actually the case.. The map instance is only passed into the lifecycle methods (like the `onAdd` for example) and it isn’t exposed directly on the `drawControl` object.

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
dzole0311: Yes, I think the issue came from two places:

1. We couldn’t guarantee the map was fully loaded before adding the Mapbox Draw events. So I added this check to make sure `main.loaded()` is `true` before switching modes when `DrawControl` is initialized
2. There was a strange synchronization issue in the previous code that initialized `useControl`. Specifically `control.current` was `undefined` later in the `CustomAoiControl` and adding safeguards or null checks didn't fix the problem. So I ended up [assigning the control instance](https://github.com/NASA-IMPACT/veda-ui/pull/1205/files#diff-d1b613ffea6e16bb9dcd389b291cc036a6e838535e87113bb3c43de86e7c2990R62) returned by `useControl` directly to a variable (like `const drawControl = useControl(...)`) to make sure we're always using the correct and initialized control instance

app/scripts/components/common/map/controls/aoi/index.tsx
```diff
@@ -47,33 +52,38 @@ export default function DrawControl(props: DrawControlProps) {
         },
         ...props
       });
-      return control.current;
+      return control;
     },
-    ({ map }: { map: any }) => {
-      map._drawControl = control.current;
+    ({ map }) => {
+      mapRef.current = map as ExtendedMapRef;
```
hanbyul-here, line 58: Where is `mapRef` used? Am I missing anything?

app/scripts/components/common/map/controls/aoi/index.tsx
```diff
@@ -47,33 +52,38 @@ export default function DrawControl(props: DrawControlProps) {
...
+      const extendedMap = map as ExtendedMapRef;
```
hanbyul-here, line 59: Any specific reason you decide to use object reference here? I tend to lean towards not using object reference (I think it doesn't help readability and makes it challenging to see where the states are mutated), but I would like to hear your thoughts behind the decision!

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
hanbyul-here: oh- interesting. It will be great to figure out why 2. is happening (not a blocker, though!)

app/scripts/components/common/map/controls/aoi/index.tsx
dzole0311: Removed

app/scripts/components/common/map/controls/aoi/index.tsx
hanbyul-here: Yeah, I was wondering if ts was the reason, and I think your workaround should work 👍

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -323,29 +323,43 @@ export default function CustomAoIControl({
...
+    const changeDrawMode = () => {
+      const mbDraw = main._drawControl;
+      if (!mbDraw) return;
+
+      if (isDrawing) {
+        mbDraw.changeMode(DRAW_POLYGON);
+      } else {
+        mbDraw.changeMode(SIMPLE_SELECT, {
+          featureIds: mbDraw.getSelectedIds()
+        });
```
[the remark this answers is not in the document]
dzole0311: The original [commit](https://github.com/NASA-IMPACT/veda-ui/pull/767/commits/89d23479beef87d5aab774bb6a9ff776eb96fd98) addressed an issue where polygons were not being selected after completing the drawing by double-clicking. I tried this and it is reproducible locally using this [line](https://github.com/NASA-IMPACT/veda-ui/pull/767/commits/89d23479beef87d5aab774bb6a9ff776eb96fd98#diff-7c073c4bc2683442a5f008251eeac5bab257f96acb39d9778fe62a08f4675392L129). 

However you're correct that the switch to `simple_select` mode is redundant as it seems to be managed internally by Mapbox Draw GL (note: the explicit switch might have been needed back when the fix was implemented). So this behavior makes chunks of the code unnecessary. I removed the obsolete parts and added a comment.

sandrahoang686:
@dzole0311 unsure how to test then? could we re-comment out what was causing that error again and add an `@TODO` to fix that because this would block E&A then. Looking at the original PR hanbyul opened, doesn't look like we have a ticket for it either. We would need to create that and have it as part of shipping/exposing E&A

sandrahoang686:
> @dzole0311 unsure how to test then? could we re-comment out what was causing that error again and add an `@TODO` to fix that because this would block E&A then. Looking at the original PR hanbyul opened, doesn't look like we have a ticket for it either. We would need to create that as part of a follow-up and have it as part of shipping/exposing E&A

sandrahoang686:
@dzole0311 this works in safari for me but in chrome I keep seeing that error even after hard refreshing and going in-cognito so must be something on my end but it looks to be working 👍🏼

sandrahoang686 (approved):

dzole0311:
@sandrahoang686 Thanks again for testing. I removed my fix attempt as it did not fix the issue, and opened a new ticket with details since it occurs in the stories as well as the E&A page: https://github.com/NASA-IMPACT/veda-ui/issues/1221

hanbyul-here:
aha I think I am seeing the problem you saw @sandrahoang686 . Each component works separately (ex. only modal was used  , only EA was used) but when they are put together, they don't seem to stably work together. I am looking at the issue 🔍

Issue #1207, opened 2024-10-22
AOI Feature Specification


AliceR:
## Context
While starting to investigate a bug related to the AOI (Area of Interest) feature, I noticed several other related issues that may be interconnected. To effectively address the root cause of these issues, it would be beneficial to clearly define the expected behavior of the AOI feature.

_I used the [Gherkin syntax](https://cucumber.io/docs/gherkin/reference/), which will be helpful when writing tests for the code as well._

## Acceptance Criteria
```
Feature: Area of Interest (AOI) Analysis

  Scenario: User selects a pre-defined AOI
    Given that I am on the map view (explore page)
    When I select the "Analyze an area" tool (Dropdown menu)
    And choose one of the listed regions
    Then map should display the selected area as the AOI
    And the AOI should not be editable when clicking on it
    And the analysis message should display the size of the area
    And the 'Delete all areas' button should be shown
    And the 'Run analysis' button should be shown

  Scenario: User tries to edit a pre-defined AOI
    Given that I am on the map view (explore page)
    When I add a pre-defined AOI
    And I click on the pre-defined AOI
    Then the AOI should remain un-editable (visually)

  Scenario: User draws an AOI when a pre-defined AOI exists
    Given that there is a pre-defined AOI on the map
    When I select the "Draw an area of interest" tool (Pen icon)
    Then the AOI from pre-defined AOIs should be deleted automatically 
    And the pre-defined selector should be reset and display the placeholder text

  Scenario: User draws an AOI
    Given that I am on the map view (explore page)
    When I select the "Draw an area of interest" tool (Pen icon)
    And I draw a polygon on the map
    Then map should display the drawn area as the AOI
    And the analysis message should display the size of the area

  Scenario: User selects an AOI
    Given that I am on the map view (explore page)
    And I have drawn one or several AOIs
    When I click on a currently unselected polygon on the map
    Then that AOI is selected 
    And all other AOIs are unselected
    And the button to delete says "Delete Area"

  Scenario: User selects multiple AOIs
    Given that I am on the map view (explore page)
    And I have drawn one or several AOIs
    When I hold shift and click on a currently unselected polygon on the map
    Then that AOI is selected 
    And all other AOIs previously selected remain selected

  Scenario: User _un_selects an AOI
    Given that I am on the map view (explore page)
    And I have drawn AOIs with at least one AOI selected
    When I click anywhere else on the map
    Then any selected AOI is unselected
    And the button to delete says "Delete all areas"

[...]

Issue #950, opened 2024-05-10
[E&A] Preselection of presets and hand-drawn polygons across page reloads


dzole0311:
1. When a preset is selected and the page is reloaded, the preset option is not shown as selected. We should probably keep this selection as a url parameter ("preset=Alaska") and then load it in the PresetSelector, so that the user knows which state was selected. 
2. Additionally, we should highlight the selected hand-drawn geometries on the map across page reloads. If I draw 3 polygons, select 2 of them and reload the page, they are not again selected due to diverging states between MapboxDraw and our React atoms. We already keep the state of the selected aois in the url, so we should switch those to "simple_select" mode so that they can be automatically highlighted
3. It might be a good idea to differentiate the colors of the geometries based on if they are "static" (presets, file uploads) or hand-drawn (@faustoperez any thoughts?)

faustoperez:
> It might be a good idea to differentiate the colors of the geometries based on if they are "static" (presets, file uploads) or hand-drawn (@faustoperez any thoughts?)

To simplify the work I suggest (for now) that they be differentiated by the presence of the edit handles, but use the same "this is a shape" stroke and fill color 👍

AliceR:
Hi @dzole0311 ! Thanks for outlining these issues! Here’s how the new proposed user flow (#1395)  addresses points 2 and 3, along with some thoughts on the first issue:

2. Highlighting selected hand-drawn geometries across reloads:
The new approach simplifies AOI management by always having a single active AOI. Since multiple selectable polygons are no longer supported, the issue of diverging states between MapboxDraw and our React atoms is resolved. Any drawn geometry is (after confirmation) treated as the active AOI, synced with the URL, and reloaded automatically as part of the map state.

3. Differentiating geometry colors based on source:
Since all AOIs—whether presets, file uploads, or hand-drawn—are now treated equivalently in the new workflow, it wouldn’t make sense to differentiate them by color. All AOIs are equal in terms of functionality and use in the analysis. Adding a visual distinction could introduce unnecessary complexity or confusion. The consistent treatment of AOIs ensures a simpler and more streamlined user experience.

---

1. Dropdown indicating selected preset upon reload
This one is worth discussing further. Once a preset is selected, it essentially becomes "just an AOI" and is treated the same way as any other AOI, with no dynamic link to the preset. Because of this, I’d suggest not persisting the dropdown selection upon reload. Doing so might confuse users by implying that the preset remains active as something separate from the AOI.

[...]

PR #1215, opened 2024-10-24
Pre-select AOIs that have the "selected" prop set to true

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/950 https://github.com/US-GHG-Center/veda-config-ghg/issues/406

### Description of Changes
The fix make sures that selected AOIs persist across page reloads. Previously when an AOI was selected, the selection was lost after a page reload (visually, the polygon was de-selected). The fix checks for existing AOI selections on load and adjusts the mode accordingly.

We also started to enhance the documentation and tests around AOI behavior (thanks to @AliceR): https://github.com/NASA-IMPACT/veda-ui/issues/1207

### Validation / Testing
1. Draw a polygon
2. Once you finish drawing, see that the AOI is selected
3. Reload the page
4. Verify that the AOI is still selected
5. Choose a preselected area (one of the state presets)
6. Reload the page and verify that it remains un-selectable

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
@@ -102,7 +102,7 @@ function CustomAoI({
     return () => {
       map.off('draw.selectionchange', onSelChange);
     };
-  }, [map, selectedForEditing]);
+  }, [selectedForEditing]);
 
   const resetAoisOnMap = useCallback(() => {
     const mbDraw = map?._drawControl;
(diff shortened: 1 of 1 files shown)


AliceR (approved):
This seems to work!
It is not zooming to the area, but that seems to be a different issue.

Also, I noticed we have the `map` object in the list of [dependencies](https://github.com/NASA-IMPACT/veda-ui/pull/1215/files#diff-7c073c4bc2683442a5f008251eeac5bab257f96acb39d9778fe62a08f4675392R105) for the useEffect, and I don't think it should be there. We want the effect only to listen for changes of the selected state, don't we? Again, not related to this change per se, but might be the reason for some flaky behavior around aois?

dzole0311:
> We want the effect only to listen for changes of the selected state, don't we? Again, not related to this change per se

Good call, I removed the dependency. Let's see if it improves some of the AOI behavior

> It is not zooming to the area, but that seems to be a different issue.

Yeah, it wasn't mentioned in both related issues so I didn't add it here

PR #1216, opened 2024-10-24
Add tests for the AOI feature specification

AliceR:
**Related Ticket:** #1207

### Description of Changes
So far, added a first test for the scenario "User selects a pre-defined AOI". I'm adding some `data-testid` properties to relevant dom elements in order to select them in the test.

Also, added some documentation on how to run the tests. @hanbyul-here if you want to try that on your machine, and let me know if it works? 

more tests to be added

### Validation / Testing
Follow the instructions added to the SETUP.md guide to run the tests locally. See if they pass!

### DIFF
--- a/app/scripts/components/exploration/components/map/analysis-message-control.tsx
+++ b/app/scripts/components/exploration/components/map/analysis-message-control.tsx
@@ -78,7 +78,9 @@ const MessageStatusIndicator = styled.div<MessageStatusIndicatorProps>`
     }
   }}
 `;
-const MessageContent = styled.div`
+const MessageContent = styled.div.attrs({
+    'data-testid': 'analysis-message'
+  })`
   line-height: 1.5rem;
   max-height: 1.5rem;
 
--- a/docs/development/SETUP.md
+++ b/docs/development/SETUP.md
@@ -62,6 +62,20 @@ VEDA-UI includes a GitHub action for checking TypeScript and lint errors. If you
 #### Pre-commit Hook
 Additionally, there's a pre-commit hook that performs the same error checks. This helps developers identify and address issues at an earlier stage. If you need to bypass the check (to make a local temporary commit etc.), include the `--no-verify`` flag in your commit command.
 
+### Testing
+
+## Unit tests
+
+`yarn test`
+
+## End-to-end tests
+Make sure the development server is running (`yarn serve`)
+
+`yarn test:e2e --ui`
+
+Alternatively, you can install the playwright extension for vs code (ms-playwright.playwright) and run the tests directly from there. It allows to run the tests in watch mode, open the browser or trace viewer, set breakpoints,... 
+Again, make sure that the development server is running (`yarn serve`).
+
 ## Deployment
 To prepare the app for deployment run:
 
--- a/test/playwright/pages/explorePage.ts
+++ b/test/playwright/pages/explorePage.ts
@@ -14,7 +14,7 @@ export default class ExplorePage {
     this.mapboxCanvas = this.page.getByLabel('Map', { exact: true });
     this.firstDatasetItem = this.page.getByRole('article');
     this.closeFeatureTourButton = this.page.getByRole('button', { name: 'Close feature tour' });
-    this.presetSelector = this.page.locator('#preset-selector');
+    this.presetSelector = this.page.getByTestId('preset-selector');
   }
 
   async closeFeatureTour() {
--- a/test/playwright/tests/exploreAoi.spec.ts
+++ b/test/playwright/tests/exploreAoi.spec.ts
@@ -23,33 +23,38 @@ test.describe('Area of Interest (AOI) Analysis', () => {
   );
 
   test('User selects a pre-defined AOI', async ({ page }) => {
-    // When I select the "Analyze an area" tool (Dropdown menu)
-    const toolbar = page.getByTestId('preset-selector');
-    // And choose one of the listed regions
-    await toolbar.selectOption('Hawaii');
+    await test.step('When I select the "Analyze an area" tool (Dropdown menu)', async () => {
+      const toolbar = page.getByTestId('preset-selector');
 
-    // Then map should display the selected area as the AOI
-    // const aoi = await page.$('selector-for-aoi'); // Adjust the selector as needed
-    // expect(aoi).not.toBeNull();
+      await test.step('And choose one of the listed regions', async () => {
+        toolbar.selectOption('Hawaii');
+      });
+    });
 
-    // // And the AOI should not be editable when clicking on it
-    // await page.click('selector-for-aoi'); // Adjust the selector as needed
-    // const isEditable = await page.$eval('selector-for-aoi', el => console.log(el)); // Adjust the selector as needed
-    // expect(isEditable).toBe(false);
+    await test.step('Then the map should display the selected area as the AOI', async () => {
+      // const aoi = await page.$('selector-for-aoi'); // Adjust the selector as needed
+      // expect(aoi).not.toBeNull();
 
-    // And the analysis message should display the size of the area
-    const analysisMessage = await page
-      .getByTestId('analysis-message')
-      .textContent()
+      await test.step('And the AOI should not be editable when clicking on it', async () => {
+        // await page.click('selector-for-aoi'); // Adjust the selector as needed
+        // const isEditable = await page.$eval('selector-for-aoi', el => console.log(el)); // Adjust the selector as needed
+        // expect(isEditable).toBe(false);
+      });
+    });
 
-    expect(analysisMessage).toContain('An area of 17 K km2 is selected.');
+    await expect(
+      page.getByTestId('analysis-message'),
+      'And the analysis message should display the size of the area'
+    ).toHaveText(/An area of 17 K km2/i);
 
-    // And the 'Delete all areas' button should be shown
-    const deleteButton = await page.$('text=Delete all areas');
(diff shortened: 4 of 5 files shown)


AliceR:
Hi @stephenkilbourn 👋 Thanks for setting up playwright and the basic functionality to test the explore page! It was very useful!

test/playwright/tests/exploreAoi.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      page.on('pageerror', (exception) => {
+        console.log(`Uncaught exception: "${JSON.stringify(exception)}"`);
+        pageErrorCalled = true;
+      });
+
+      const mapboxResponsePromise = page.waitForResponse(
+        /api\.mapbox.com\/v4\/mapbox\.mapbox-streets-v8/i
+      );
+
+      // Given that I am on the map view (explore page)
```
stephenkilbourn, line 17: if we want the gherkin steps to show up in the test report, we can wrap these in the playwright [test.step](https://playwright.dev/docs/api/class-test#test-step)

```
test('Given that I am on the map view (explore page)', async ({ page }) => {
    await page.goto('/exploration');
});
and that text appears in the saved reports

File: test/playwright/tests/exploreAoi.spec.ts [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+
+    // And the analysis message should display the size of the area
+    const analysisMessage = await page
+      .getByTestId('analysis-message')
+      .textContent()
+
+    expect(analysisMessage).toContain('An area of 17 K km2 is selected.');
+
+    // And the 'Delete all areas' button should be shown
+    const deleteButton = await page.$('text=Delete all areas');
```
stephenkilbourn, line 48: Playwright recommends using their locators that will auto-retry
`page.getByText('Delete all areas');`

or to validate that a button role was used for this
`page.getByrole('button', {name: Delete all areas'});`

docs/development/SETUP.md
```diff
@@ -62,6 +62,20 @@ VEDA-UI includes a GitHub action for checking TypeScript and lint errors. If you
 #### Pre-commit Hook
 Additionally, there's a pre-commit hook that performs the same error checks. This helps developers identify and address issues at an earlier stage. If you need to bypass the check (to make a local temporary commit etc.), include the `--no-verify`` flag in your commit command.
 
+### Testing
```
stephenkilbourn, line 65: 👏

test/playwright/tests/exploreAoi.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      await page.goto('/exploration');
+      await consentBanner.acceptButton.click();
+      await datasetSelectorComponent.addFirstDataset();
+      await explorePage.closeFeatureTour();
+    }
+  );
+
+  test('User selects a pre-defined AOI', async ({ page }) => {
+    // When I select the "Analyze an area" tool (Dropdown menu)
+    const toolbar = page.getByTestId('preset-selector');
```
stephenkilbourn, line 27: I think we can update the explorePage.ts to use this better locator

test/playwright/tests/exploreAoi.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    // await page.click('selector-for-aoi'); // Adjust the selector as needed
+    // const isEditable = await page.$eval('selector-for-aoi', el => console.log(el)); // Adjust the selector as needed
+    // expect(isEditable).toBe(false);
+
+    // And the analysis message should display the size of the area
+    const analysisMessage = await page
+      .getByTestId('analysis-message')
+      .textContent()
+
+    expect(analysisMessage).toContain('An area of 17 K km2 is selected.');
```
stephenkilbourn, line 45: we can also make this gherkin step explicit in the report: 
```
expect(analysisMessage. 'the analysis message should display the size of the area').toContain('An area of 17 K km2 is selected.');

File: test/playwright/tests/exploreAoi.spec.ts [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+
+    // And the 'Delete all areas' button should be shown
+    const deleteButton = await page.$('text=Delete all areas');
+    expect(deleteButton).not.toBeNull();
+
+    // And the 'Run analysis' button should be shown
+    const runButton = await page.$('text=Run analysis');
+    expect(runButton).not.toBeNull();
```
stephenkilbourn, line 53: same toBeVisible comment

app/scripts/components/exploration/components/map/analysis-message-control.tsx
```diff
@@ -78,7 +78,9 @@ const MessageStatusIndicator = styled.div<MessageStatusIndicatorProps>`
     }
   }}
 `;
-const MessageContent = styled.div`
+const MessageContent = styled.div.attrs({
+    'data-testid': 'analysis-message'
+  })`
```
stephenkilbourn, line 83: this wasn't getting rendered.  looks like you need to spread it on the MessageContent component in line 200:
```
<MessageContent {...props}>
```

test/playwright/tests/exploreAoi.spec.ts
stephenkilbourn: oh, strike that, the text was adding in the date range so the test failed.  This limits it to the area, and now the test passed for me: 
```
    await expect(page.getByTestId('analysis-message'), 'the analysis message should display the size of the area').toHaveText(/An area of 17 K km2/i)
```

stephenkilbourn (changes requested):
Great @AliceR ! I'm happy that they are proving useful and you're brave enough to be the first one to take this on. 

I added a few suggestions.  I got it passing locally with getting the testid rendering in the analysis message and tweaking the assertion.  If anything is confusing, let me know and I can chat via slack.

test/playwright/tests/exploreAoi.spec.ts
AliceR: That was an interesting situation, as the message gets displayed with the simple area statement and then updated with the date range later. I wasn't sure how to handle the situation in the test, as it would most of the times be quick and catch the first statement and not wait for the later update. 
Making the expectation generic enough to accept both statements works 👍

app/scripts/components/exploration/components/map/analysis-message-control.tsx
AliceR: Are you sure it does not show up in the DOM? 
Screenshot 2024-10-28 at 11 34 51 
It does for me... and I don't want to touch all the places where the `MessageContent` component is being used, if its not really necessary. From the [styled component docs](https://styled-components.com/docs/faqs#when-to-use-attrs) it seems we don't need to spread the props, but please let me know if you come to a different conclusion!

stephenkilbourn: I pulled down your latest commits and re-ran things, and I see it now.  I don't know why I couldn't before. 🤷‍♂️

test/playwright/pages/explorePage.ts
```diff
@@ -14,7 +14,7 @@ export default class ExplorePage {
     this.mapboxCanvas = this.page.getByLabel('Map', { exact: true });
     this.firstDatasetItem = this.page.getByRole('article');
     this.closeFeatureTourButton = this.page.getByRole('button', { name: 'Close feature tour' });
-    this.presetSelector = this.page.locator('#preset-selector');
+    this.presetSelector = this.page.getByTestId('preset-selector');
```
stephenkilbourn, line 17: I was having to set this to get the parent for it to run.
```
const toolbar = page.locator('#preset-selector');
```

otherwise, the test would fail for me.  I'd prefer to use your test Id if it is working for you, though.

AliceR: I don't have any issues with using the test id. What exactly is the problem that you are seeing?

stephenkilbourn: The tests always fails unless I switch to locating by ID.  I've attached the trace from what I see after syncing with origin and deleting and re-installing node_modules to try minimizing a chance of mismatch.

the only data-testid I see is on the analysis message
Screenshot 2024-10-29 at 9 22 25 AM

the select has the id, but no data-testid.  
Screenshot 2024-10-29 at 9 23 33 AM

This is the test run with `yarn playwright test exploreAoi.spec.ts`

[traceAOI.zip](https://github.com/user-attachments/files/17559393/traceAOI.zip)

AliceR: Screenshot 2024-10-31 at 11 44 12
This is very strange, the data-testid does show up as expected for me. Can you try to run the whole test suite, instead of only the exploreAoi.spec.ts? Also, I assume you did re-start the local server with `yarn serve` at some point? Maybe somehow the local build isn't updating as it should? Wonder what else it could be that is different for me...

dzole0311: I ran the whole test suite locally including running this test individually, and it always passes with a ✅ in my case.

Screenshot 2024-10-31 at 12 12 33

AliceR:
I’d like to go ahead and merge this PR now, if it works for you @stephenkilbourn. The CI passes ✅ ! I think it makes sense to get this integrated so we can focus on bug fixing and adding new features. There are definitely more issues around AOI that we need to tackle, and we can add more tests later as we work through them.
Merging now will help us keep things moving forward and allow us to build on this test suite as we go.

stephenkilbourn (approved):

Issue #1028, opened 2024-07-02
[Design / UX] Gather requirements to understand scientist/decision maker workflows


faustoperez:
### Description
In recent conversations with NASA, there is a need to target decision-makers as well as scientists on the US GHG Center website in the short-mid term.

### Acceptance criteria

- Understand the needs of the decision-maker group of users.
- List goals and expected outcomes of the US GHG Center website visit.
- Understand how the experience is different from scientific users.
- List the existing content we can surface to improve their experience.

faustoperez:
This ticket is being reviewed together with https://github.com/US-GHG-Center/ghgc-architecture/issues/290

aboydnw:
@faustoperez based on the meeting today, I don't think we are going to get meaningful review time from the GHG team on this topic until after R3. Should we close this out?

Alternatively, maybe we can review with @freitagb since part of this speaks to the longer term vision of the product. What do you think?

faustoperez:
I think both options are compatible.

I am linking the Figma with the exploration work we did a few weeks ago 👉 https://www.figma.com/design/tcCUA06Knwq7nA9DUOlplo/US-GHG-Center-test-file?node-id=50-480&t=NVZ0Nn7FPw46DbOa-1 and invited @freitagb 

I'll close the ticket for now 🙂

PR #1218, opened 2024-10-25
Handle the case when source parameters + render parameters are both empty

hanbyul-here:
**Related Ticket:** #1028 

### Description of Changes
- It turns out that there are datasets (such as globalscope bluetarp) that should not have any titiler parameters. 
Bluetarp data from stac backend: https://staging.openveda.cloud/api/stac/collections/blue-tarp-planetscope

This makes me wonder about if we can validate anything about render parameters - ex. When there is no render parameter at all from stac back-end points, is there any way to be sure that this dataset is meant to be without render parameters ? - Tagging @smohiudd for this question.


dzole0311 (approved):
Tested via veda-config and datasets are loaded properly.

Issue #1186, opened 2024-10-08
[Refactor] Make exposing type interfaces as part of library build work


sandrahoang686:
veda-ui needs to be able to expose type interfaces so they can be accessible to instances like the nextJs instance. For example we should expose these types here https://github.com/NASA-IMPACT/veda-ui/blob/c80d6a305ac19b45c95dec51f9773b65aafa48ae/app/scripts/components/common/page-header/types.d.ts so we can type  the nav configuration on the instance side. 

We also dont seem to have type hinting when using these veda-ui components.

PR #1223, opened 2024-10-29
Update tsconfig and make nav interfaces exposable for consumption

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1186
**Related PR:** https://github.com/developmentseed/next-veda-ui/pull/10

**Note**
To expose interfaces from veda-ui - they should be defined in `.ts` files vs `.d.ts` files. This is because parcel isn't building the typescript definition files into `lib/index.d.ts` but does so when the file is `.ts`. 

**When building the library with `.d.ts` extension - there are NO type hints from NextInstance side**
Screenshot 2024-11-04 at 9 23 19 AM

**When building the library with `.ts` extension - there are type hints form NextInstance side**
Screenshot 2024-11-04 at 9 20 02 AM

### Description of Changes
* Update typescript definition file of nav types to typescript file
* Add declaration true to tsconfig

Found this thread about typescript definition files and parcel build here https://github.com/parcel-bundler/parcel/issues/1240

### Validation / Testing


dzole0311 (approved):

dzole0311:
Thanks for fixing, I'm getting the correct type annotations now ✅

hanbyul-here (approved):
I checked the built asset (`lib/index.d.ts`) has Nav related types. thanks for working on it 🙇 Maybe easy lint warnings can be fixed (ex. linebreaks) can be fixed since we are on it? but not a blocker !

tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "moduleResolution": "node",
     "esModuleInterop": true,
     "strictNullChecks": true,
+    "declaration": true,
```
hanbyul-here, line 8: for my education, can you tell me what this line does?

sandrahoang686: here is more about it https://www.typescriptlang.org/tsconfig/declaration.html

Issue #1220, opened 2024-10-28
[Refactor] Connect E&A flow with Data Layer Selection in NextJs Instance and state persists


sandrahoang686:
There are a few bugs/improvements that we still need to tackle to make exportable E&A page work. 
 The identified items so far are: 
- MapboxDraw not added to the map yet when its method is called
- Date value that gets fed to atom is not valid
- TimelineDatasetAtom still depends on veda faux module datasets

PR #1229, opened 2024-10-30
Various fixes for E&A

hanbyul-here:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1220

### Description of Changes
This PR is WIP, opening in case it helps anybody. (That says, please don't spend time reviewing the changes in  this PR yet, but it will be really helpful if you can validate that this branch can run with Next instance with stability.)

I've noticed E&A still doesn't run with stability in Next instance. I traced down all the possible errors that I can spot. That includes 
#### Fixed in this PR
- export useAtom hook for timelinedataset (to remove the possibility of more than one jotai instances)
- LinkProperties problem 
- MapboxDraw not added to the map yet when its method is called
- Date value that gets fed to atom is not valid  - this was because our polyfill for Array.last was not getting exported

#### Not fixed in this PR yet

- TimelineDatasetAtom still depends on veda faux module datasets 

 I put temporary work-arounds for unsolved problems to validate that those are the errors that can interrupt Next instance. 

### Validate & Test
Please test this branch with `test-branch` Next branch : https://github.com/developmentseed/next-veda-ui/tree/test-branch  - It will still lack of some functionalities, but hopefully the e&a page doesn't return abrupt errors anymore.


app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -86,6 +86,7 @@ function CustomAoI({
   // from feature to point.
   const [, forceUpdate] = useState(0);
   useEffect(() => {
+    if(!map.loaded()) return;
```
dzole0311, line 89: I suspect the issue with the AOIs disappearing is related to the `flyTo` method from Mapbox and/or the way (and when) we invoke it. I've removed it in all cases where we call the `flyTo` and I'm not able reproduce AOIs disappearing at all (presets or drawn) 🤔

hanbyul-here: I added this line because of a warning sentence from mapboxgl-draw - https://github.com/mapbox/mapbox-gl-draw/blob/main/docs/API.md#api-reference ( `Draw only works after the Mapbox GL JS map has loaded, so you must interact with Draw only after your map's load event:`) but I am actually not sure if we need this or even this hook is getting triggered after map is loaded (probably not - maybe as a byproduct.) I will check if this is needed again.

app/scripts/index.ts
```diff
@@ -16,9 +16,21 @@ import { PageMainContent } from '$styles/page';
...
+// Adding .last property to array
+/* eslint-disable-next-line fp/no-mutating-methods */
+Object.defineProperty(Array.prototype, 'last', {
+  enumerable: false,
+  configurable: true,
+  get: function () {
+    return this[this.length - 1];
+  },
+  set: undefined
+});
```
dzole0311, line 32: Not a blocker, but probably it's better to define it once in some kind of a `polyfills.ts` and use it / import it where needed.

dzole0311 (approved):

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
hanbyul-here: ok we looked at this together that this line doesn't likely do anything. And I tested this with the instance that it does not break anything to remove this line. I am removing this.

sandrahoang686:
Ran this locally with nextJs and `test-branch` and looks good enough to merge, I didn't get to do an in-depth code review but dont want to block this. We can look at this code as a whole in the feature branch though and be thorough before merging that feature branch 👍🏼 cc @hanbyul-here @dzole0311

PR #1231, opened 2024-10-31
Hotfix to hide the external link badge from cards

dzole0311:

### Description of Changes
This update addresses a mistake I made from the initial fix involving the new `isExternalLink` prop, which unintentionally altered the behavior of external links. Basically, this prop not only hid the "External Link" badge but also incorrectly routed external links through React's internal routing, resulting in URLs like `earth.gov/visit/https://earth.gov/mobile-climate-mapper` instead of correctly navigating to `earth.gov/mobile-climate-mapper`

**What's fixed here:**

1. Removed the `isExternalLink` prop to revert unintended internal link behavior
2. Introduced a new prop, `hideExternalLinkBadge` that more clearly describes it's role and is used specifically for **visually hiding** the "External Link" badge without altering the link's external behavior

To minimize team effort for a hotfix release and since this fix is specific to the EIC instance only, I did the following:

1. Created a branch from the latest tag specifically for this hotfix
2. Bumped the EIC instance to the latest UI version using this hotfix branch
3. Protected the hotfix branch, which will remain in place until the next planned release (in two weeks)

### Validation / Testing
1. Use the `hideExternalLinkBadge` prop to the `External Link Test` story
2. Check that the "External Link" badge is hidden from the story card
3. Check that the link is opened correctly once the card is clicked

Or check it out in [practice on the EIC instance here](https://deploy-preview-144--earth-information-center.netlify.app/visit), where the Mobile Climate Mapper card has no "External Link" badge anymore:


app/scripts/components/common/card/index.tsx
```diff
@@ -289,25 +289,23 @@ function CardComponent(props: CardComponentPropsType) {
...
       onLinkClick,
       pathAttributeKeyName: 'to',
-      LinkElement: SmartLink,
-      isLinkExternal
+      LinkElement: SmartLink
     };
   }
 
-  const isExternalLink = linkProperties.isLinkExternal ?? /^https?:\/\//.test(linkProperties.linkTo);
+  const isExternalLink = /^https?:\/\//.test(linkProperties.linkTo);
```
hanbyul-here, line 301: Ah yes!  this is what we all needed. Thanks for thinking through this again 🙇

hanbyul-here (approved):

sandrahoang686:
@dzole0311 if you created the 5.9.1 patch from here, could you update the package.json too to when it merges, it has the udpate version reflected?

PR #1240, opened 2024-11-04
Cookie consent render logic

hanbyul-here:
This pr fixes infinite rerendering issue (it was from the parsing error of cookie value - so I added a unit test to make sure this works. - I used the solution from this post, feel free to edit if you see any improvement  https://stackoverflow.com/questions/5142337/read-a-javascript-cookie-by-name)  - also feel free to edit the test, I mainly put it for my dev purpose.

I removed currenwindowurl logic, replaced it with pathname passed from the layout root. I started making this change and then realized that this is not the cause of infinite rendering problem, but I think it is probably safer to depend on the pathname change  than location.href since href can include so many things  like query parameters?

### DIFF
--- a/app/scripts/components/common/layout-root/index.tsx
+++ b/app/scripts/components/common/layout-root/index.tsx
@@ -5,7 +5,7 @@ import React, {
   useEffect,
   useState
 } from 'react';
-import { Link } from 'react-router-dom';
+import { Link, useLocation } from 'react-router-dom';
 import { useDeepCompareEffect } from 'use-deep-compare';
 import styled from 'styled-components';
 import { Outlet } from 'react-router';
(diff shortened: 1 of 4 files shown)


app/scripts/components/common/layout-root/index.tsx
```diff
@@ -56,6 +56,8 @@ function LayoutRoot(props: { children?: ReactNode }) {
   const { children } = props;
   const [sessionStart, setSesstionStart] = useState<string | undefined>();
   const sessionItem = window.sessionStorage.getItem(SESSION_KEY);
+  const { pathname } = useLocation();
```
snmln, line 59: Im a fan of using `useLocation` but want to double check that this isn't going to cause a problem when implemented in NEXT.js?

hanbyul-here: yes, thanks for checking. We are not going to export layout root for library. This is where things that are unique to instances should happen.!

snmln (approved):

PR #1241, opened 2024-11-05
Simplify logic for cookie consent

hanbyul-here:
First of all, I apologize that I could not make this suggestion as a part of review. It is still challenging for me to make a suggestion without actually touching the code. 

This PR handles several things. The first  to fix is cookieconsent form gradually disappears on the first page load. The second is not depending on session storage - mainly because I feel like it was diverging the logic flow. The third is setting up the logic flow in a bit more straight way with some comments.


snmln (approved):
I think this looks good thank you Hanbyul!

PR #1244, opened 2024-11-06
Card image/description independent of hero image/description

dzole0311:
**Related Ticket:** https://github.com/US-GHG-Center/veda-config-ghg/issues/636

### Description of Changes

1. Introduced new `cardDescription` and `cardMedia` props to enable customizable card content
2. Implemented a fallback background color for cases where no media (e.g image source) is provided

These updates are applied across various places where the card component is used (story cards) but also layer cards in the Data Catalog and the E&A Catalog to better control displayed descriptions and images.

### Validation / Testing

1. Add `cardDescription` to the 'This is the life of water' story in the MDX configuration file
2. Verify that `/stories` displays the newly added `cardDescription` on the card
3. Click on the story -> verify that the hero section retains its previous description, separate from the card description
4. Try the same with `cardMedia`
5. Also try to remove both the `cardMedia` and the `media` prop -> verify that there's a fallback background color shown 

Verify that the behavior below (copied from the issues' acceptance criteria) is applied:

1. Story and dataset administrators can define a hero image and a card image separately from one another, within the mdx file
2. Story and dataset administrators can define a hero description and a card description separately from one another, within the mdx file
3. If a story or dataset does not have any hero or card images configured, then we default to a blank space of a solid primary color (like the blue hero for Data Catalog in GHGC)
4. If a story or dataset has a hero image defined, but no card image defined, then the card image shows the hero image (like how it currently functions)

### DIFF
--- a/app/scripts/components/common/card/horizontal-info-card.tsx
+++ b/app/scripts/components/common/card/horizontal-info-card.tsx
@@ -81,14 +81,15 @@ export default function HorizontalInfoCard(props: Props) {
         <div id='description'>
           <p>{description}</p>
         </div>
-        <div id='tags'>
-          {tagLabels &&
-            tagLabels.map((label) => (
+        {tagLabels && tagLabels.length > 0 && (
+          <div id='tags'>
+            {tagLabels.map((label) => (
               <Pill variation='primary' key={label}>
                 {label}
               </Pill>
             ))}
-        </div>
+          </div>
+        )}
       </CardContent>
     </HorizontalCard>
   );
--- a/app/scripts/components/common/card/horizontal-info-card.tsx
+++ b/app/scripts/components/common/card/horizontal-info-card.tsx
@@ -1,9 +1,6 @@
 import React from 'react';
 import styled, { css } from 'styled-components';
-import {
-  glsp,
-  themeVal,
-} from '@devseed-ui/theme-provider';
+import { glsp, themeVal } from '@devseed-ui/theme-provider';
 import { CardTitle } from './styles';
 import { variableBaseType } from '$styles/variable-utils';
 import { Pill } from '$styles/pill';
@@ -17,6 +14,8 @@ const CardImage = styled.div`
   min-width: 10rem;
   width: 10rem;
   height: 100%;
+  background: ${themeVal('color.primary-100')};
+
   img {
     width: 100%;
     height: 100%;
@@ -48,8 +47,8 @@ export const HorizontalCardStyles = css`
     /* stylelint-disable-next-line value-no-vendor-prefix */
     display: -webkit-box;
     -webkit-line-clamp: 2; /* number of lines to show */
-            line-clamp: 2;
-  
+    line-clamp: 2;
+
     /* stylelint-disable-next-line property-no-vendor-prefix */
     -webkit-box-orient: vertical;
   }
@@ -70,18 +69,12 @@ interface Props {
 }
 
 export default function HorizontalInfoCard(props: Props) {
-    const {
-    title,
-    description,
-    imgSrc,
-    imgAlt,
-    tagLabels,
-  } = props;
+  const { title, description, imgSrc, imgAlt, tagLabels } = props;
 
   return (
     <HorizontalCard>
       <CardImage>
-        <img src={imgSrc} alt={imgAlt} loading='lazy' />
+        {imgSrc && <img src={imgSrc} alt={imgAlt} loading='lazy' />}
       </CardImage>
       <CardContent>
         <CardTitle>{title}</CardTitle>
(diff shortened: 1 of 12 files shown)


app/scripts/components/common/card/horizontal-info-card.tsx
```diff
@@ -70,36 +69,27 @@ interface Props {
...
-    imgAlt,
-    tagLabels,
-  } = props;
+  const { title, description, imgSrc, imgAlt, tagLabels } = props;
 
   return (
     <HorizontalCard>
       <CardImage>
-        <img src={imgSrc} alt={imgAlt} loading='lazy' />
+        {imgSrc && <img src={imgSrc} alt={imgAlt} loading='lazy' />}
```
AliceR, line 77: I am wondering if we should wrap the whole `<CardImage />` component in the conditional? Seems unnecessary to just keep an empty component, when there is no `imgSrc`

app/scripts/components/common/card/horizontal-info-card.tsx
```diff
@@ -70,36 +69,27 @@ interface Props {
...
-          {
-            tagLabels && (
-              tagLabels.map((label) => (
-                <Pill variation='primary' key={label}>
-                  {label}
-                </Pill>
-              ))
-            )
-          }
+          {tagLabels &&
```
AliceR, line 85: same here?

app/scripts/components/common/catalog/catalog-card.tsx
```diff
@@ -76,32 +87,77 @@ const CardSelectable = styled(Card)<{ checked?: boolean, selectable?: boolean }>
...
   const sources = getTaxonomy(dataset, TAXONOMY_SOURCE)?.values;
   const allTaxonomyValues = getAllTaxonomyValues(dataset).map((v) => v.name);
 
   const title = layer ? layer.name : dataset.name;
-  const description = layer ? layer.description : dataset.description;
-  const imgSrc = layer?.media?.src ?? dataset.media?.src;
-  const imgAlt = layer?.media?.alt ?? dataset.media?.alt;
+  const description = getDescription(layer, dataset);
+  const imgSrc = getMediaProperty(layer, dataset, 'src');
+  const imgAlt = getMediaProperty(layer, dataset, 'alt');
```
AliceR, line 160: Extracting into separate functions looks a lot cleaner, thanks!

app/scripts/components/home/featured-stories.tsx
```diff
@@ -87,10 +87,12 @@ function FeaturedStories() {
...
                     parentTo={STORIES_PATH}
-                    description={i === 0 ? d.description : undefined}
+                    description={
+                      i === 0 ? d.cardDescription ?? d.description : undefined
+                    }
                     date={d.pubDate ? new Date(d.pubDate) : undefined}
-                    imgSrc={d.media?.src}
-                    imgAlt={d.media?.alt}
+                    imgSrc={d.cardMedia?.src ?? d.media?.src}
+                    imgAlt={d.cardMedia?.alt ?? d.media?.alt}
```
AliceR, line 95: The logic here is a bit hard to read. Maybe we can reuse the getDescription functions above, to remove a few `??` `:` `?`

AliceR (commented):
Yay, prettier formatting 🙌  
Just leaving some thoughts I had when reading the code.

Testing the branch locally works as stated, thank you for listing the detailed verification steps!
Do you think we can consolidate the logic of when to display what in some testable functions, and actually add tests for them? Maybe unit tests, or even via playwright? This way we have a good documentation of the indented behavior as well, for future reference.

app/scripts/components/common/card/horizontal-info-card.tsx
dzole0311: Good point @AliceR. I initially wrapped the entire `<CardImage />` component but I want to avoid collapsing the layout entirely when there’s no `imgSrc`. So I kept the styled component to display a fallback blue-ish square.

app/scripts/components/home/featured-stories.tsx
dzole0311: Thanks, I now re-factored the utils which also got me thinking about the testing part, so I added a bunch of unit tests that cover all the possible cases I could think of.

app/scripts/components/common/featured-slider-section.tsx
```diff
@@ -140,9 +131,9 @@ function FeaturedSliderSection(props: FeaturedSliderSectionProps) {
                           )}
                         </CardMeta>
                       }
-                      description={cardDescription}
-                      imgSrc={cardMedia?.src}
-                      imgAlt={cardMedia?.alt}
+                      description={getDescription(d)}
+                      imgSrc={getMediaProperty(undefined, d, 'src')}
```
AliceR, line 135: I'd prefer `null` instead of `undefined` in this case, to mark it as intentional. 
We don't have to change it now, since we use `undefined` in several places throughout the code already, but maybe consider it for the future?

AliceR (approved):
Awesome! Great to have extensive tests, love it 🙌 
Tested the branch locally and run the tests, it all works perfectly.

app/scripts/components/common/featured-slider-section.tsx
dzole0311: Noted, will consider switching to `null` next time 👍

PR #1154, opened 2024-09-17
[FEATURE BRANCH] 902 E&A breakout

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/902

### Description of Changes
**Related NextJs PR/Branch:** https://github.com/developmentseed/next-veda-ui/pull/4

**GHG Preview** with this branch pulled in generated here => https://github.com/US-GHG-Center/veda-config-ghg/pull/659

**Iteration 1** - Issue https://github.com/NASA-IMPACT/veda-ui/issues/902, changes by @sandrahoang686 
* Exposed `ExplorationAndAnalysis` component
* Exposed `timelineDatasetsAtom` to be used NextJs side
* Exposed & Moved/decoupled `DatasetSelectorModal` out of `ExplorationAndAnalysis`. ExplorationAndAnalysis only needs to know the dataset layers to be displayed. DatasetSelectorModal or whatever layer selection method should be controlled at the parent level above `ExplorationAndAnalysis`.
* Commented out certain controls that were breaking `ExplorationAndAnalysis` in NextJs instance 
   * `AnalysisMessageControl`
   * `CustomAoIControl`

### Notes & Questions About Changes
_{Add additonal notes and outstanding questions here related to changes in this pull request}_

### Validation / Testing
* Validate that Datalayer Selection Modal works as expected
* Validate that E&A features work as expected - aois, analyzation, adding/removing cards
* Validate that all card linking works as expected

PR #1251, opened 2024-11-11
Various cleanup on EA breakout feature branch 

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/pull/1154

### Description of Changes
* Fixed to have changes from https://github.com/NASA-IMPACT/veda-ui/pull/1231/files# which I accidentally removed when resolving conflicts in main feature branch
* Remove `onClick` from `LinkProperties`
   * When something is a link, it doesn't need an onClick event... The onClick event should be placed on the `Card` Component instead of link props
* LinkProperties should not be required for the card component. Only when the card re-routes would it need the linkProperties. For example, cards on the dataset selector modal dont re-route and we shouldn't have to pass in linkProps
* Remove try/catch around aoi area/point selection logic and into local state with useEffect

### Note
Built `v5.9.1-ea.0` off of these changes
Update: Bulit `v5.9.1-ea.1` has been built

### Validation / Testing
* Please make sure Dataset Layer Selection modal and E&A works as expected
* Please make sure all cards in the dashboard are working as expected
   * Make sure external cards link externally and have the external badge
   * Make sure internal cards link internally correctly
   * Make sure Dataset Layer Selection modal selects cards as expected

**NextJs Preview with updated version:** https://github.com/developmentseed/next-veda-ui/pull/4

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
@@ -76,15 +76,25 @@ function CustomAoI({
   const [selectedState, setSelectedState] = useState('');
   const [presetIds, setPresetIds] = useState([]);
   const [fileUploadedIds, setFileUplaodedIds] = useState([]);
+  const [, forceUpdate] = useState(0);   // @NOTE:  Needed so that this component re-renders to when the draw selection changes from feature to point.
+  const [isAreaSelected, setAreaSelected] = useState<boolean>(false);
+  const [isPointSelected, setPointSelected] = useState<boolean>(false);
 
   const [selectedForEditing, setSelectedForEditing] = useAtom(selectedForEditingAtom);
 
   const { onUpdate, isDrawing, setIsDrawing, features } = useAois();
   const aoiDeleteAll = useSetAtom(aoiDeleteAllAtom);
 
-  // Needed so that this component re-renders to when the draw selection changes
-  // from feature to point.
-  const [, forceUpdate] = useState(0);
+  // @NOTE: map?._drawControl?.getSelected() needs access to mapboxgl draw context store,
+  // but the function gets called before mapboxdraw store is initialized (before being added to map) resulting in an error
+  useEffect(() => {
+    if (!map) return;
+
+    const mbDraw = map?._drawControl;
+    setAreaSelected(mbDraw?.getSelected().features.length);
+    setPointSelected(mbDraw?.getSelectedPoints()?.features.length)
+  }, [map])
+  
   useEffect(() => {
     const mbDraw = map?._drawControl;
     if (!mbDraw) return;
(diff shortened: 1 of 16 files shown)


sandrahoang686:
@dzole0311 I had to resolve a good amount of conflicts in the original Feature branch `902-ea-breakout` and had missed some thing which I then fixed alongside this PR. Particularly the code related to your external link hotfix (look at description). If you could check that I didn't miss anything though? Also if there is anything in the main feature branch PR that you think needs to be cleaned up more, please feel free to add to this PR with those fixes. 🙏🏼

sandrahoang686:
Not so happy with the linkProperties, they are sprinkled in many places prop drilled quite deep, also the work around to support instances since they use the card component is just not ideal, we should probably update their code to pass in linkProps.. but I think we should spend more time looking at the navigation more holistically sometime in the future 😞

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -76,15 +76,25 @@ function CustomAoI({
...
-  const [, forceUpdate] = useState(0);
+  // @NOTE: map?._drawControl?.getSelected() needs access to mapboxgl draw context store,
+  // but the function gets called before mapboxdraw store is initialized (before being added to map) resulting in an error
+  useEffect(() => {
+    if (!map) return;
+
+    const mbDraw = map?._drawControl;
+    setAreaSelected(mbDraw?.getSelected().features.length);
+    setPointSelected(mbDraw?.getSelectedPoints()?.features.length);
+  }, [map]);
```
hanbyul-here, line 96: It looks like the hook isn’t triggered after the initial load, - which makes sense since its only dependency is map.

You can check this by drawing two polygons, and select only one. the message at the bottom should say 'delete area' (not Delete All Areas) like [this page](https://earthdata.nasa.gov/dashboard/exploration?datasets=%5B%5D&taxonomy=%7B%7D&search=&selectedForEditing=true&aois=%5B%22xs%60aJcbujBypqnB%7DraC%3Fnuu%7BC%7Czv_Cgjjc%40%22%2C%2295e0ec%22%2Cfalse%2C%22fczjEeeh%7CAyn%7DnDgcqo%40q%7EzK%60rqwCb%7CbyAfsnoA%22%2C%2269e741%22%2Ctrue%5D) but the message doesn't change. (because the state didnt' change - because the hook was not triggered 💦 )

hanbyul-here (commented):
🙇 thanks for doing the extra work.

Adding my 2 cents to Link Properties: Now that I think, what we should have done is instead of `pathAttributeKeyName`, we should have just made `LinkElementProps` and make it take whatever props that the component requires since we can't really predict what kind of props that LinkElement will need. 🤔  We can def separate the issue but what do you think?

dzole0311 (commented):
1. Hiding the external badge works as on prod ✅ Found a minor, non-urgent issue which I will address in another PR
2. Tested card functionality and everything worked as expected ✅ The only change I noticed is that the cursor no longer shows as a "pointer" when hovering over E&A Catalog Cards due to the empty link props (which seems appropriate for this case)

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
```diff
@@ -249,19 +259,6 @@ function CustomAoI({
...
-  // @NOTE: map?._drawControl?.getSelected() needs access to mapboxgl draw context store,
-  // but the function gets called before mapboxdraw store is initialized (before being added to map) resulting in an error
-  // Wrapping with try block so the values get subbed even when the store is not available
-  try {
-    isAreaSelected = !!(map?._drawControl?.getSelected().features.length);
-    isPointSelected = !!(map?._drawControl?.getSelectedPoints()?.features.length);
-  } catch(e) {
-    isAreaSelected = false;
-    isPointSelected = false;
-  }
```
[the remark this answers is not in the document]
hanbyul-here: Yes - but the reason that this could not be wrapped is addressed in this comment: https://github.com/NASA-IMPACT/veda-ui/pull/1251#discussion_r1837479489

app/scripts/components/common/map/controls/aoi/custom-aoi-control.tsx
sandrahoang686: Thanks for catching 🙏🏼. Fixed.

aboydnw:
I also did some manual testing and couldn't find any issues with this pr/preview link. looks good to me ✅

PR #1252, opened 2024-11-12
[DRAFT] Pass env vars as props

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1166

Opening for reference to assess the level of prop drilling

### Description of Changes
_{Update with description of changes made in this pull request}_

### Notes & Questions About Changes
_{Add additonal notes and outstanding questions here related to changes in this pull request}_

### Validation / Testing
_{Update with info on what can be manually validated in the Deploy Preview link for example "Validate style updates to selection modal do NOT affect cards"}_

Issue #1166, opened 2024-09-23
[Refactor] Make environment variables work with the library build


hanbyul-here:
## Context

Our library build currently builds with the default environment variables (of VEDA UI) because the components are using environment variables directly. (ex.https://github.com/NASA-IMPACT/veda-ui/blob/d2bc0e55a575f3158792b0f2179807edd762b0f3/app/scripts/components/exploration/analysis-data.ts#L132) 

One of the solutions is to make these variable passed as props, so it can be passed from the components that re not exported to ui-library (this can be a lot of prop drilling though.)

## Acceptance Criteria

- [ ] Given when instance can pass the environment variales such as api endpoint from its side.

## Related Tickets

[If applicable, link any tickets that are related]

PR #1253, opened 2024-11-13
[1156] Expose a Veda UI EnvConfigProvider for providing env variables

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1166

### Problem

We need a way for future Next.js instances using veda-ui components to pass their environment variables to the library

[The previous PR](https://github.com/NASA-IMPACT/veda-ui/pull/1252) tried to pass env variables as props to our components. However, this might be difficult to maintain and to scale as new environment variables will be added.

This PR introduces a centralized config place through a new `VedauiConfigProvider` context to allow devs to configure all environment variables in one place via a config object.

### Description of Changes

**Veda-UI changes**

- Created `VedauiConfigProvider` context provider
- Currently supports these configuration options:
  - `mapboxToken` ( `MAPBOX_TOKEN`)
  - `apiRasterEndpoint` (`API_RASTER_ENDPOINT`)
  - `apiStacEndpoint` (from `API_STAC_ENDPOINT`)
- Refactored internal components to consume the environment variables from this context

**Next.js ([PR here](https://github.com/developmentseed/next-veda-ui/pull/13)):**

- Imported and used the new provider
- ℹ️ In Next.js, environment variables should be named according to the Next.js naming standards so that they can be applied e.g `MAPBOX_TOKEN` should be `NEXT_PUBLIC_MAPBOX_TOKEN` etc. 

### Validation / Testing

Test that the correct env variables are applied when using the library in Next.js

### DIFF
--- a/app/scripts/context/config-context.tsx
+++ b/app/scripts/context/config-context.tsx
@@ -1,0 +1,27 @@
+import React, { createContext } from 'react';
+
+interface VedauiConfig {
+  mapboxToken: string;
+  envApiStacEndpoint: string;
+  envApiRasterEndpoint: string;
+}
+
+export const VedauiConfigContext = createContext<VedauiConfig>({
+  mapboxToken: '',
+  envApiStacEndpoint: '',
+  envApiRasterEndpoint: ''
+});
+
+export function VedauiConfigProvider({
+  config,
+  children
+}: {
+  config: VedauiConfig;
+  children: any;
+}) {
+  return (
+    <VedauiConfigContext.Provider value={config}>
+      {children}
+    </VedauiConfigContext.Provider>
+  );
+}
--- a/app/scripts/context/config-context.tsx
+++ b/app/scripts/context/config-context.tsx
@@ -1,13 +1,13 @@
 import React, { createContext } from 'react';
 
 interface VedauiConfig {
-  mapboxToken: string;
+  envMapboxToken: string;
   envApiStacEndpoint: string;
   envApiRasterEndpoint: string;
 }
 
 export const VedauiConfigContext = createContext<VedauiConfig>({
-  mapboxToken: '',
+  envMapboxToken: '',
   envApiStacEndpoint: '',
   envApiRasterEndpoint: ''
 });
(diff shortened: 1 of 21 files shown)


hanbyul-here:
Thanks for showing both approaches @dzole0311 🙇 

Yah I agree with @sandrahoang686 . The need for something that can save the configuration-ish (I am saying -ish since I don't think all the configurations can go to providers.) data has been rising, and this might be time to do so. 

This is a bit of implementation detail that doesn't have to be tackled now, but mainly to jot down my thoughts -
1.  The other properties that I think worth going to Context is path for each page ex. e&a page needs dataset page route and vice versa - and I think it is confusing for users to pass datasetpath as a prop to e&a page. (and of course, link) 
2. I wonder if it is better to differentiate the values that come from dotenv vs. not -ex. the context  assigns the initial value from dotenv without asking the client to hydrate.

dzole0311:
@hanbyul-here @sandrahoang686 thanks for the comments!

> The other properties that I think worth going to Context is path for each page ex. e&a page needs dataset page route and vice versa - and I think it is confusing for users to pass datasetpath as a prop to e&a page. (and of course, link)

@hanbyul-here Agreed, this would definitely be confusing. But I think we should provide all route-triggering actions as callbacks within the components so that components don't have to handle routing directly (if that's feasible). 

> (and of course, link) 

For the link, did you mean differentiating between the different routers (like react-router vs next.js routing)?

> I wonder if it is better to differentiate the values that come from dotenv vs. not -ex. the context assigns the initial value from dotenv without asking the client to hydrate.

Good point! We could place dotenv vars under an env prop and have separate (e.g theming) or other props within the provider’s config object. I can address this in a follow up as we better define what else could sit in the config.

sandrahoang686:
> But I think we should provide all route-triggering actions as callbacks within the components so that components don't have to handle routing directly (if that's feasible)
Yes, very much agree

app/scripts/context/config-context.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { createContext } from 'react';
+
+interface VedauiConfig {
+  mapboxToken: string;
+  envApiStacEndpoint: string;
+  envApiRasterEndpoint: string;
```
sandrahoang686, line 6: ⛏️ : the endpoints are prefixed with `env` - maybe `envMapboxToken` to keep the standard 🤷🏼‍♀️

Issue #1238, opened 2024-11-04
[REFACTOR - Bug] timelinedataset hydration on new E&A page through URL parameter not working


hanbyul-here:
### Context
**This contents below only apply to refactored registry, not to application that use veda-ui as a submodule.** 

Timelinedataset hydration through URL parameter not working. To simply describe the issue, we are using the dataLayers from Veda faux module (already compiled, so it can be used like a dependency) to reconcile the datasets from URL parameters. : https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/atoms/datasets.ts#L54

We would need to change how we approach static datasets for this part.

#### Notes
- one of the possible solution is to make the static datasets also an atom and hydrate it so it can be used inside of other atoms? 
- I wonder if we need to consolidate how other components pass static datasets as props to standardize the dataflow once the issue above is resolved.

URL: 

Blocking a release? (y/n): 

Regression issue? (y/n):
[If yes, add Regression label]

Priority: 

### Steps to reproduce the issue

1. [Step 1 to reproduce]
2. [Step 2 to reproduce]

aboydnw:
We've had some discussions on this ticket this week, summarizing here.

The short version is: If there is a simpler solution that would make state management easier, then let's at least propose it, even if it doesn't seem to fulfill existing requirements.

For example, right now we support the ability to refresh and have your AOIs and other selections persist. We also have a really complex AOI selection mechanism. Lastly, we support the ability for somebody to copy and paste the URL so they can share their work with somebody else. 

If there was a simpler solution that means you will lose your work on refresh, maybe we just show a dialog box warning users about refreshing. It is easy to create an AOI again and run an analysis, so that would not be the worst thing in the world.

If there is a simpler solution that means you won't be able to share the URL with your colleagues, maybe we explore another way for people to share analyses. For example, maybe they can download a file that defines the AOIs they created and even other selections like layers/dates. Then, we augment the AOI upload feature to allow for the additional information. Only to say, there are other solutions we could explore.

So, I think step 1 of this ticket should be to ask ourselves the question, if we were going to start from scratch and we knew some of these requirements were flexible, would we build this the same way? Then we can move forward from there.

sandrahoang686:
[...]

PR #1266, opened 2024-11-20
[#1238] Create a derived atom for dataset hydration from Next.js

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1238
**Related Next.js PR:** https://github.com/developmentseed/next-veda-ui/pull/17

### Description of Changes
- Removed dependency on static datasets (veda_faux_module_datasets) when reconciling datasets from URL params in E&A
- Created `externalDatasetsAtom` to store datasets from host applications (eg Next.js)
- Modified `datasetLayersAtom` to use external datasets instead of the bundled static datasets from VEDA UI

### Validation / Testing
VEDA UI
1. Open E&A and select a few datasets
2. Once they are loaded, refresh the page
3. Verify that the datasets still load without issues, no errors are thrown

Next.js
1. Same as for the VEDA UI above

### DIFF
--- a/app/scripts/components/exploration/container.tsx
+++ b/app/scripts/components/exploration/container.tsx
@@ -1,12 +1,13 @@
 import React, { useState } from 'react';
 import { TourProvider } from '@reactour/tour';
 import { DevTools } from 'jotai-devtools';
-import { useAtom } from 'jotai';
+import { useAtom, useSetAtom } from 'jotai';
 import { PopoverTourComponent, TourManager } from './tour-manager';
 
 import { DatasetSelectorModal } from './components/dataset-selector-modal';
 import { allExploreDatasets } from './data-utils';
 import useTimelineDatasetAtom from './hooks/use-timeline-dataset-atom';
+import { externalDatasetsAtom } from './atoms/datasetLayers';
 import ExplorationAndAnalysis from '.';
 import { urlAtom } from '$utils/params-location-atom/url';
 import { DATASETS_PATH, EXPLORATION_PATH } from '$utils/routes';
@@ -32,6 +33,8 @@ const tourProviderStyles = {
 };
 
 export default function ExplorationAndAnalysisContainer() {
+  const setExternalDatasets = useSetAtom(externalDatasetsAtom);
+  setExternalDatasets(allExploreDatasets);
   const [timelineDatasets, setTimelineDatasets] = useTimelineDatasetAtom();
   const [datasetModalRevealed, setDatasetModalRevealed] = useState(
     !timelineDatasets.length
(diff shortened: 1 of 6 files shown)


app/scripts/components/exploration/container.tsx
```diff
@@ -32,6 +33,8 @@ const tourProviderStyles = {
 };
 
 export default function ExplorationAndAnalysisContainer() {
+  const setExternalDatasets = useSetAtom(externalDatasetsAtom);
+  setExternalDatasets(allExploreDatasets);
```
dzole0311, line 37: The current approach using `setExternalDatasets` works, but it might need to be called in multiple places where we depend on the static datasets. 

I was wondering if a more scalable solution would be to use [our existing](https://github.com/NASA-IMPACT/veda-ui/blob/14fc8ea4dabb10a0c1912230c6e71359798ca614/app/scripts/context/env-config.tsx#L9) `EnvConfigProvider` and evolve it into a more general `ConfigProvider`. Like this it would centralize both our env and data configuration in one place.

The provider could be used like this (note the extra datasets prop passed):

```
<EnvConfigProvider
      config={{
                envMapboxToken: process.env.NEXT_PUBLIC_MAPBOX_TOKEN ?? '',
                envApiStacEndpoint: process.env.NEXT_PUBLIC_API_STAC_ENDPOINT ?? '',
                envApiRasterEndpoint: process.env.NEXT_PUBLIC_API_RASTER_ENDPOINT ?? '',
                datasets: static-datasets
      }}
>
              {children}
</EnvConfigProvider>
```

And internally, the provider would handle dataset hydration:

```
const setExternalDatasets = useSetAtom(externalDatasetsAtom);
setExternalDatasets(allExploreDatasets);
```

Like that the setter won't have to be manually called several times if it's needed.

vgeorge (changes requested):
@dzole0311 I was able to do the steps on Veda UI without errors. But, on the Next.js Veda UI, I'm getting the following when trying to access the `/exploration` route:

```
Uncaught TypeError: right-hand side of 'in' should be an object, got undefined
    setAtom react.js:141
    ExplorationAnalysis exploration.tsx:38
    renderWithHooks react-dom.development.js:11121
    updateFunctionComponent react-dom.development.js:16290
    mountLazyComponent react-dom.development.js:16765
    beginWork$1 react-dom.development.js:18464
    callCallback react-dom.development.js:20565
    invokeGuardedCallbackImpl react-dom.development.js:20614
    invokeGuardedCallback react-dom.development.js:20689
    beginWork react-dom.development.js:26949
    performUnitOfWork react-dom.development.js:25748
    workLoopConcurrent react-dom.development.js:25734
    renderRootConcurrent react-dom.development.js:25690
    performConcurrentWorkOnRoot react-dom.development.js:24504
    workLoop scheduler.development.js:256
    flushWork scheduler.development.js:225
    performWorkUntilDeadline scheduler.development.js:534
```

dzole0311:
Thanks @vgeorge for testing! Did you try to rebuild and re-link the veda-ui library using this branch (`yarn buildlib`) and then recreate the alias in the Next.js branch (`yarn link @developmentseed/veda-ui`)? I saw the same error just now and I realized that I had another veda-ui library build locally. 

Also, to simplify testing, I’ve just published a new version of the veda-ui library with these changes and updated the dependency in the [Next.js PR](https://github.com/developmentseed/next-veda-ui/pull/17). So reinstalling the node packages should hopefully resolve the issue on your end. It appears to work fine in the deployment preview: [Deployment Preview](https://deploy-preview-17--veda-ui-next-test.netlify.app/). Let me know if the issue persists!

vgeorge (approved):
Thanks for your support, @dzole0311. The issue was in my local env for the Next.js UI instance. I followed the steps and it is working as described.

app/scripts/components/exploration/container.tsx
hanbyul-here: How you described makes sense. My concern for now is how nicely it would work out irl considering Jotai places itself as a replacement for context (I am reading https://jotai.org/docs/basics/comparison) - or it just means that we are really not taking full advantage of Jotai. I will put more thoughts into this from the perspective of global state management.

PR #1269, opened 2024-11-21
ci: automate release

### LINKED ISSUE #1236: CI/CD for releasing new version and updating instances
### Context

We release to veda-ui at least every 2 weeks, sometimes more frequent for hot fixes or special occasions. This can take a considerable amount of time for the team, that could be better spent working on the roadmap. We would like to minimize the time spent releasing and testing, without sacrificing quality or reliability. 

Also to consider is releasing to instances. Currently, we allow each instance to manage their own release. For instances like Earthdata and earth.gov, though, we are often the responsible party. So there is incentive to automate the process for instances to release new content to production and intake new versions of veda-ui.

Some previous thoughts were gathered in [this doc](https://docs.google.com/document/d/1MYwF2chy8LcRTRyja---x7sIs4Wod44jXFPYLUQbta4/edit?tab=t.0#heading=h.hpihcughr9w)

### Acceptance Criteria

- [ ] A release is automatically generated every 2 weeks, on Mondays
- [ ] A release can be triggered manually by people on the core team
- [ ] Automatically in the releases process, changelogs are generated, git tags are created, and package.json is updated with the tagged version

hanbyul-here:
Close #1236 

### Description of Changes
Add packages, and configuration to make automated release work
Add an action to release a package every Monday 
Add an action to check PR title to follow conventional commit pattern

If you would like to peek at how our release will look like with automation, I experimented it in this repo. Heads-up that there are many empty releases because I was experimenting with scheduling (cron) : https://github.com/hanbyul-here/release-test/releases/

Heads-up that this doesn't do scheduled releases yet. It will be not a big lift to add it, but I wanted us to trigger the action through GitHub Action page first and then kick the scheduled releases out. Currentlh the only way to trigger this action is through UI from action tab. 

Screenshot 2024-12-02 at 10 55 45 PM

#### Conventional commit format
For conventional commits, we have a couple of options for adoption:

1. Require PR title to follow the conventional commit format.
2. Require every commit message to follow the conventional commit format.

I think either approach can work as long as we maintain consistency. Personally, I find option 2 a bit challenging since features or fixes often involve multiple commits, making conventional commit prefixes redundant information. 

Currently, this PR only enforces that the PR title follows the conventional commit format. However, this requires some changes to the repository setup. Specifically:

1. We need to modify the merge commit message to use the PR title instead of GitHub's default automated message (Merged...). - We can do this through GitHub repo settings.
2. Optionally, we could set "squash and merge" as the default merge method. This would make sure that each PR results in a single commit on the main branch, creating a cleaner commit history. (do we really need all those "WIP" commits on the main branch?)
That said, I’m slightly hesitant about squash-and-merge because it makes it harder to trace individual commits, even though we can still use reflog to recover them if necessary.

#### Using conventional commit as a productivity tool

The action that I am using actually offers a lot of functionality, such as ticket number validation and labeling. I adopted the most minimum possible settings, but maybe we can take more advantage of it? @aboydnw - This is the package I am using and there are detailed writings about what it can do on its read me page: https://github.com/ytanikin/PRConventionalCommits would you find anything useful ? 

### Validation / Testing

I'd like to go over this PR together. Some items to check together
- How deploy key is tied to an individual (If I need to offboard from VEDA, deploy key would need to be refreshed)
- How to trigger release manually
- Change GitHub PR default setting (to use the title of PR instead of creating an additional message Merged...)


AliceR:
Thank you for your efforts to improve our release process @hanbyul-here !

I personally agree that prefixing every commit might be excessive, especially when writing smaller feature commits. I’d lean towards the approach of requiring the PR title to follow the conventional commit format (`1. Require PR title to follow the conventional commit format`).

Regarding squashing, I think it’s best to clean up the WIP/fix/comment commits locally before opening a PR or marking it as ready for review. I usually handle this with an interactive rebase, allowing me to merge, rewrite, or refine commit messages and descriptions as needed. That way, we can keep the history clean but still complete in the main branch, which can sometimes be useful for tracing back a thought or decision.

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          # @TODO: Add deploy key pair
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
```
hanbyul-here, line 21: Noting that this was used to bypass the branch protection rule. But I would like to replace it to use GitHub App in the future. (Specifically, when this actions starts triggering the instance build.)

yarn.lock
```diff
@@ -17,6 +17,15 @@
...
+  resolved "http://verdaccio.ds.io:4873/@babel%2fcode-frame/-/code-frame-7.26.2.tgz#4b5fab97d33338eff916235055f0ebc21e573a85"
+  integrity sha512-RJlIHRueQgwWitWgF8OdFYGZX328Ax5BCemNGlqHfplnRT9ESi8JkFlvaVYbS+UubVY6dpv87Fs2u5M29iNFVQ==
+  dependencies:
+    "@babel/helper-validator-identifier" "^7.25.9"
+    js-tokens "^4.0.0"
+    picocolors "^1.0.0"
+
 "@babel/compat-data@^7.13.11", "@babel/compat-data@^7.18.6", "@babel/compat-data@^7.18.8":
   version "7.18.8"
   resolved "http://verdaccio.ds.io:4873/@babel%2fcompat-data/-/compat-data-7.18.8.tgz#2483f565faca607b8535590e84e7de323f27764d"
```
sandrahoang686, line 31: did you mean to push this up?

.github/workflows/conventional-commit.yml
```diff
@@ -0,0 +1,15 @@
...
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  validate-pr-title:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Conventional Commit Validation
+        uses:  ytanikin/pr-conventional-commits@1.4.0
+        with:
+          task_types: '["feat","fix", "docs", "test", "ci", "refactor", "chore", "revert"]'
```
sandrahoang686, line 14: ❓ : Would general updates like to wording count as a `chore` or `fix`?

sandrahoang686:
Hey Hanbyul, great stuff! I agree with the sentiment of having the PR title following format instead of individual commits. I dont think we have to do the squash and merge though, it'd be nice to preserve history of each.

yarn.lock
hanbyul-here: Yes - (these are probably dependencies of `release-it` , the package I added. Do you see any problem?

.github/workflows/conventional-commit.yml
hanbyul-here: I think it depends on what general updates do - if it is a fix for something I would prefix it with`fix:`.  This cheatsheet might be helpful: https://gist.github.com/qoomon/5dfcdf8eec66a051ecd85625518cfd13#types

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: git config
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+      - run: yarn
```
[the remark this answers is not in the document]
hanbyul-here: Yah good point, I think it works with whatever default Node that GitHub action has right now. (at least from my test) but probably better to be explicit.

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
+name: Release Every Other Monday
+on:
+  workflow_dispatch:
+  # @TODO: byweekly check
+  # Run action at 16:15 PM on Monday (UTC)
+  # schedule:
+  #   - cron: '15 16 * * 1'
```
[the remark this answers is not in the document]
hanbyul-here: Yah, I left it as a @TODO but it might have not been clear, (hopefully) made a more clear comment.

hanbyul-here:
I will merge this so we can see if we can use this action for the next week's release. 
I also enabled Squash and Merge option - I just wanted to enable it and it seems that Squash and merge becomes the default when it is enabled - if you see any problem from this set-up, please let me know!

Issue #1108, opened 2024-08-16
[VEDA2 Refactor] Establish Pattern and Remove Routing Dependencies from Data Catalog Feature


sandrahoang686:

hanbyul-here:
From Data Catalog implementation try, we could find the routing problems that need to be handled.

### Base finding: react-router-dom component doesn't work with Next instance, and Next has its own Link component.

1. `<NavLink>` of `react-router-dom` embedded in source code
  One of the possible solutions is to Pass Link Components from Consumer level
2. Components using  react-router for their behaviors (ex. sticky header using `useLocation` of `react-router-dom` to detect page transition )
  We should first identify these components and approach them holistically. ex. Do we need this behavior when it is not used in SPA environment? 
3. Query parameter manager needing `react-router-dom` (`useQSState hook` requires `navigate` from `react-router-dom`'s `useNavigate` as a parameter)
  One of the possible solutions is to make this as a prop

4. Make sure the link works with routings defined on the instance level. (ex. does dataset catalog card work with different DATACATALOG_PATH ?)

sandrahoang686:
_Update_: This has been kindof been worked on when during breakout components by having the Link Component pass down as a prop. But there are some hacks in there that need to be clean up and we should relook at how we are doing this to see if it is the most optimal and clean... 

Going to think about what we should do with this ticket...

vgeorge:
I will have limited availability in the next few weeks and wanted to provide a status update in case someone else has time to tackle the remaining work.

In #1270 and #1276, we reduced the dependency on React Router. I believe the next steps are:

1. **Remove usage of the `useLocation` hook**:  
   This should be straightforward since most occurrences involve reading `location.pathname`. These can be replaced with the new custom `usePathname` hook.

2. **Refactor usage of linkProperties**:  
   I suggest focusing on making the components independent of routing and navigation logic, as proposed by @sandra. Adding a context provider might complicate usage in Next.js, so I'd avoid it.

   I'd recommend addressing this interactively at the leaf component level to prevent unintended breaks in different parts of the app. For example, the Catalog Card is a good starting point. We could introduce an `href` prop that, when provided, makes the card act as a link and ignores `onClick` events. We would still need a `Link` element to support both Next.js and SPAs. Below is a sample implementation of a `CustomLink` component:

```tsx
   import React from 'react';

  // try loading dependencies during build time
   let NextLink: any;
   try {
     NextLink = require('next/link').default;
[...]

PR #1270, opened 2024-11-21
Remove React Router's useNavigate dependency

vgeorge:
**Related Ticket:** Contributes to #1108

This aims to ensure catalog filters work correctly in Next.js.

### Description of Changes

Removes the usage of React Router's `useNavigate` hook in the library to make it more agnostic of routing frameworks, allowing it to work better with Next.js instances.

Changes included:

- Added a custom `commit` function to `useQsStateCreator`. It appears this was the only place where the `useNavigate` function was effectively in use. I tried using the default `commit` function from the [qs-state-hook](https://github.com/danielfdsilva/qs-state-hook) library, but it doesn't work with Next.js
- Formatted the touched files

### Validation / Testing

Using `veda-config`:

1. Visit the data catalog page
2. Click on filters, and URL params are updated as usual
3. Reload the page with filters selected; the filter panel is populated as expected
4. Repeat the steps on the Exploration and Stories pages

Using `next-veda-ui` instance:

1. Link the source of this PR to the Next.js instance using the instructions in `veda-ui/docs/development/REGISTRY.md`
2. Replace the file `app/(datasets)/data-catalog/catalog.tsx` (this avoids build errors caused by type issues)

```js
'use client';
import React from 'react';
import { CatalogView, useFiltersWithQS } from '@lib';
import { usePathname } from 'next/navigation';
import Link from 'next/link';

export default function Catalog({ datasets }: { datasets: any }) {
  const pathname = usePathname();
  const controlVars = useFiltersWithQS();

  return (
    <CatalogView
      datasets={datasets}
      onFilterChanges={() => controlVars}
      pathname={pathname}
      linkProperties={{
        LinkElement: Link,
        pathAttributeKeyName: 'href',
      }}
    />
  );
}
```

3. Access `http://localhost:3000/data-catalog`. Filter changes are reflected in the URL, and reloading populates the filter panel as expected

This is ready for review.

### DIFF
--- a/app/scripts/components/common/catalog/controls/hooks/use-filters-with-query.ts
+++ b/app/scripts/components/common/catalog/controls/hooks/use-filters-with-query.ts
@@ -32,14 +32,8 @@ export function useFiltersWithURLAtom(): UseFiltersWithQueryResult {
   };
 }
 
-export function useFiltersWithQS({
-  navigate
-}: {
-  navigate: any;
-}): UseFiltersWithQueryResult {
-  const useQsState = useQsStateCreator({
-    commit: navigate
-  });
+export function useFiltersWithQS(): UseFiltersWithQueryResult {
+  const useQsState = useQsStateCreator();
 
   const [search, setSearch] = useQsState.memo(
     {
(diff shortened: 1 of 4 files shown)


dzole0311 (approved):
Looks good to me @vgeorge 

Did some (smoke) tests using the standalone veda ui:

1. Filters + search query work and are persisted as before in the Data Catalog 
2. Clearing them all works
3. Smoke tested the routing via clicking on the "Explore Data" works as before (opens the E&A with the applied filter)
4. Smoke tested the catalog in the E&A
5. Hard reloading works as before

Same for the next.js instance (except the step 3 above)

PR #1273, opened 2024-11-22
Fix exploration types file name and imports

sandrahoang686:
Not sure how this happened but heres a fix


app/scripts/components/exploration/components/timeline/timeline.tsx
```diff
@@ -63,7 +63,7 @@ import {
   DatasetStatus,
   TimelineDatasetSuccess,
   ZoomTransformPlain
-} from '$components/exploration/types.d.ts';
+} from '$components/exploration/types.d';
```
snmln, line 66: @sandrahoang686 do these need to be changed to `types.d.ts?`

snmln: Looking at the file name change in this pr it appears `app/scripts/components/exploration/types.d.ts` to be accurately changes to the previous name.

hanbyul-here:
I am a bit unclear what this PR is trying to address - ts-check is not passing since some files are still using `types.d.ts` as an import path. Can you clarify?

PR #1276, opened 2024-11-26
refactor(catalog): introduce usePathname hook, remove React Router dependency from the Catalog

vgeorge:
**Related Ticket:** Contributes to https://github.com/NASA-IMPACT/veda-ui/issues/1108

### Description of Changes

- Remove `react-router` dependency from the Catalog components
- Introduce custom `usePathname` hook
- Remove prop drilling of pathname from the Catalog

The `pathname` prop is passed down for two things:

1. Generating a link to the dataset at https://github.com/NASA-IMPACT/veda-ui/blob/820730fd7c1e25e68997b200d31bc4ee27525f7e/app/scripts/components/common/catalog/catalog-card.tsx#L126
2. Controlling scroll in the filters control at https://github.com/NASA-IMPACT/veda-ui/blob/820730fd7c1e25e68997b200d31bc4ee27525f7e/app/scripts/components/common/catalog/filters-control.tsx#L57

I suspect that the usage on item 1 was intented to cover the use case of an instance running on a subpath (e.g. `https://myinstance.org/veda`), why is hard to replicate locally but this change shouldn't affect that.

### Validation / Testing

Using `veda-config`:

1. Visit the data catalog page
1. Clicking on cards should open the proper URL (same as before)
1. Visit Explorer page, clicking on cards selected them, instead of visiting an URL (same as before)

Using `next-veda-ui` instance:

1. Link the source of this PR to the Next.js instance 
1. Visit the catalog page
1. Clicking on cards should open the proper URL  (same as before)
1. Visit Explorer page, clicking on cards selected them, instead of visiting an URL (same as before) 

This is ready for review.

### DIFF
--- a/app/scripts/utils/use-pathname.ts
+++ b/app/scripts/utils/use-pathname.ts
@@ -1,0 +1,59 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * usePathname
+ * *
+ * This hook is implemented to work in both client-side rendering
+ * and server-side rendering environments. During SSR, it initializes the
+ * `pathname` as an empty string, ensuring the application remains stable in
+ * non-browser environments.
+ *
+ * @returns {string} The current `pathname`. Returns an empty string during SSR
+ * or if the `window` object is unavailable.
+ */
+export const usePathname = () => {
+  const [pathname, setPathname] = useState(
+    typeof window !== 'undefined' ? window.location.pathname : ''
+  );
+
+  useEffect(() => {
+    if (typeof window === 'undefined') return;
+
+    const updatePathname = () => {
+      setPathname(window.location.pathname);
+    };
+
+    // Listen to popstate events (back/forward navigation)
+    window.addEventListener('popstate', updatePathname);
+
+    // Detect programmatic navigation by dispatching a custom event
+    const originalPushState = history.pushState;
+    const originalReplaceState = history.replaceState;
+
+    const customEvent = new Event('pathnamechange');
+    const dispatchPathnameChange = () => {
+      window.dispatchEvent(customEvent);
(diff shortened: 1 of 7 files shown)


vgeorge:
The `ts-check` failure made me realized this is also used in the datasets modal in the Exploration page. I added a fix and udpated the validation steps above.

app/scripts/utils/use-pathname.ts
```diff
@@ -0,0 +1,59 @@
...
+ * *
+ * This hook is implemented to work in both client-side rendering
+ * and server-side rendering environments. During SSR, it initializes the
+ * `pathname` as an empty string, ensuring the application remains stable in
+ * non-browser environments.
+ *
+ * @returns {string} The current `pathname`. Returns an empty string during SSR
+ * or if the `window` object is unavailable.
+ */
+export const usePathname = () => {
```
dzole0311, line 14: 👍

dzole0311 (approved):
I tested the changes in both the Veda UI and the Next.js repository as well as on the [GHG Center side](https://github.com/US-GHG-Center/veda-config-ghg/pull/680). Everything works as expected with no regressions.

dzole0311:
> I suspect that the usage on item 1 was intented to cover the use case of an instance running on a subpath (e.g. https://myinstance.org/veda), why is hard to replicate locally but this change shouldn't affect that.

Yeah, that seems likely. For example the GHG Center instance is running under `https://earth.gov/ghgcenter/` subpath and the Veda routes are appended to it.

vgeorge:
@dzole0311 thanks for the review! Let's keep an eye on that instance once we roll out the changes.

Issue #1135, opened 2024-08-29
[Refactor] Layout Components - Footer


aboydnw:
### Context

This ticket is for a layout component as part of the VEDA refactor. This components are used to generate the overall layout and site architecture of the Dashboard. They match the overall look and feel of the app and will be used in the template instance. Our recommendation will be to use these layout components, and we anticipate that most instances will use these components, but some may decide to bring their own.

Our footer implementation on [GHG](https://earth.gov/ghgcenter) is probably the most recent and complete implementation so far. This also depends on the design ticket #1123

This ticket might be broken down into smaller tickets after initial investigation.

**Notes:**
* Mark new files for VEDA2 refactor with `@VEDA2-REFACTOR-WORK`

**Acceptance Criteria:**
**General**
- [ ] Reusability => modularize stateless/presentational features to be as reuseable as possible within the core feature view containers
- [ ] Composite Feature should be entirely data agnostic => Separate out the data logic from this feature component into the data layer. _Note: Data will now have to be passed in as props so we will need to update the page views to supply this_
   - [ ] Composite Features should have stricter typed interfaces
- [ ] This feature should still support current instances and be backwards compatible

**Specific**
_{TO BE UPDATED}_

**Related to:** https://github.com/NASA-IMPACT/veda-ui/issues/801

**Resources:**
* ADR: https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/adr/002-application-architecture-for-configurability.md#decision-outcome
* [Miro Board outline Core Features](https://miro.com/app/board/uXjVNlIZVJI=/)

snmln:
@aboydnw & @faustoperez I wanted to confirm are these the design that we should be working off of? 

**Design:** https://www.figma.com/design/PG7AhRloW6qJd9U00VdQ77/VEDA-Template-Instance?node-id=178-2&t=HAlyBRbfjH7jZgHy-1

aboydnw:
Yes sorry I should have update this ticket with that most recent link. I believe most recent designs for the footer happened during this ticket: https://github.com/NASA-IMPACT/veda-ui/issues/1193
That has the same figma link that you shared @snmln 

@faustoperez correct me if that's wrong or if there is a better figma link

faustoperez:
Please use the VEDA - USWDS Figma file for components, the one you list is the Template instance file

[...]

PR #1247, opened 2024-11-07
1137 implement new ds page header

sandrahoang686:
**Related Ticket:** _{link related ticket here}_

### Description of Changes
_{Update with description of changes made in this pull request}_

### Notes & Questions About Changes
_{Add additonal notes and outstanding questions here related to changes in this pull request}_

### Validation / Testing
_{Update with info on what can be manually validated in the Deploy Preview link for example "Validate style updates to selection modal do NOT affect cards"}_

PR #1285, opened 2024-12-03
feat: Implement page footer with USWDS components

snmln:
**Related Ticket:** #1135

### Description of Changes
TO DO: 

- [x] Create slim footer front-end, using USWDS.
- [x] Pulling USWDS Styles.
- [x] Connect to veda.config.js data.
- [x] Setup USWDS feature flag.
- [x] hideFooter functionality (props)
- [x] Fix Nasa svg file to not rely on css translate. 
- [ ] ~~Checkin with Fausto on progress.~~
- [x] Import into and use in next-veda-ui (https://github.com/developmentseed/next-veda-ui/pull/26)
- [ ] Connect to instance theme styling (font, color, etc.)
- [x] Ensure styling looks as expected on next-veda-ui
- [x] Add Feature Flag documentation 

### Validation / Testing


.env
```diff
@@ -14,5 +14,5 @@ GOOGLE_FORM = 'https://docs.google.com/forms/d/e/1FAIpQLSfGcd3FDsM3kQIOVKjzdPn4f
 
 SHOW_CONFIGURABLE_COLOR_MAP = 'TRUE'
 
-# Enables the refactor page header component that uses the USWDS design system
-ENABLE_USWDS_PAGE_HEADER = 'TRUE'
\ No newline at end of file
+ENABLE_USWDS_PAGE_HEADER = 'TRUE'
```
AliceR, line 17:
```suggestion
# Enables the refactor page header component that uses the USWDS design system
ENABLE_USWDS_PAGE_HEADER = 'TRUE'
# Enables the refactor page footer component that uses the USWDS design system
```

app/scripts/components/common/page-footer/styles.scss
```diff
@@ -0,0 +1,43 @@
...
+  align-self: center;
+  display: flex;
+  font-size: themeVars.$veda-uswds-fontsize-sm;
+  font-weight: themeVars.$veda-uswds-fontweight-regular;
+  gap: themeVars.$veda-uswds-spacing-105;
+}
+
+#return-to-top-container {
+  padding: themeVars.$veda-uswds-padding-4 themeVars.$veda-uswds-padding-5;
+  max-width: 64rem;
```
AliceR, line 14: TODO: replace magic number with theme var

app/scripts/components/common/page-footer/styles.scss
```diff
@@ -0,0 +1,43 @@
+@use '$styles/veda-ui-theme-vars.scss' as themeVars;
+
+.footer-text {
+  color: white;
```
AliceR, line 4: TODO: replace color with theme color

app/scripts/components/common/page-header/logo-container/styles.scss
```diff
@@ -4,6 +4,7 @@
   display: flex;
   font-family: themeVars.$veda-uswds-basefont-sans;
   gap: themeVars.$veda-uswds-spacing-105;
+  max-height: 2.5rem;
```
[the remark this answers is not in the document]
snmln: Yes, this is probably the best way for us to go about this without creating a dedicated footer & header version of the svg.

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+            </div>
+          </div>
+        }
+        secondary={
+          <div id='footer_secondary_container' className='grid-row'>
+            <div id='logo-container'>
+              <a id='logo-container-link' href='#'>
+                {logoSvg as JSX.Element}
+                <span className='footer-text'>
+                  NASA EarthData 2024 • v0.17.0
```
AliceR, line 118: - [ ] get text from config

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+                  {/* {version} */}
+                </span>
+              </a>
+            </div>
+            <div className='grid-col-4 footer-text grid-gap-6 flex-justify-end'>
+              <span>NASA Official: </span>
```
AliceR, line 124: - [ ] get text from config

app/scripts/components/common/page-header/nav/nav-item-cta.tsx
```diff
@@ -5,23 +5,29 @@ import { ActionNavItem } from '../types';
...
+              customClasses && customClasses != ''
+                ? customClasses
+                : 'usa-nav__link'
+            }
             type='button'
             tabIndex={0}
             id={item.id}
             onClick={show}
-            style={{background: 'none', border: 'none', cursor: 'pointer'}}
+            style={{ background: 'none', border: 'none', cursor: 'pointer' }}
```
AliceR, line 30: what was the reason for using inline styles?

sandrahoang686: This existed previously with pageheader PR - (commit [here](https://github.com/NASA-IMPACT/veda-ui/pull/1247/commits/7652f0b129c7e6e7b7521d66ac491d35e8624342) to address this comment [here](https://github.com/NASA-IMPACT/veda-ui/pull/1247#discussion_r1869149971)). So removes button default styling but feel free to move it elsewhere 👍🏼 .

AliceR:
image

Style problems detected over in next-veda-ui (https://github.com/developmentseed/next-veda-ui/pull/26):

- [ ] `.usa-footer__contact-info {  line-height: 1.2;  }` applied to only second item in USWDSAddress (secondaryItems). Why?
- [ ] Footer does not stick to the bottom of the page
- [ ] Background colors not applied
- [ ] Primary nav items should have an underline
- [ ] ...

snmln:
@AliceR I have corrected the issues on the nextjs instance. Heads up on one of the bullet points:  

> Footer does not stick to the bottom of the page
This is an issue with the content section not respecting the `flex-grow: 1` attribute. I tested it out appearance wise and we have everything dialed in. I will leave this as a call out for another ticket though since it is outside of the footer work we have completed.

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+//TODO: clean up PageFooterProps, Unexpected any. Specify a different interface.
+
+export default function PageFooter({
+  settings,
+  mainNavItems,
+  subNavItems,
+  hideFooter,
+  logoSvg
+}: PageFooterProps) {
+  const returnToTopButton = () => {
```
dzole0311, line 30: Move this out into a separate component `ReturnToTopButton`?

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+      >
+        <a className='usa-link text-primary' href='#'>
+          Return to top
+        </a>
+      </div>
+    );
+  };
+
+  const { returnToTop, secondarySection } = settings;
+  /* eslint-disable */
```
dzole0311, line 44: Probably not needed, we could make `cleanedNavItems` const?

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+        <div id='footer_secondary_container' className='grid-row'>
+          <div id='logo-container'>
+            <a id='logo-container-link' href='#'>
+              {logoSvg as JSX.Element}
+              <span className='footer-text'>
+                {secondarySection.division} • {secondarySection.version}
+              </span>
+            </a>
+          </div>
+          <div className='grid-col-4 footer-text grid-gap-6 flex-justify-end'>
```
dzole0311, line 123: Great job using the util classes throughout the PR 👍

app/scripts/components/common/page-footer/styles.scss
```diff
@@ -0,0 +1,46 @@
+@use '$styles/veda-ui-theme-vars.scss' as themeVars;
```
dzole0311, line 1: Due to the recent changes in https://github.com/NASA-IMPACT/veda-ui/pull/1337, we need to rename the component stylesheet files bit differently. This one would be `page-footer.scss` and can be imported in https://github.com/NASA-IMPACT/veda-ui/blob/283ae3b84e8005b354d826c53518912865650f10/app/scripts/styles/veda-components.scss so that these custom VEDA styles are automatically applied in the Next.js instance, after the USWDS stylings.

docs/content/CONFIGURATION.md
```diff
@@ -109,6 +109,33 @@ theme: {
...
+
+| Option | Type | Description| Example|
+|---|---|---|---|
+| returnToTop | boolean | Set to display a return to top button above the primary and secondary secionts of the footer ||
+| secondarySection.division | String | Pass a string for the content you would like to display in the division section  | NASA EarthData 2024 |
+| secondarySection.version | String | Set what version to display next to Nasa logo and title | BETA VERSION |
+| secondarySection.title | String |  | NASA Official |
+| secondarySection.name | String | Name of individual you would like users to contact | NASA EarthData 2024 |
+| secondarySection.to | String | Individuals email contact information to link to  | NASA EarthData 2024 |
+| secondarySection.type | String | Currently hard coded to accept only email type | NASA EarthData 2024 |
```
dzole0311, line 138: Thanks for documenting!

dzole0311 (changes requested):

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  hideFooter,
+  logoSvg
+}: PageFooterProps) {
+  const returnToTopButton = () => {
+    return (
+      <div
+        id='return-to-top-container'
+        className='margin-left-auto margin-right-auto'
+      >
+        <a className='usa-link text-primary' href='#'>
```
dzole0311, line 36: In this context, it would be more semantically correct to use an [unstyled button](https://designsystem.digital.gov/components/button/) rather than anchor tag. It also prevents the '#' fragment from being added to the url on click.

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -105,7 +112,17 @@ function LayoutRoot(props: { children?: ReactNode }) {
...
       </PageBody>
-      <PageFooter isHidden={hideFooter} />
+      {isUSWDSEnabled ? (
+        <PageFooter
+          settings={footerSettings}
+          mainNavItems={mainNavItems}
+          subNavItems={subNavItems}
+          hideFooter={hideFooter}
+          logoSvg={<NasaLogoColor />}
+        />
```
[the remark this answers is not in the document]
sandrahoang686: We will want to hook this up later with the [ConfigProvider introduced in this PR](https://github.com/NASA-IMPACT/veda-ui/pull/1338) but you dont have to worry about that right now and can just follow pageheader.

app/scripts/components/common/page-footer/index.tsx
snmln: We can do that for more clear organization but this button will most likely only be applicable to the footer and likely wont be reused outside of the footer.

app/scripts/components/common/page-header/default-config.ts
```diff
@@ -64,6 +69,18 @@ let defaultSubNavItems: (
...
+const defaultFooterSettings = {
+  secondarySection: {
+    division: 'NASA EarthData 2024',
+    version: 'BETA VERSION',
+    title: 'NASA Official',
+    name: 'Manil Maskey',
+    to: 'test@example.com',
+    type: 'email'
+  },
+  returnToTop: true
```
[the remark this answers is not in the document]
snmln: Good callout!

sandrahoang686 (approved):
In the spirit of not wanting to drag this out any longer, this looks good to me. Related to dropdowns i've just updated this https://github.com/NASA-IMPACT/veda-ui/issues/1323 to address it since its not a blocker. There are issues with lint and test in the built which will need to be fixed before merging otherwise 👍🏼 and great work!

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,141 @@
...
+  USWDSFooter,
+  USWDSFooterNav,
+  USWDSAddress
+} from '$components/common/uswds';
+import './styles.scss';
+
+interface PageFooterProps {
+  //use of NavItem is causing issues with TS and throwing erros in the .
+  mainNavItems: (NavLinkItem | DropdownNavLink | ActionNavItem)[];
+  subNavItems: (NavLinkItem | DropdownNavLink | ActionNavItem)[];
```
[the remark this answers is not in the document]
sandrahoang686: I think it makes more sense to be explicit hence why we have these config driven nav/menus and why there were created but thats fine. I'll add a note for this to be re-looked at in #1323 .

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -105,7 +114,18 @@ function LayoutRoot(props: { children?: ReactNode }) {
           />
         )}
       </PageBody>
-      <PageFooter isHidden={hideFooter} />
+      {isUswdsFooterEnabled ? (
+        <PageFooter
+          settings={footerSettings}
+          mainNavItems={mainNavItems}
+          subNavItems={subNavItems}
+          hideFooter={hideFooter}
```
sandrahoang686, line 122: curious, whats the point of having the `hideFooter` prop?

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { useMemo } from 'react';
+import { Icon } from '@trussworks/react-uswds';
+import { DropdownNavLink, FooterSettings, NavLinkItem } from 'veda';
```
sandrahoang686, line 3: Sorry.. just catching this now but we need to be decoupling from from any parcel/virtual modules/(parcel-resolver-veda) dependency for the new architecture.

Its fine for legacy (because it still depends on parcel) to reference these types from here but for new components for refactor - we should be referencing from this [file](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/page-header/types.ts). Feel free to move those types to a more common directory to be shared between Footer and Header. ([Header for example](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/page-header/nav/nav-item-links.tsx#L2))

sandrahoang686 (changes requested):

app/scripts/components/common/page-footer/index.tsx
sandrahoang686: FYI to help out, there has been an ongoing effort since the very beginning of when this refactor work started and is still going on (see [current PR](https://github.com/NASA-IMPACT/veda-ui/pull/1367/files)) to remove these veda virtrual modules/parcel dependencies for new architecture.

app/scripts/components/common/layout-root/index.tsx
snmln: Its a carry over functionality from the old footer. No real documentation to back up its use. But I would infer that certain instances may not want to show this veda footer for some reason.

app/scripts/components/common/page-footer/index.tsx
snmln: Sounds good, I updated this import to go from the common `common/types.d.ts` level since these types are leveraged in multiple areas.

app/scripts/components/common/layout-root/index.tsx
```diff
@@ -64,7 +71,8 @@ function LayoutRoot(props: { children?: ReactNode }) {
   useEffect(() => {
     // When there is no cookie consent form set up
     !cookieConsentContent && setGoogleTagManager();
-  }, []);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []); // Empty dependency array ensures this effect runs only once, and not during SSR
```
hanbyul-here, line 75: Isn't `layout-root` only used for the legacy way?

app/scripts/components/common/layout-root/index.tsx
hanbyul-here: I think this is for the case that the instance wants to have their own footer. ex. ghg has their own footer https://github.com/US-GHG-Center/veda-config-ghg/blob/bf3e66b7299d474dae01576e850bccaaaca5856c/overrides/components/page-footer/component.tsx#L136

app/scripts/components/common/page-footer/return-to-top-button/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { USWDSButton } from '$components/common/uswds';
+
+export default function ReturnToTopButton() {
+  return (
+    <div
+      id='return-to-top-container'
+      className='margin-left-auto margin-right-auto'
+    >
+      <USWDSButton
```
hanbyul-here, line 10: This button doesn't move the focus of keyboard back to the top.

app/scripts/styles/_uswds-theme.scss
```diff
@@ -14,11 +14,13 @@
     ),
   $theme-type-scale-md: 8,
   $theme-utility-breakpoints: (
-    'mobile-lg': false,
-    'desktop': false
+
+    'mobile-lg': true,
```
hanbyul-here, line 18: Did we figure out if this change is needed?

app/scripts/components/common/page-footer/page-footer.scss
```diff
@@ -0,0 +1,46 @@
...
+.footer-text {
+  color: themeVars.$veda-uswds-color-base-white;
+  align-self: center;
+  display: flex;
+  font-size: themeVars.$veda-uswds-fontsize-sm;
+  font-weight: themeVars.$veda-uswds-fontweight-regular;
+  gap: themeVars.$veda-uswds-spacing-105;
+}
+
+#return-to-top-container {
```
hanbyul-here, line 12: I am seeing some styles using IDs, Was the specificity of ID needed for the style?
(Also, can some of these styles can be replaced with utility classes?)

app/scripts/components/common/layout-root/index.tsx
sandrahoang686: but then why have hideFooter prop when instances already have control with enabling footer through env? in Hanbyul's explanation it makes sense because Footer was built into layout component (veda-ui) where remember for legacy code, instances have to do overrides and didn't have the flexibility to compose. Which is no longer the case for new architecture. We can remove the `hideFooter` prop then.

app/scripts/components/common/layout-root/index.tsx
snmln: I do not believe so. Page-header was implemented using layout-root as well.

app/scripts/components/common/page-footer/index.tsx
sandrahoang686: That should be fine since InternalNavLinks and ExternalNavLinks can actually be used throughout the app and not just header/footer 🤔 . 

Btw, we should remove these from `/common/page-header/types.ts` then so we dont have dups. I should've also mentioned this before, my bad.. We expose these types to Next instance as well, this import will need to be updated [here](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/index.ts#L21-L25). Along with the other imports that previously imported from og path. And then because we expose these types to the Next instance, we will have to update the types file name from `types.d.ts` to `types.ts` to have them exposeable through library build (context [here](https://github.com/NASA-IMPACT/veda-ui/pull/1223)).

app/scripts/components/common/layout-root/index.tsx
hanbyul-here: Then why would SSR matter in this context?

app/scripts/components/common/layout-root/index.tsx
hanbyul-here: (Not a blocker just a question) - are we keeping the environmental variable to turn on/off footer? I thought those are temporary solutions for instances not ready to adopt the new footer, header.

app/scripts/styles/_uswds-theme.scss
snmln: Yes, these changes are needed for the USWDS footer component to function appropriately.

app/scripts/components/common/page-footer/page-footer.scss
hanbyul-here: Unless we absolutely need a specificity of ID, imo it is better not to use ID for styling. (I think there are a lot of smart articles explaining this including this one https://css-tricks.com/specifics-on-css-specificity/) 

I know this PR has been opened for long, so I will leave it up to you if you want to handle this separately.

app/scripts/components/common/layout-root/index.tsx
sandrahoang686: env vars are for legacy architecture / current instances to turn on new header/footer if they wanted. New architecture can just import and use the footer or not which is why `hideFooter` made me scratch my head.

sandrahoang686: exploration page connects the dot 👍🏼

app/scripts/components/common/page-footer/page-footer.scss
snmln: This was a carry over from the exploration to solve the initial issues we ran up against attempting to get the media queries to work appropriately. We can remove them as id's but we do need these custom classes since they are setting the overall padding and appearance of the footer inline with the custom USWDS theme variables.

app/scripts/components/common/layout-root/index.tsx
AliceR: In the sense of leaving the camp ground cleaner I muted the lint warning by disabling the rule `react-hooks/exhaustive-deps` for this line - and left a comment to explain why. Obviously my comment wasn't helpful though 😅

hanbyul-here: Yeah during SSR gave me impression that the empty array was used to work around Next instance. I will suggest a omdification.

hanbyul-here: Ah it is already handled 😓 thanks!

snmln:
@faustoperez Could you take a quick look at this and provide your review: https://deploy-preview-1285--veda-ui.netlify.app/

sandrahoang686 (approved):
thanks for making the changes

hanbyul-here (approved):

faustoperez (approved):
I understand this ticket captures the desktop work. If that's the case, everything is looking good!

The mobile adaptation will be implemented in a separate ticket, right?

snmln:
That is correct the mobile breakpoints will be handled in this ticket: https://github.com/NASA-IMPACT/veda-ui/issues/1288

PR #1287, opened 2024-12-05
Hide new USWDS footer behind feature flag

AliceR:
Hej @snmln ! Thanks for setting up the feature branch and foundations for the new footer component! 
I renamed the old footer to PageFooterLegacy, that way we can use the regular import path for our new footer component. Also added a .env var that allows us to switch between old and new footer implementations easily.

### DIFF
--- a/app/scripts/components/common/layout-root/index.tsx
+++ b/app/scripts/components/common/layout-root/index.tsx
@@ -11,8 +11,11 @@ import styled from 'styled-components';
 import { Outlet } from 'react-router';
 import { reveal } from '@devseed-ui/animation';
 import { getBannerFromVedaConfig, getCookieConsentFromVedaConfig } from 'veda';
+
 import MetaTags from '../meta-tags';
-import PageFooter from '../page-footer/index';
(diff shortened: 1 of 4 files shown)


app/scripts/components/common/layout-root/index.tsx
```diff
@@ -98,7 +104,11 @@ function LayoutRoot(props: { children?: ReactNode }) {
           />
         )}
       </PageBody>
-      <PageFooter hideFooter />
+      {isUSWDSEnabled ? (
```
snmln, line 107: This all looks good! I may have missed some context from the header nav work, but what is the reason we are hiding the new refactor work behind a **isUSWDSEnabled** flag?

snmln: Is this related to the USWDS token work and requires additional implementation steps from the instances?

snmln: We discussed this verbally in our devsync.

snmln (approved):

PR #1308, opened 2024-12-09
refactor: 1135 slim footer

snmln:

### Description of Changes
Creating slim footer.

- Establishing initial Front-end
- Creating footer object model in veda.config.js
- Setting up initial default.config.js 
- Wiring up Config data flow
- Turning USWDS breakpoints on

TO DO's for this PR:

- Implement dynamic creation of links in footer component


app/scripts/components/common/page-header/default-config.ts
```diff
@@ -8,44 +18,125 @@ import {
...
+const defaultFooterSettings = {
+  secondarySection: {
+    title: 'email test',
+    to: '/data-catalog',
+    type: 'Email'
+  },
+  returnToTop: true
+};
+
+const defaultFooterPrimaryContactItems = [
```
AliceR, line 81: do we want the footer items to be different from the header? In my understanding, it was just a repetition of the main nav items - which would mean we should use the same configuration that we base the header on.

app/scripts/components/common/page-footer/styles.scss
```diff
@@ -48,18 +48,25 @@
...
-  padding: themeVars.$veda-uswds-padding-4 themeVars.$veda-uswds-padding-5;
-}
 
 .usa-footer__secondary-section {
   background-color: themeVars.$veda-uswds-color-base-darkest;
+  padding: themeVars.$veda-uswds-padding-4 themeVars.$veda-uswds-padding-5;
+  & > .grid-container {
+    max-width: 100%;
+    padding: 0;
+    //NEED to determine how to remove 64rem max-width on grid-container
```
AliceR, line 58: why do we need to remove the max-width? This is the desktop breakpoint afaik

app/scripts/components/common/page-footer/styles.scss
```diff
@@ -48,18 +48,25 @@
...
 }
 .usa-footer__primary-section {
   background-color: themeVars.$veda-uswds-color-base-lightest;
 }
 
-#mail_icon{
-  top:50%;
+#mail_icon {
+  top: 50%;
   transform: translateY(-50%);
```
AliceR, line 67: this is only needed because our svg is off, right? maybe we should add a comment, to remember us about cleaning up after fixing the svg

app/scripts/components/common/page-footer/index.tsx
```diff
@@ -34,13 +34,14 @@ export default function PageFooter({
   };
 
   const { returnToTop, secondarySection } = settings;
+  /* eslint-disable */
```
AliceR, line 37: Just wondering: What was the issue, and why did you decide to disable eslint instead of fixing it?

app/scripts/components/common/page-footer/styles.scss
```diff
@@ -1,7 +1,72 @@
 @use '$styles/veda-ui-theme-vars.scss' as themeVars;
+
+#logo-container {
```
AliceR, line 3: I don't think we want all the `#logo-container` styles here. I am guessing you only meant to add the styles for `#nasa-logo-pos`?

AliceR (commented):
The footer is taking shape, thank you @snmln 🙌 

The only thing that requires an actual conversation is the change to the configuration: I am questioning if this is necessary at all. Could we generate the footer from everything that we already have in the config for the header?

app/scripts/components/common/page-footer/styles.scss
snmln: That is correct this is a hold over from the header work I copied in to get the logo to preview.

app/scripts/components/common/page-footer/index.tsx
snmln: non-used variable errors. That is only there to get the checks to pass on this PR so we can merge it. They should be deleted prior to this being merged into main.

app/scripts/components/common/page-footer/styles.scss
snmln: Correct, the existing footer function when at desktop view expands to the full width of the entire window. With this USWDS max-width in place it only allows for it to stretch to 64rem and leaves pretty large gutters on either side:  Example: https://trussworks.github.io/react-uswds/iframe.html?args=&id=components-footer--slim-footer&viewMode=story

This might be more of a @faustoperez question. At desktop do we want the [large gutter to the left and right](https://trussworks.github.io/react-uswds/iframe.html?args=&id=components-footer--slim-footer&viewMode=story) of the footer content at desktop or do we want the content to stretch to[full width of the screen](https://deploy-preview-1308--veda-ui.netlify.app/)?

app/scripts/components/common/page-header/default-config.ts
snmln: As I was working through this, this question came up and I think we should discuss further. My thoughts are that  we can and should model the footer materials after the header data model. But the instances are going to request for individualized configurability of the footer so they can customize past the header navigation. What are your thoughts?

app/scripts/styles/_uswds-theme.scss
```diff
@@ -14,9 +14,11 @@
...
   $theme-utility-breakpoints: (
-    'mobile-lg': false,
-    'desktop': false
+
+    'mobile-lg': true,
+    'desktop': true
   ),
   $theme-font-type-serif: baseFontFamily,
   $theme-font-type-sans: baseFontFamily
 );
```
snmln, line 23: One item I want to call out @AliceR for the footer class to work out of the box I needed to turn on the breakpoints in the theme utility.

app/scripts/components/common/page-footer/styles.scss
AliceR: IMO we should follow the header and leave the gutter. If I understand it correctly we wil have a page layout later that will align all the components on a page.

app/scripts/components/common/page-header/default-config.ts
AliceR: This sounds like a good case for [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) - let's wait for that request to come in first :)

app/scripts/styles/_uswds-theme.scss
AliceR: I wonder why we had them set to false, and what the consequences are with the change?

snmln:
Merging this into the main feature branch. @AliceR if you could implement the nav elements how you see fit and leverage the main nav items from veda config i think we should be good to keep this moving.

PR #1310, opened 2024-12-09
chore: Remove linkProperties from DataCatalog Component

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1108
**Related PR:** https://github.com/developmentseed/next-veda-ui/pull/25

 **v5.11.3-alpha.0** published from this branch

### Description of Changes
As part of the greater holistic approach - I think its best to move away from having our library components tightly integrated w/ routing. This is the first iteration to remove **linkProperties**. This PR only worries about removing **linkProps** the **DataCatalog Component**. DataCatalog view no longer has to pass in linkProperties or have to directly worry about routing, we can just pass in a callback now that decides what to do during some action.

This is an iterative approach, i've created tickets to remove routing from the other core components. But we can't remove routing directly from the Card component itself easily as GHG uses the card component directly [here](https://github.com/US-GHG-Center/veda-config-ghg/blob/dfe611f27f910bc6428a2b13fe50678710511e68/custom-pages/news-and-events/component.tsx#L5). So that should be its own separate ticket BUT... we are to be redoing the card component for the new instances - where its probably best to rewrite it. As currently its not in an ideal place to scale... so i'll create a placeholder ticket for card component for now but we can probably tackle removing routing when rewriting the card component.

cc @vgeorge @hanbyul-here @dzole0311 

Follow-up tickets created: 
* https://github.com/NASA-IMPACT/veda-ui/issues/1325
* https://github.com/NASA-IMPACT/veda-ui/issues/1326
* https://github.com/NASA-IMPACT/veda-ui/issues/1327

### Validation / Testing
- [ ] Validate DataCatalog cards are linking correctly
- [ ] Validate selecting cards in DatasetSelectorModal is working as it should


sandrahoang686:
@hanbyul-here i wouldn't really classify this as a `refactor` but more so `tech-debt` or `chore` but its a little more than a chore.. What do you think is best to classify this as? `tech-debt` doesn't exist as one of the CI options. Also this isn't breaking for current legacy architecture but is breaking for new architecture / nextJs instance - so should I add the `!`?

hanbyul-here:
`chore` sounds right for me. + Conventional commit is to make the automatic versioning seamless - so let's not worry too much about if we have a 'right' prefix as long as it will version things properly.

dzole0311 (approved):
Tested the routing & selecting for the Catalog cards, it works as expected. Also confirming the overline prop isn't needed for the selectable (`horizontal-info`) catalog cards.

Issue #1259, opened 2024-11-18
[Refactor][Bugs] Preset AOIs in E&A


sandrahoang686:
### Context
* AOI preselected states are taking a while to load and are erroring

Goal of this ticket to reduce errors and improve load times

aboydnw:
@sandrahoang686 I adjusted this ticket to only be about preselected AOIs since I already had another ticket for https://github.com/NASA-IMPACT/veda-ui/issues/1256

PR #1311, opened 2024-12-10
fix: detect Next.js environment and adjust the AOI preset file paths

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1259
**Related Next.js PR for testing:** https://github.com/developmentseed/next-veda-ui/pull/23

### Description of Changes
- Added new `isNextJs()` util function to detect Next.js environments
- Adjust the AOI preset path to use the new util for setting the correct path to the assets
- Persist backwards compatibility for legacy instances that require the `/public/` path prefix
- Removed redundant comments

### Validation / Testing
Please test in both repositories:

veda-ui:

1. Using the preview deployment, check that preset AOIs load correctly when selecting different states
2. Test locally to confirm the same functionality
3. Optional: Check in a legacy instance (e.g. GHG Center) that AOIs still load with the /public/ prefix

In next-veda-ui:

1. Using the [preview deployment](https://deploy-preview-23--veda-ui-next-test.netlify.app/exploration), check that preset AOIs load correctly when selecting different states
2. Test locally to confirm the same functionality works

### DIFF
--- a/app/scripts/components/common/map/controls/hooks/use-preset-aoi.tsx
+++ b/app/scripts/components/common/map/controls/hooks/use-preset-aoi.tsx
@@ -3,23 +3,28 @@ import { useCallback, useEffect, useState } from 'react';
 import axios from 'axios';
 import { getAoiAppropriateFeatures } from './use-custom-aoi';
 
-const presetFilePath = `${process.env.PUBLIC_URL ?? ''}/public/geo-data/states/`;
-const presetSuffix = `.geojson`;
-
 function usePresetAOI(selectedState) {
-  const [error, setError] = useState<string|null>('');
+  const [error, setError] = useState<string | null>('');
   const [isLoading, setIsLoading] = useState<boolean>(false);
   const [features, setFeatures] = useState<Feature<Polygon>[] | null>(null);
 
   useEffect(() => {
     if (!selectedState) return;
-    const abortController = new AbortController();  // Create an instance of AbortController
+    const abortController = new AbortController();
 
     async function loadData() {
       setIsLoading(true);
       try {
-        const res = await axios.get(`${presetFilePath}${selectedState}${presetSuffix}`, {
-          signal: abortController.signal  // Pass the abort signal to Axios
+        // We're dynamically adjusting the base path based on the environment:
+        // 1. If running in Next.js, omit "/public" from the path
+        // 2. For legacy instances, we include "/public" as part of the path
+        const basePath =
+          typeof window !== 'undefined'
+            ? `/geo-data/states/`
+            : `${process.env.PUBLIC_URL ?? ''}/public/geo-data/states/`;
+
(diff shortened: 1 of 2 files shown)


AliceR (approved):
Thanks for adding comments, and cleaning up as well! 🧹

app/scripts/components/common/map/controls/hooks/use-preset-aoi.tsx
```diff
@@ -2,24 +2,29 @@ import { Feature, Polygon } from 'geojson';
...
-          signal: abortController.signal  // Pass the abort signal to Axios
+        // We're dynamically adjusting the base path based on the environment:
+        // 1. If running in Next.js, omit "/public" from the path
+        // 2. For legacy instances, we include "/public" as part of the path
+        const basePath = isNextJs()
+          ? `/geo-data/states/`
+          : `${process.env.PUBLIC_URL ?? ''}/public/geo-data/states/`;
+
+        const res = await axios.get(`${basePath}${selectedState}.geojson`, {
+          signal: abortController.signal
```
sandrahoang686, line 27: 🤔 hmm have we thought about other ways that wont constrain us to check by the framework being used? I imagine this being fine for a temporary solution but not long term. Early on we talked about how we plan on using NextJs for the template instance and ideally for the other instances but really the other instances would have free reign to decide their framework. I dont think we want to default to solving framework related issues by checking if its a specific framework.

if you wanting to move forward with this for now to unblock something, could a `@TECH-DEBT` or `@TODO` note be left?

hanbyul-here: Yeah I agree it would be ideal if we don't need to hardcode and have an assumption about what environment VEDA is running. If we need a work around, can we put these files into `/public/geo-data/states/` from next instance side for now?

dzole0311: @hanbyul-here We had it like that before which works locally, but that won't work when the Next.js app is built for production and deployed since the `/public` path seems to be omitted ([here's an older preview](https://deploy-preview-22--veda-ui-next-test.netlify.app/exploration) from Next.js with that approach).

> have we thought about other ways that wont constrain us to check by the framework being used

@sandrahoang686 Yeah, @AliceR raised the same question. Two alternative solutions I can think of with the current setup:

1. Pass the path as an env variable via the new `EnvConfigProvider` to the `PresetSelector` component, or
2. Pass the path as a prop to the E&A component

Any preferences for the team or other ideas?

dzole0311: Closing this PR since we decided to fix it by creating a `/public` subdirectory within the `/public` directory on the Next.js instance side.

dzole0311:
Closing this PR since we decided to fix it by creating a /public subdirectory within the /public directory on the Next.js instance side.

PR #1315, opened 2024-12-11
fix: hideFooter prop name, small cleanup

AliceR:
Hi @snmln , I fixed the prop name, did some cleanup in the css styles and merged the main branch.


AliceR:
Ok sure, I removed the merge commit. 

Also noting, the tests are failing on the feature branch, they just don't run there because it in draft mode.

PR #1316, opened 2024-12-11
WIP fix svg logo

AliceR:
I am downloading the svg from https://www.nasa.gov/wp-content/themes/nasa/assets/images/nasa-logo.svg to be able to use it without any translation or other css. It has some differences to the current logo, so that I need to make sure everything is still backwards compatible... did not finish in time, hence WIP. Just for visibility!


snmln:
Sounds good! Here is the export SVG from Faustos designs incase you run into any issues with the linked one. 
NASA_logo

snmln (approved):

PR #1320, opened 2024-12-11
fix: 1135 wiring up veda config

snmln:
@AliceR I am getting the menu items wired up.


app/scripts/components/common/page-footer/index.tsx
```diff
@@ -36,7 +39,53 @@ export default function PageFooter({
...
+          return (
+            <a className={linkClasses} href={item.to} key={item.id}>
+              {item.title}
+            </a>
+          );
+
+        default:
+          return <></>;
+      }
+    });
```
snmln, line 78: I know we have similar functionality from the header work here: https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/page-header/nav/create-dynamic-nav-menu-list.tsx

The issue is that the header styling is implicit to this function. We should reuse this functionality but expand it to accept different stylings. I need to look into this a little more to wrap my head around it better. But based on my current understanding that seems like a decent amount of work to accomplish. 

My opinion (with my current understanding) is that we note it and create a dedicated clean up ticket to integrate that styling functionality since this will bleed into the header work that was just created. 

With your familiarity with the functionality what are your thoughts?

sandrahoang686: you could move the nav directory with `createDynamicNavMenuList` et al outside of header to be accessible to header&footer and still use the [nav-item-link](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/page-header/nav/nav-item-links.tsx) and [CTA](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/page-header/nav/nav-item-cta.tsx) components like `createDynamicNavMenuList` already does but make it take in the className now as a param and pass it down as a prop to the nav link components. I dont think footer will need to worry about dropdowns. For CTA, you could make the style a prop as well but default it. 

That doesn't have to be done here though, feel free to leave an `@TODO` or `@TECH-DEBT` comment to consolidate and create a ticket to come back and consolidate this functionality to make it more flexible to styling.

snmln: Sweet I have gone ahead and created a new issue: https://github.com/NASA-IMPACT/veda-ui/issues/1323 for this.

Issue #1235, opened 2024-10-31
Add evergreen banner for earth.gov


aboydnw:
### Context

earth.gov would like to use the banner component to consistently display the message that it is an official site of the US government, similar to the example from [USWDS here](https://designsystem.digital.gov/components/banner/).

### Acceptance Criteria
- [ ] A banner is shown on earth.gov every time a user visits, with content matching the preview for .gov domains [here](https://designsystem.digital.gov/components/banner/)

aboydnw:
There are some UX decisions for us to make here. The main question is that the banner component from USWDS seems to be intended as a permanent banner only speaking to the legitimacy of the website as an official site. We have been using this component as a banner for temporary announcements. The earth.gov team, however, would like the permanent banner stating it is an official site.

I could see a couple options.

### Option 1 - Adjust existing component to be announcements banner and add permanent banner

We could implement the permanent banner as-is from USWDS. Then, we could change the styling of the existing banner and label it "announcement banner" or something similar. This would function the same with an expiration date. This announcements banner could stack on top of the permanent banner (if both are turned on), be a different color, and include an "x" or something for the user to close it. Whereas the permanent banner would always remain in place with no way to hide it.

Upside of this is that it's a cleaner implementation of the USWDS banner component and also retains the ability to make announcements. But, potentially confusing UX to have two similarly styled components that function differently. Also, as far as announcements go, it's pretty subtle. 

### Option 2 - Use existing component with an expiration date way in the future, so it seems permanent

We use the existing component but put a date of 2050 so it never disappears.

Upside being, less development work and no potentially conflicting banners. But we also won't have the dropdown seen in the USWDS component and we won't have the ability to also do a separate announcement. Although, to be fair, only GHGC has used the announcement feature so far.




If we pick this up @dzole0311 then we probably want to make an intentional decision on this with @faustoperez . I wrote this ticket assuming Option 2 just for the sake of speed, but if we see a good solution to Option 1, then I'm okay going that direction too. I just don't think it's worth our time to go too deep designing a solution for this feature right now. So if we don't quickly see a solution, I'd stick to Option 2.

aboydnw:
Just closing the loop that we would like to go with Option 1 above. I updated the original comment/AC to reflect that

faustoperez:
There are two separate USWDS components for this as they serve different purposes:

- The **banner** is the official website identifier 👉 https://designsystem.digital.gov/components/banner/
[...]

PR #1328, opened 2024-12-16
feat: Introduce the USWDS Site Alert as info banner, use the USWDS Banner as intended

dzole0311:
**Closes:** https://github.com/NASA-IMPACT/veda-ui/issues/1235

Demo: https://www.loom.com/share/f7c5868033054d859a0a4177722951a6?sid=42347137-07f4-46de-8f1d-25021ccf3204

### Description of Changes
1. Replaced the `Banner` component with `SiteAlert` but kept some of the core functionalities such as the expiration date feature and close button
2. Re-factored the `Banner` to align with its intended USWDS role i.e. showing government information at the top of every page

I'm still a little unsure whether the Banner and SiteAlert components should live within the veda-ui library or be implemented specifically within the instances. Since Banner was already part of the veda-ui, I followed the same approach for the SiteAlert.

### Validation / Testing

See the configuration for the SiteAlert and the Banner [here](https://github.com/NASA-IMPACT/veda-ui/pull/1328/files#diff-828262a6a39b2c45c85b5fcdf1b4615ff5d3d69f662078fd27aa9846fb29f36fR74). Note: Run `yarn clean` to clear the Parcel cache if the changes you make in the config are not reflected in the UI

Banner testing:

1. Check that the banner appears consistently across all pages when navigating routes
2. Check that the expand/collapse button ("Here's how you know") works
3. Check that the banner looks good on smaller screens as well

SiteAlert testing:

1. Check that the SiteAlert is shown when:
- Content exists for the alert
- The expiration date is not reached
- The alert has not been dismissed by the user
2. Check these refresh scenarios:
- Without closing the alert, refresh the page to verify that it reappears
- After closing the alert, refresh the page to verify that it remains hidden


package.json
```diff
@@ -181,6 +182,7 @@
     "google-polyline": "^1.0.3",
     "gulp-postcss": "^10.0.0",
     "gulp-sass": "^6.0.0",
+    "he": "^1.2.0",
```
AliceR, line 185: What is he doing? 😄

app/scripts/components/common/banner/index.tsx
```diff
@@ -1,78 +1,141 @@
...
-  }
+  const leftContent = {
+    ...defaultGuidance.left,
+    ...leftGuidance
+  };
+
+  const rightContent = {
+    ...defaultGuidance.right,
+    ...rightGuidance
+  };
```
sandrahoang686, line 76: Does using this banner always default to having two pieces of content since this means there will always be a right and left and never just a single piece of content? This is fine for now though as it meets the need, other needs we could defer to later but just curious

app/scripts/components/common/banner/index.tsx
```diff
@@ -1,78 +1,141 @@
...
+                <strong>{rightContent.title}</strong>
+                <br />
+                <span
+                  dangerouslySetInnerHTML={{
+                    __html: decode(rightContent.text)
+                  }}
+                />
+              </p>
+            </USWDSMediaBlockBody>
+          </USWDSBannerGuidance>
```
sandrahoang686, line 136: ⛏️ : Since this JSX block repeats twice you could extract it into a fn like `displayGuidanceBlock` and have it take in `(content: GuidanceContent)` as a param

app/scripts/components/common/banner/index.tsx
```diff
@@ -1,78 +1,141 @@
...
+  className = '',
+  defaultIsOpen = false,
+  contentId = 'gov-banner-content'
 }: BannerProps) {
+  const [isOpen, setIsOpen] = useState(defaultIsOpen);
 
-  const showBanner = localStorage.getItem(BANNER_KEY) !== url;
-  const [isOpen, setIsOpen] = useState(showBanner && !hasExpired(expires));
+  const defaultHeaderText =
+    'An official website of the United States government';
```
sandrahoang686, line 66: ⛏️ : headerActionText could probably follow this pattern too or headerText could follow headerActionText with a default value in props. If we keep it like this, this could probably be defined outside like `defaultGuidance` which would be beneficial too if we ever do test this to export for ref

app/scripts/components/common/site-alert/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  appTitle,
+  expires,
+  content,
+  type = infoTypeFlag,
+  heading,
+  showIcon = true,
+  slim = false,
+  className = ''
+}: SiteAlertProps) {
+  const alertId = content;
```
sandrahoang686, line 43: curious to why creating a copy?

mock/veda.config.js
```diff
@@ -103,10 +118,24 @@ module.exports = {
...
+    leftGuidance: defaultGuidance.left,
+    rightGuidance: defaultGuidance.right,
+    className: '',
+    defaultIsOpen: false,
+    contentId: 'gov-banner-content'
+  },
+  siteAlert: {
+    content:
+      '<p class="usa-alert__text"><a target="_blank" rel="noreferrer" href="stories/emit-and-aviris-3">Read the new data insight</a> on using EMIT and AVIRIS-3 for monitoring large methane emission events.</p>',
+    expires: '2026-08-03T12:00:00-04:00',
```
sandrahoang686, line 135: This is just mock but clicking on it leads to Page not found since we dont have this story in veda-ui. Just an FYI if you weren't already aware since I dont know if we want

app/scripts/components/common/site-alert/index.tsx
dzole0311: I dded it thinking that we can pass an `alertId` prop, as I feel it's better to handle different alerts like that than to rely on the content string as an unique identifier. But since the current approach works fine, I'm removing the `alertId`

sandrahoang686 (approved):
lgtm :shipit:

snmln (approved):

app/scripts/components/common/banner/index.tsx
```diff
@@ -1,78 +1,136 @@
...
+  className?: string;
+}) => (
+  <USWDSBannerGuidance className={className}>
+    <USWDSBannerIcon src={content.icon} alt={content.iconAlt ?? ''} />
+    <USWDSMediaBlockBody>
+      <p>
+        <strong>{content.title}</strong>
+        <br />
+        <span
+          dangerouslySetInnerHTML={{ __html: decode(content.text ?? '') }}
```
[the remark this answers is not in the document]
dzole0311: Thanks @hanbyul-here! Does this mean we will restrict the Banner component to only allow users to specify a predefined set of content types (e.g., a url prop and a text prop) for which we would then render the appropriate html tags (`<p />`, `<a />`) directly within the Banner?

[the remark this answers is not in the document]
hanbyul-here: No - I meant that user can pass the contents as markdown like `::markdown ### heading` - I see that you used some html class for mock data https://github.com/NASA-IMPACT/veda-ui/blob/main/mock/veda.config.js#L133 I think since markdown accepts html as a part of markdown, it should be handled ok with markdown parser? I mainly want all the user-configured inputs for our component contents to be in a coherent format, and the parsing of the contents to be handled in the same place.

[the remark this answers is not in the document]
dzole0311: Created a ticket for myself: https://github.com/NASA-IMPACT/veda-ui/issues/1348

PR #1330, opened 2024-12-16
fix:1335 creating footer test suite

snmln:
Creating initial test suite

### DIFF
--- a/app/scripts/components/common/page-footer/index.tsx
+++ b/app/scripts/components/common/page-footer/index.tsx
@@ -84,13 +84,12 @@ export default function PageFooter({
       createNavElement(subNavItems, 'usa-link text-base-dark text-underline'),
     [mainNavItems]
   );
-  return hideFooter ? (
-    <></>
-  ) : (
+  return (
     <>
       <USWDSFooter
         size='slim'
         returnToTop={returnToTop && returnToTopButton()}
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/page-footer/index.tsx
```diff
@@ -82,12 +84,13 @@ export default function PageFooter({
       createNavElement(subNavItems, 'usa-link text-base-dark text-underline'),
     [mainNavItems]
   );
-  return (
+  return hideFooter ? (
+    <></>
```
AliceR, line 88: I am not a fan of returning empty tags. We could leave 'display: none' and test with '.not.toBeVisible()'.

Unless there is another reason for '< />' I am not seeing?

snmln: The original reason was to avoid the footer being consumed by the accessibility tree but after doing some research it appears that `display-none;` Shows the element in dev tools but doesnt consume it in the accessibility tree. So we should be good to use `display-none`

snmln: Heads up `.not.toBeVisible()`didnt work, since it was being applied through the util class it didnt fully compile the overlapping styles and returned a dispaly: block. So I opted to test that the correct css class was applied.

PR #1331, opened 2024-12-17
fix: Add secondary footer content to footer config

AliceR:
I am adding some more fields to the secondary section in the footer section, in order to have any text content configurable: Added `division`, `version`, and `name`, repurposed `title`. 

I'd be curious to know how other instances might use and configure the footer, this could better inform naming decisions (e.g. `division`?).


snmln (approved):
This all looks good to me!

Issue #1332, opened 2024-12-17
[Bug] Data catalog links duplicating dashboard in url


aboydnw:
### Context

URL: https://www.earthdata.nasa.gov/dashboard/data-catalog

Blocking a release? (y/n): n

Regression issue? (y/n): y 
[If yes, add Regression label]

Priority: P2

### Steps to reproduce the issue

1. [Step 1 to reproduce]
2. [Step 2 to reproduce]

PR #1333, opened 2024-12-17
fix(usePathname): prevent duplication of subpath in pathname

vgeorge:
**Related Ticket:** #1332

### Description of Changes

When using `usePathname()` hook, strip the application pathname to avoid duplicating it in generated links. 

### Validation / Testing

App is running from the root path:

- Make sure `PUBLIC_URL` is undefined in environment files
- Start the app with `yarn dev`, visit http://localhost:9000/data-catalog
- Clicking on the cards takes to the correct page

Apps is running from a subpath:

- Set `PUBLIC_URL=/subpath` in the environment file
- Start the app with `yarn dev`, visit http://localhost:9000/subpath/data-catalog
- Card links are correct

### DIFF
--- a/app/scripts/utils/use-pathname.ts
+++ b/app/scripts/utils/use-pathname.ts
@@ -13,14 +13,18 @@ import { useEffect, useState } from 'react';
  */
 export const usePathname = () => {
   const [pathname, setPathname] = useState(
-    typeof window !== 'undefined' ? window.location.pathname : ''
+    typeof window !== 'undefined'
+      ? window.location.pathname.replace(process.env.PUBLIC_URL ?? '', '')
+      : ''
   );
 
   useEffect(() => {
     if (typeof window === 'undefined') return;
 
     const updatePathname = () => {
-      setPathname(window.location.pathname);
+      setPathname(
+        window.location.pathname.replace(process.env.PUBLIC_URL ?? '', '')
+      );
     };
 
     // Listen to popstate events (back/forward navigation)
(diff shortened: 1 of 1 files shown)


app/scripts/utils/use-pathname.ts
```diff
@@ -13,14 +13,18 @@ import { useEffect, useState } from 'react';
...
+      : ''
   );
 
   useEffect(() => {
     if (typeof window === 'undefined') return;
 
     const updatePathname = () => {
-      setPathname(window.location.pathname);
+      setPathname(
+        window.location.pathname.replace(process.env.PUBLIC_URL ?? '', '')
```
hanbyul-here, line 26: Some of our instances configure public_url as a full url, (not only subpath - like https://earth.gov/ghgcenter) https://github.com/US-GHG-Center/veda-config-ghg/blob/develop/.github/workflows/deploy.yml#L82

Would this approach still work in this case?

vgeorge: @hanbyul-here good catch, thanks. I'll push a fix for this case.

hanbyul-here (approved):
🙇 thanks for looking at it - I do think we need this fix. I also want to add that all the frameworks have a different way of handling this public_url thing (ex.Next uses basePath) 

It will be really ideal if we can fully let the router of the application handle this. But we need this fix for now for instance, so I will approve it - but I will think about how we should handle this further.

Issue #1326, opened 2024-12-16
Remove LinkProperties from DatasetSelectorModal Component


sandrahoang686:
This is a follow-up ticket from https://github.com/NASA-IMPACT/veda-ui/issues/1108#issuecomment-2544614096.
And a follow-up from this PR https://github.com/NASA-IMPACT/veda-ui/pull/1310.

We need to remove routing/nav dependencies from DatasetSelectorModal Component. Remove `LinkProperties` prop dependency with a callback to handle navigation instead.

sandrahoang686:
@vgeorge this [bug-fix PR](https://github.com/NASA-IMPACT/veda-ui/pull/1335/files) recently merged has to do with this ticket as an FYI! left an `@TODO` note in the code https://github.com/NASA-IMPACT/veda-ui/blob/ea766a188495cad69ff89464442ca825dc19fa10/app/scripts/components/exploration/container.tsx#L91

PR #1335, opened 2024-12-17
chore: BUG-FIX, pass in nav callback to exploration / layer-info-modal

sandrahoang686:
**THIS IS TO FIX A BUG** 🐛 🐞  look at linked thread...
**Related Ticket:** https://github.com/US-GHG-Center/veda-config-ghg/issues/689
**Related Thread for GHG** https://developmentseed.slack.com/archives/C03LGTM2SSH/p1734459208927079?thread_ts=1734456710.644149&cid=C03LGTM2SSH

**DEPLOY PREVIEW FOR GHG that points to this commit** https://github.com/US-GHG-Center/veda-config-ghg/pull/690

### Description of Changes
_{Update with description of changes made in this pull request}_

### Notes & Questions About Changes
_{Add additonal notes and outstanding questions here related to changes in this pull request}_

### Validation / Testing
_{Update with info on what can be manually validated in the Deploy Preview link for example "Validate style updates to selection modal do NOT affect cards"}_

PR #1338, opened 2024-12-18
refactor: enable routing config via provider, remove linkProps from DatasetSelectorModal

vgeorge:
**Related Tickets:** #1326 #1108

### Description of Changes

- Remove custom link handling in favor of `VedaUIProvider`'s `Link`
- Configure `VedaUIProvider` with `react-router`'s `Link` component
- Remove `SmartLink` from the exploration container
- Update `DatasetSelectorModal` to use the new `Link` component

We discussed in #1108 approaches for making the library agnostic to the routing library. In [this comment](https://github.com/NASA-IMPACT/veda-ui/issues/1108#issuecomment-2549610903), I explained why I think we should avoid relying solely on navigation callbacks, as they degrade the accessibility of the library.

In the approach introduced in this PR, we expand the existing configuration provider to also allow passing routing configuration, making the library compatible with `react-router`, `next.js`, or any other routing framework.

This PR became quite large but most changes are on invocation calls to the refactored hook. Please let me know if you have any difficulties when reviewing it.

### Validation / Testing

The `EnvConfigProvider` was replaced by `VedaUIProvider`. The maps and any other functionality depending on the Mapbox token and API endpoints should work as before.

The `DatasetSelectorModal` only uses `Link` when empty. To test, replace the line [here](https://github.com/NASA-IMPACT/veda-ui/blob/bed9af9f420e1766e1ae606928ad1eb983ad61b2/app/scripts/components/exploration/components/dataset-selector-modal/index.tsx#L126) with `datasets={[]}`, visit the exploration page, and confirm that the link in the empty state message works correctly.


sandrahoang686:
> I explained why I think we should avoid relying solely on navigation callbacks

@vgeorge - Yep, I agree. Callbacks are meant for actions like buttons which is why I didn't include pageheader as part of this work! Since it should retain the link component. But I have some thoughts with this approach, it might be best to meet and talk over a call.

app/scripts/context/veda-ui-provider.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+interface LinkProps {
+  to: string;
+  children: ReactNode;
+  [key: string]: any;
+}
+
+interface NavigationConfig {
+  LinkComponent: React.ElementType<Record<string, any>>;
+  linkPropName: string;
```
hanbyul-here, line 17: (sorry this is not directly related to what this PR changes but since we are on it) - I wonder if we need to make this as an object (linkProps) and make the linkPropName as a part of it since we can't predict what props to expect.

vgeorge:
We had a chat today about this (@sandrahoang686, @hanbyul-here, @dzole0311 and I). The output was:

- @sandrahoang686 pointed out we should be agnostic on the DatasetSelectModal with routing/nav and will propose a new approach to handling empty state for it to this PR
- The VedaUI provider, while not necessary in the Modal, is still necessary in other parts of the app and would be kept

Please let me if I missed something, thanks!

sandrahoang686:
@vgeorge @hanbyul-here @dzole0311 this is ready to be reviewed again based on our convo before holiday break. I also used this PR to reverse the navigation callback introduced in the E&A LayerInfoModal introduced in this [bug-fix PR](https://github.com/NASA-IMPACT/veda-ui/pull/1335/files). We will want to follow similar patterns using the ContextProvider from this PR in other places that use the callback (datacatalog) and other places that still use LinkProps.

We should validate that the E&A layer selection modal empty state links to the `/data-catalog` page and the layerInfoModal also links to catalog page.

app/scripts/context/veda-ui-provider.tsx
```diff
@@ -0,0 +1,115 @@
...
+  linkProps: {
+    pathAttributeKeyName: string; // href, to, ...
+    [key: string]: any;
+  }
+}
+
+interface RoutesConfig {
+  dataCatalogPath?: string;
+  storiesCatalogPath?: string;
+}
```
sandrahoang686, line 27: introducing this was the best solution i could think of  for instances who might name their paths differently

vgeorge:
@sandrahoang686 thanks for refactoring the DatasetSelectModal, it looks good to me. In addition to the comment above related to the routes configuration, I changed the following:

- Replaced direct usage of react-router's Link in E&A container
- Fixed linkProps usage in VedaUIProvider config

I'm flagging this for another review.

app/scripts/context/veda-ui-provider.tsx
sandrahoang686: I do use the route constants directly as a default https://github.com/NASA-IMPACT/veda-ui/blob/06cf39f6df8dd7b69e911e5e18002841cc7fd921/app/scripts/context/veda-ui-provider.tsx#L50-L53 but we need to allow being able to pass this in as part of the configuration since instances should be able to name their routes as they please. Technically instances might not even have a data catalog or stories page, its completely flexible now.

app/scripts/components/exploration/container.tsx
```diff
@@ -33,6 +33,8 @@ const tourProviderStyles = {
 };
 
 export default function ExplorationAndAnalysisContainer() {
+  const { Link } = useVedaUI();
```
sandrahoang686, line 36: @vgeorge we dont need to do this at this level because this is just the container to support current legacy architecture. It will always just be for veda-ui and never used in the new instance side architecture, therefor we can just use Link from react-router-dom directly. This container goes away btw once we migrate all instances over.

Issue #1242, opened 2024-11-05
Automate instance preview generation process


aboydnw:
### Context

It takes developer's valuable time to update veda-ui and create previews for testing.  We would like to make this easier for us to do as well as for any instance owners/contributors to conduct these releases. This issue is adjacently related to #1236 

### Acceptance Criteria

- [ ] When a veda-ui release is triggered, PRs for selected instances (GHG, earth.gov, and Earthdata) are created for testing purposes
- [ ] When a veda-ui release is triggered, appropriate team members are tagged on the release PR(s) and/or notified through slack

hanbyul-here:
@aboydnw I edited the ticket so it is more catered towards creation of preview URLs, not the release of instances.

PR #1339, opened 2024-12-18
fix(release): fix dispatch action

hanbyul-here:
**Related Ticket:** #1242 

I am sorry that I keep patching this action - it turned out that I cannot use app token across the jobs 🙇

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@developmentseed/veda-ui",
   "description": "Dashboard",
-  "version": "5.11.4",
+  "version": "5.11.3",
   "author": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@developmentseed/veda-ui",
   "description": "Dashboard",
-  "version": "5.11.4",
+  "version": "5.11.3",
```
hanbyul-here, line 4: Reverting this to `5.11.3` since the action for 5.11.4 succeeded only halfway 😓

dzole0311 (approved):

Issue #1296, opened 2024-12-06
Close the new USWDS page header dropdown when clicking outside of it


dzole0311:
### Context

The new USWDS page header nav dropdowns, after being opened, persist their open state even after attempting to click outside of the dropdown (e.g visiting another nav item, or clicking outside of the dropdown).

This is a follow up from the new USWDS page header work: https://github.com/NASA-IMPACT/veda-ui/pull/1247

Priority: 

### Steps to reproduce the issue

1. Enable the new page header via a feature flag (`ENABLE_USWDS_PAGE_HEADER`)
2. Click on one of the dropdowns
3. Then click on another nav item (e.g the Exploration nav item)
4. Notice that the dropdown remains open

PR #1342, opened 2024-12-18
fix: Menu Drop down

snmln:
Close the new USWDS page header dropdown when clicking outside of it
[#1296](https://github.com/NASA-IMPACT/veda-ui/issues/1296)

### Validation / Testing
Open one of the two drop down options, click on another link on the page. When the browser navigates to the new page that original dropdown menu should be closed.

### DIFF
--- a/app/scripts/utils/use-click-outside.ts
+++ b/app/scripts/utils/use-click-outside.ts
@@ -1,0 +1,15 @@
+import { useEffect, useRef } from 'react';
+export const useClickOutside = (onClose: () => void) => {
+  const ref = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    const handleClickOutside = (event: MouseEvent) => {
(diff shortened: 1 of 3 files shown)


app/scripts/utils/use-click-outside.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useRef } from 'react';
```
AliceR, line 1: Not very important, but maybe we can move both use-fix-hooks in the same directory? app/scripts/components/common/page-header/use-mobile-menu-fix.ts is the one i implemented for the resize bug. I don't care which directory, but I like them together :slightly_smiling_face:

dzole0311: @AliceR I think the new hook is generic enough to be useful in other places not limited to this component, so it might make sense to keep it in the general utils directory?

dzole0311 (approved):

Issue #1263, opened 2024-11-19
Ensuring a Clean, Reliable, and Developer-Friendly Codebase


AliceR:
Maintaining high code quality and consistency is crucial for the long-term success and maintainability of our open-source project. Adhering to a unified code style, implementing automated checks, and enforcing quality gates will help improve developer productivity, reduce bugs, and foster collaboration within the community.

Currently, there might be inconsistencies in how code is formatted, linted, or tested, which can lead to:

- Code review overhead due to unnecessary formatting or style issues.
- Potential introduction of errors if testing or linting is not consistently applied.
- Reduced contributor experience due to unclear or varying standards.

To address these issues, we propose implementing structured practices to automate and enforce code quality.

### Proposed Solution
**Step 1: Enable Format-on-Save in Local Development Environments**
Developers should configure their IDEs or text editors to automatically format code on save.
This ensures code is consistently styled as per project standards before it even reaches version control.
We can provide clear documentation in our repository on how to configure format-on-save for common editors (I think most of us use VS Code?).

**Step 2: Implement Git Hooks to Run Quality Checks**
We do have husky installed to the project, but the pre-commit hook is not running for me, at least. I need to investigate, and make sure it works for other new contributors as well!

Use Husky to set up pre-commit hooks that:
- Automatically format code.
- Run linters to catch syntax or style issues.
This ensures that no code can be committed unless it meets the project's standards.

Additionally, we could set up a pre-push hook to:
- Execute unit and end to end tests to catch issues early.

**Step 3: Enforce Quality Gates in CI/GitHub Actions**
Configure GitHub Actions (or any other CI tool) to run checks for formatting, linting, and tests.
We should the CI pipeline to fail and block merges for any violations.
This guarantees that only high-quality code reaches the main branch.

### Benefits
**Consistent codebase**: Code will always adhere to the project's style and standards.
**Improved productivity and collaboration**: Reduces time spent on manual corrections during code reviews, and less time discussing style issues.
**Early detection of issues**: Bugs and inconsistencies are caught before they reach production.
**Contributor-friendly**: Provides a clear and automated way for developers to meet standards.

-----------------------

### Action Items
- [x] Document format-on-save setup for popular editors in the repository's CONTRIBUTING.md. #1306 
- [x] Investigate and potentially fix our git hooks for formatting, linting, and testing using Husky. #1345 
- [x] Investigate and potentially update CI/GitHub Actions to enforce quality checks and block non-compliant PRs.

[...]

PR #1345, opened 2024-12-19
chore: review commit hooks

### LINKED ISSUE #1329: Review commit hooks
### Context

Follow up from #1263 
Focusing on the second AC from that ticket:

> Step 2: Implement Git Hooks to Run Quality Checks
> We do have husky installed to the project, but the pre-commit hook is not running for me, at least. I need to investigate, and make sure it works for other new contributors as well!
> 
> Use Husky to set up pre-commit hooks that:
> 
> Automatically format code.
> Run linters to catch syntax or style issues.
> This ensures that no code can be committed unless it meets the project's standards.

### Acceptance Criteria
- [ ] Investigate and potentially fix our git hooks for formatting, linting, and testing using Husky.

stephenkilbourn:
**Related Ticket:** Close #1329 

### Description of Changes
It seems that there are some changes in both Husky and Yarn in how to make them play nicely together. Husky needs to be run once for each user before the package begins enforcing the git hooks.  When it does that, you should see a new directory `.husky/_`.   For `npm` this is done through a prepare script.  Yarn doesn't support this, so Husky recommends a `postinstall` script. https://typicode.github.io/husky/how-to.html#manual-setup

I've added that script, as well as added a pre-push hook to run the type check and unit tests.  That's easy to remove if contributors prefer pushing separate commits to fix/add tests, though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -259,6 +259,5 @@
     "$context": "~/app/scripts/context",
     "$data-layer": "~/app/scripts/data-layer",
     "$test": "~/test"
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "lint": "yarn lint:scripts && yarn lint:css",
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
+    "postinstall": "husky",
     "ts-check": "yarn tsc --noEmit --skipLibCheck",
     "test": "jest",
     "pretest:e2e": "node test/playwright/generateTestData.js",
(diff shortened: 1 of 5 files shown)


AliceR (approved):
Thank you @stephenkilbourn ! Both pre-commit and pre-push hook are working for me now 🎉 

This will also close #1263 !

package.json
```diff
@@ -25,6 +25,7 @@
     "lint": "yarn lint:scripts && yarn lint:css",
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
+    "postinstall": "husky",
```
hanbyul-here, line 28: This will run after the install of veda-ui as a registry, which we don't want it. Do we need this @stephenkilbourn

stephenkilbourn: @hanbyul-here - ah, I didn't realize that was going to be a consideration.   Looking further at the [husky docs](https://typicode.github.io/husky/how-to.html#manual-setup) for the setup of yarn, we need that postinstall setup, but we also need to disable it before and after the npm publish setup: 
```
{
  "scripts": {
    // Yarn doesn't support prepare script
    "postinstall": "husky",
    // Include this if publishing to npmjs.com
    "prepack": "pinst --disable",
    "postpack": "pinst --enable"
  }
}
```
i'm adding those in a follow-up PR here: #1384

PR #1351, opened 2024-12-23
chore: Include only lib folder when publishing, Add Dataset and Story data type to export

hanbyul-here:
We are publishing everything in the folder which should not be necessary. Explicitly scope the folder to publish (`files` in `package.json`)

Also add more types to export (Dataset, Story)

### DIFF
--- a/app/scripts/index.ts
+++ b/app/scripts/index.ts
@@ -22,6 +22,7 @@ import type {
   InternalNavLink,
   NavItemType
 } from '$components/common/page-header/types';
+import type { DatasetData, StoryData } from '$types/veda';
 
 import ExplorationAndAnalysis from '$components/exploration';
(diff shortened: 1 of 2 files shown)


dzole0311 (approved):
Tested the publishing of the package, works as expected (the tarball from Verdaccio contains only the lib as a directory) ✅ 

Screenshot 2024-12-24 at 13 06 32

@hanbyul-here The `README.md` and `package.json` that correspond to the veda-ui are also published. Not a blocker, but I guess we should eventually update these as well (e.g. `README.md` specific to the library)?

hanbyul-here:
@dzole0311 Yah good point - we can't separate package.json but we probably can do something better for README before we actually start publishing to npm registry

Issue #1324, opened 2024-12-15
Expose Scrollytellingblock MDX component for template instance


sandrahoang686:
Currently Adding more story content is breaking in this PR https://github.com/developmentseed/next-veda-ui/pull/22.

We need to expose the MDX components that haven't been exposed yet which the stories mdx depends on like `Scrollytellingblock` and `Chapter`. Chapter is exposed but we need to hook it up, Scrolltellingblock still has veda virtual module dependencies which we should decouple from and then expose.

AC:
- [ ] Get this PR https://github.com/developmentseed/next-veda-ui/pull/22 working

PR #1361, opened 2025-01-07
Fix prop type for Map and Scrollytelling

hanbyul-here:
**Related Ticket:** #1324 

### Description of Changes
While contemplating on what @sandrahoang686  mentioned about the data proptype, I realized that our types are wrong for Map and Scrollytelling (And We have different types, but named same of `DatasetData` which we should address soon.) 

Map and Scrollytellinglblock seems to expect the same shape of the dataset which is an object (`VedaData<DatasetData>` - am I missing anything?

### DIFF
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -36,7 +36,7 @@ import {
   getDatasetLayers
 } from '$components/exploration/data-utils-no-faux-module';
 import { useReconcileWithStacMetadata } from '$components/exploration/hooks/use-stac-metadata-datasets';
-import { ProjectionOptions, VedaDatum, DatasetData } from '$types/veda';
+import { ProjectionOptions, VedaData, DatasetData } from '$types/veda';
 import { EnvConfigContext } from '$context/env-config';
 
 export const mapHeight = '32rem';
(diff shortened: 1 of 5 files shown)


sandrahoang686:
Thanks @hanbyul-here for catching this 🙇🏼‍♀️. I pushed up a change though to just rename the type which makes more sense to do. So the type could be more informative instead of just giving it a new type alias in different files. I'm going to approve and merge in. 

Also, We are going to need to do a general cleanup by the way between these two files because the `/exploration/types.d.ts.ts` references some types in `$types/veda.ts`. For example `EnhancedDatasetLayer` is duplicated between exploration types [here](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/types.d.ts.ts#L88-L91) and veda types [here](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/types/veda.ts#L276-L279). Smalls things like `ParentDatset` is mispelled so we should probably just take a good sweeping look between the two.

We also should not be casting [here](https://github.com/NASA-IMPACT/veda-ui/blob/db6eff4ffcb2a2188e1a21460477916db54893c5/app/scripts/components/common/blocks/scrollytelling/index.tsx#L167). Casting tends to mask a deeper issue with our types and having bad types or mis-transformations which means we should revisit our types here again.

Created ticket for the tech-debt here => https://github.com/NASA-IMPACT/veda-ui/issues/1363

Issue #1344, opened 2024-12-19
Have Card Component adopt ConfigProvider for navigation


sandrahoang686:
As the outcome from this PR https://github.com/NASA-IMPACT/veda-ui/pull/1338. We want to retain the LinkComponents for accessibility reasons instead of using a callback. We should update the Card component to adopt the ConfigProvider hooks to handle navigation. 

Acceptance Criteria:
- [x] Data Catalog should be free of passing in LinkProperties
- [x] StoriesHub should be free from passing in LinkProperties

vgeorge:
Status update: in #1374 we partially removed `linkProperties` from CatalogView, but component `FeaturedSliderSection`, used by the catalog, still uses the deprecated approach. I believe we need to address this in this ticket.

vgeorge:
A possible side effect of these changes, raised by @aboydnw, is that Google Analytics events on instances relying on click events might not work as before now that cards are links. It would be nice to identify which instances are and track any issues.

PR #1365, opened 2025-01-09
refactor(card): deprecate linkProperties on the Card Component

vgeorge:
**Related Ticket:** #1344

### Description of Changes

- Replace all links in the Card with the Link component from Veda UI provider
- Updated type definitions to ensure `to` and `onClick` cannot be present simultaneously
- Added JSDoc comments
- Included a warning for deprecated props `linkTo` and `linkProperties`
- Handled non-interactive card case where neither `to` nor `onClick` is provided

The main focus of this change was to use the new Link component from the provider but to keep backwards compatibility with existing implementations.

### Validation / Testing

I suggest visiting E&A and Stories page and confirm the previous behaviour hasn't changed. Another thing is to check if a lint warning is shown when trying to using `to` and `onClick` properties at the same time on a Card component.

This is ready for a review.

### DIFF
--- a/app/scripts/components/common/card/index.tsx
+++ b/app/scripts/components/common/card/index.tsx
@@ -64,15 +64,10 @@ interface ClickCardComponentProps extends BaseCardComponentProps {
   onClick: MouseEventHandler;
 }
 
-interface NonInteractiveCardComponentProps extends BaseCardComponentProps {
-  to?: never;
-  onClick?: never;
-}
-
 export type CardComponentProps =
   | LinkCardComponentProps
   | ClickCardComponentProps
-  | NonInteractiveCardComponentProps;
+  | BaseCardComponentProps;
 
 export interface DeprecatedCardComponentProps {
   linkProperties?: LinkProperties & { linkTo?: string };
@@ -318,14 +313,14 @@ function CardComponent(
     parentTo,
     footerContent,
     hideExternalLinkBadge,
-    onCardClickCapture,
-    onClick
+    onCardClickCapture
   } = props;
 
   const { Link } = useVedaUI();
 
   // For backwards compatibility with deprecated props
-  const to = props.to || props.linkTo || props.linkProperties?.linkTo;
+  const to =
+    ('to' in props && props.to) || props.linkTo || props.linkProperties?.linkTo;
 
   if (props.linkProperties || props.linkTo) {
     // eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


app/scripts/components/common/card/index.tsx
```diff
@@ -390,6 +397,54 @@ function CardComponent(props: CardComponentPropsType) {
...
+  // Click Card
+  if (onClick) {
+    return (
+      <ElementInteractive
+        as={CardItem}
+        cardType={cardType}
+        className={className}
+        linkLabel={linkLabel ?? 'View more'}
+        onClickCapture={onCardClickCapture}
+        onClick={onClick}
```
dzole0311, line 431: Might be better to extract the shared / duplicated props into a `const` (eg `sharedProps`) and spreading the values in the different card variations

dzole0311 (changes requested):
Found an issue when clicking on external link cards under `/stories`:

1. Visit `https://deploy-preview-1365--veda-ui.netlify.app/stories`
2. Click on `External Link Test` in the featured stories slider
3. The external link seems to be appended to the url

vgeorge:
@dzole0311 thanks for the review! I made a change to the provider to render `<a>` when `to` is an external link and refactored the card variations to avoid props repetition. I still kept returning linked, clickable, and non-interactive variants separately, as I believe it is more readable that way.

This is ready for another review.

sandrahoang686 (changes requested):
The ticket also deals with getting rid of the linkProperties in the hub and catalog component. In `hub-content` we no longer need to pass in `linkProperties` and can just pass in `to` with your changes 🙌🏼 . But Catalog, we would need to reverse the navigation callback. Should be easy to do but also feel free to do it in an iteration after this PR but we should update the ticket to reflect that.

_validation note:_ External link is no longer opening in a new tab which is how it was originally behaving before

dzole0311 (approved):

vgeorge:
@sandrahoang686 Thank you for the review. I made the following changes:

- External links now open in a new tab when using the Veda UI provider
- Removed `NonInteractiveCardComponentProps` in favor of `BaseCardComponentProps`. There’s likely a better way to handle type checking than what I did [here](https://github.com/NASA-IMPACT/veda-ui/blob/f1a065d1dd18e91534e489a17e475e49becfd154/app/scripts/components/common/card/index.tsx#L422), but we can improve it later if it becomes a blocker

In this PR, I aimed to keep the scope limited to the Card component itself for easier review. I'll address the remaining work from #1344 in a separate PR.

Issue #1360, opened 2025-01-07
Streamline Package Release, Code cleanup and Next.js Deployment related to refactoring


hanbyul-here:
Each of the item can be a separate issue. Using this issue to collect the items. 

### Package Management
- [x] Publish the package through NPM: Currently using a private registry (Verdaccio). Decide whether to continue with Verdaccio or publish to the public NPM registry. @dzole0311 via https://github.com/NASA-IMPACT/veda-ui/issues/1369
- [x] Establish a release pattern: Releases are currently ad hoc with arbitrary suffixes. Consider aligning the package release process with the current general release process or standardizing the versioning pattern. @dzole0311 via https://github.com/NASA-IMPACT/veda-ui/issues/1369

### Code Cleanup
-  [x] Ensure VEDA faux module dependency + dom router dependency doesn't exist in the library: Based on our library building result, something is still depending on VEDA faux module. (This is why the result includes various media from the stories.) Trace down where this dependency happens and remove it. It might be faster to just look up any import from 'veda'   in component folder and replace it.  @hanbyul-here 

I think I might have found the culprit: https://github.com/NASA-IMPACT/veda-ui/blob/552b7b80e1ce9dc91adebd5c3b0837a4d631843a/app/scripts/components/common/page-header/nav/nav-item-cta.tsx#L3 This is basically importing everything from layout-root.

### Type Definitions
-  [x] Export proper types, and use it from instance side: especially  DatasetData and StoryData. (one gotcha: We have two DatasetData types, one exported from E&A and the other exported from VEDA - @sandrahoang686  https://github.com/NASA-IMPACT/veda-ui/pull/1361/files#diff-af7791d70ec6eacf00bc029b1e7b44b287a9d2c6cc737a56acd4d1b20cda8f1cR40 It might be a good chance to clean up) _(Note: Partially addressed in this https://github.com/developmentseed/next-veda-ui/pull/34)_

PR #1367, opened 2025-01-09
chore: Remove VEDA dependency (as much as possible) 

hanbyul-here:
**Related Ticket:** #1360 

### Description of Changes
The reason I started looking at is I noticed that our library build has assets that are not supposed to be there (It started happening from ~ v5.9.0. If you build the library with v5.8, it builds clean.)

All are quite small changes,  (mainly changing the path for imports) + but lint change makes it look big :/

Screenshot 2025-01-09 at 11 21 28 AM

The changes in this PR so far do not fix the problem, but while we are on it  let's clean up 🧹


app/scripts/components/exploration/components/datasets/dataset-list-item.tsx
```diff
@@ -89,6 +90,8 @@ const DatasetData = styled.div`
 `;
 
 interface DatasetListItemProps {
+  // @TODO: rename
+  datasets: TimelineDataset[];
```
sandrahoang686, line 94: what is this note? if we wanna rename the `TimelineDataset interface` maybe we should make this note in the types file

sandrahoang686 (approved):
great cleanup! 🙇🏼‍♀️

app/scripts/components/exploration/components/datasets/dataset-list-item.tsx
hanbyul-here: No I think `TimelineDataset` naming is fine. I honestly can't remember what the context was. removing it.

Issue #1371, opened 2025-01-13
[Bug]Cookie consent form color functionality suddenly failing


snmln:
### Context

URL: GHG center cookie consent edits not functioning correctly. reported by @paridhi-parajuli  

Blocking a release? (y/n): N

Priority: P3

### Steps to reproduce the issue

Editing the colors in the veda.config removes all styling and throws an error. Emotion css library is suddenly failing in browser and not stringifying dynamically assigned classes. 

Remedy: add this line to top of cookie-consent/index.tsx /** @jsxImportSource @emotion/react */

PR #1372, opened 2025-01-13
fix: updating emotion implementation

snmln:
[**Related Ticket:**](https://github.com/NASA-IMPACT/veda-ui/issues/1371) 

### Description of Changes
Updating emotion implementation to address stinginfy errors 

### Validation / Testing
change the settings of the `veda.config.js` `cookieConsentForm: {...}` contents and validate that they correctly update

### DIFF
--- a/app/scripts/components/common/cookie-consent/index.tsx
+++ b/app/scripts/components/common/cookie-consent/index.tsx
@@ -1,13 +1,11 @@
+/** @jsxImportSource @emotion/react */
+
 import React, { useState, useEffect } from 'react';
 import { debounce } from 'lodash';
 import { Icon } from '@trussworks/react-uswds';
 import { css } from '@emotion/react';
 import { setCookie, getCookie } from './utils';
-import {
-  USWDSAlert,
-  USWDSButton,
-  USWDSButtonGroup
-} from '$uswds';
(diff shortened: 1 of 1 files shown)


hanbyul-here:
My apologies that I didn't ask this earlier. Would you tell me why we need emotion/css for Cookie Component? This is only used in this component, and I would like for us to be aware what kind of tools we are using for styles and not to diverge them whenever possible.

snmln:
Emotion was a solution to dynamically set more rich css inline effectively, since hover state and child selectors arent a trust worthy option through standard inline style={} css.

hanbyul-here:
I think there are several things we can go over which might involve a bigger discussion. Again my apologies that I didn't bring this up earlier. 
- We should only allow the Customization based on primary, secondary color (theme).  
- We couldn't figure out how to customize uswds when veda-ui is used as a submodule. So it might be unavoidable that we need to use some css in js solution here, since primary and secondary value are exposed as js. 

Is our system not supporting this way of using emotion? https://emotion.sh/docs/css-prop#jsx-pragma

Also a tip that it makes easy to review especially a hotfix pr like this if you include preview links!

hanbyul-here (approved):

snmln:
I completely agree, I think that starting the conversation around how we would like to adjust is critical as we start to accommodate the instance request for customization.

PR #1384, opened 2025-01-16
chore: disable preinstall before npm publish

stephenkilbourn:
**Related Ticket:**
#1329 #1345 

### Description of Changes
without this, the postinstall will run the husky setup after the install of veda-ui as a registry, which we don't want. Following the [husky docs](https://typicode.github.io/husky/how-to.html#manual-setup) for the setup of yarn, we need that postinstall setup, but we also need to disable it before and after the npm publish setup: 

### Validation / Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,9 +28,6 @@
     "lint": "yarn lint:scripts && yarn lint:css",
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
-    "postinstall": "husky",
-    "prepack": "pinst --disable",
-    "postpack": "pinst --enable",
     "ts-check": "yarn tsc --noEmit --skipLibCheck",
     "test": "jest",
     "pretest:e2e": "node test/playwright/generateTestData.js",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -29,6 +29,8 @@
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
     "postinstall": "husky",
```
hanbyul-here, line 31: Thank you for handling this swiftly. Can you explain me why we need hustky for postinstall command?

stephenkilbourn: It is a quirk to how husky operates. Running that configure command essentially installs a husky script that doesn't actually get committed on the developer’s machine. Another option would just be to tell people as part of the setup to install husky. It's a little more graceful for using npm, but for yarn you have to do this post install.

They talk about it briefly in this issue:
https://github.com/typicode/husky/issues/1016

A third option would be to enforce this with a different package than husky such as https://www.npmjs.com/package/lefthook

AliceR (commented):
Enabling and disabling around the publishing is an interesting approach, it should work. I just wonder if it would be cleaner / simpler to use the `HUSKY=0` env var in this case?https://typicode.github.io/husky/how-to.html#for-a-gui-or-globally
I just stumbled upon it on the docs that you shared, haven't actually tried this approach!

package.json
```diff
@@ -29,6 +29,8 @@
...
+    "prepack": "pinst --disable",
+    "postpack": "pinst --enable",
```
dzole0311, line 33: @stephenkilbourn I’ve added `pinst` as a devDependency here (I was using this branch to unblock some of my work).

@hanbyul-here Any thoughts on adding a dry run for `npm publish` under `yarn stage`?

`"stage": "yarn buildlib && NODE_ENV=staging gulp"`

This would help catch issues like the missing `pinst` dependency here before an actual publish attempt. It’s a rare scenario, but might be worth catching early, especially if we continue automating this further.

dzole0311 (approved):

sandrahoang686:
Can we merge this? Would like to run library builds and publish for use but this is blocking

hanbyul-here (approved):
best way to test it is try it! thanks for working on it @stephenkilbourn

PR #1385, opened 2025-01-16
chore: Expose Audience and ValueProposition for template instance home

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1386
**Related PRs:** https://github.com/developmentseed/next-veda-ui/pull/37

This is just to have template instance match current homepage as it was originally a part of the [Add Content MVP PR](https://github.com/developmentseed/next-veda-ui/pull/22). There will be other iterations as it isn't the final design but this is to have homepage match as much as possible temporily.

**`v5.11.5-1386.1` was built off of this**

### Description of Changes
* Exposes the Audience and ValueProposition components
* Remove husky line based off this [comment](https://github.com/NASA-IMPACT/veda-ui/pull/1345/files#r1917147518). This line makes `yarn install` fail in nextJs build cc @stephenkilbourn 

### Validation / Testing
* Make sure homepage renders as supposed to and everything clickable is working as supposed to (Feedback forum and cards)

### DIFF
--- a/app/scripts/components/home/value-propostion.tsx
+++ b/app/scripts/components/home/value-propostion.tsx
@@ -1,7 +1,5 @@
 import React from 'react';
 import styled from 'styled-components';
-import { getString } from 'veda';
-
 import imgVP01 from '../../../graphics/layout/vp-01-illu.png';
 import imgVP02 from '../../../graphics/layout/vp-02-illu.png';
 import imgVP03 from '../../../graphics/layout/vp-03-illu.png';
@@ -90,8 +88,8 @@ function ValueProposition() {
           <ContentBlockProseAlt>
             <h3>Science communication platform</h3>
             <p>
-              Share your {getString('stories').other.toLocaleLowerCase()} with
-              others through the VEDA Dashboard. Submit a{' '}
+              Share your data stories with others through the VEDA Dashboard.
+              Submit a{' '}
               <a
                 href='#'
                 onClick={(e) => {
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "lint": "yarn lint:scripts && yarn lint:css",
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
-    "postinstall": "husky",
     "ts-check": "yarn tsc --noEmit --skipLibCheck",
     "test": "jest",
     "pretest:e2e": "node test/playwright/generateTestData.js",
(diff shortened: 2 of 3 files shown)


app/scripts/components/home/value-propostion.tsx
```diff
@@ -90,8 +88,8 @@ function ValueProposition() {
           <ContentBlockProseAlt>
             <h3>Science communication platform</h3>
             <p>
-              Share your {getString('stories').other.toLocaleLowerCase()} with
-              others through the VEDA Dashboard. Submit a{' '}
+              Share your data stories with others through the VEDA Dashboard.
+              Submit a{' '}
```
sandrahoang686, line 92: I just hardcoded this.. there is no effect on other instances since other instances do not use this component and override the homepage (GHG & EIC). This still matches what is currently on https://www.earthdata.nasa.gov/dashboard/. Sandbox or veda-ui did original just said `stories` instead of `data stories`.

package.json
```diff
@@ -28,7 +28,6 @@
     "lint": "yarn lint:scripts && yarn lint:css",
     "lint:scripts": "eslint app/scripts/",
     "lint:css": "stylelint 'app/styles/**/**' 'app/scripts/**/*.(js|ts|tsx|jsx)'",
-    "postinstall": "husky",
```
sandrahoang686, line 31: @stephenkilbourn https://github.com/NASA-IMPACT/veda-ui/pull/1345/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R28 I had to remove this line so that our next instance could install this veda-ui library correctly, any risk here?

dzole0311 (approved):
Thanks @sandrahoang686, the previews look good to me.

However, I'm still unclear whether these components belong in the core veda-ui library - they seem more like layout patterns that can be easily composed using the USWDS grid system combined with text/media.

hanbyul-here:
I share the same thoughts with @dzole0311  - in my mind, this is something that each instance handles. Do we need this to be exported from veda-ui?

sandrahoang686:
@dzole0311 nope you are right, i'm only doing this because @aboydnw had [previously attempted to use these as part of home in the template instance](https://github.com/developmentseed/next-veda-ui/pull/22#issuecomment-2555746432) and as the first iteration but we will likely get rid of these whenever new design is complete for home 👍🏼. So really this can kindof be throwaway work but I think its fine since i'm not changing much here but maybe we dont do this at all and go straight to the new home design. cc @aboydnw 🤷🏼‍♀️

PR #1394, opened 2025-01-21
fix: release log generation

hanbyul-here:
### Description of Changes
Tweak configuration of release-it for better changelog generation

This config will generate a changelog like this:
Screenshot 2025-01-21 at 3 41 56 PM

We can't really detail the sections per type because of the error that comes from the plugin: https://github.com/release-it/conventional-changelog/issues/110

### Validation / Testing
😌 put trust on me this time (or create a branch off from this one and trigger release action.)


hanbyul-here:
oh it striked me that this means that there won't be any log generated if there is no feat: nor fix: for the release. lm think about this issue a bit...

Issue #1258, opened 2024-11-15
[Bug] Selecting predefined AOIs in draw mode results in weird behavior


aboydnw:
### Context

URL: [any E&A page](https://staging.earth.gov/ghgcenter/exploration?datasets=%5B%7B%22id%22%3A%22air-sea-co2%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%5D%2C%22colorMap%22%3A%22bwr%22%2C%22scale%22%3A%7B%22min%22%3A-0.0007%2C%22max%22%3A0.0002%7D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2022-12-01T08%3A00%3A00.000Z&aois=%5B%5D&selectedForEditing=true&dateRange=)

Blocking a release? (y/n): n

Regression issue? (y/n): n
I suspect it has been live for a while

Priority: P3

### Steps to reproduce the issue

1. Open E&A and load a layer
2. Click the draw icon, but don't draw anything
3. Click the list of predefined AOIs and select an AOI
4. Notice that the draw icon is still selected
5. Click the draw icon to deselect

https://github.com/user-attachments/assets/a1b52279-bada-4b3c-9b36-e5bf56026558

aboydnw:
This particular error is impossible in #1395 because the option for pre-selected AOIs is not present when in draw mode. 

Just to be safe, I tested the opposite a few times, having a preselected AOI or an uploaded shapefile, and then replacing with a drawn AOI. It worked without any issues.

I think we can close this when #1395 is merged.

Issue #1093, opened 2024-08-06
[Bug] AOI visualization disappears when map recenters in E&A


aboydnw:
### Context

From @sandrahoang686 's initial investigation:

> in veda-ui, before we were using deprecated staging endpoints (aka [delta-backend.com](http://delta-backend.com/)) this has been updated to the new staging openveda endpoint. But before in veda-ui, actually the layer visualization wasn't working so you were able to see the AOI. After updating to the new staging endpoint, veda-ui preview layers visualization is now working.

URL: 

Blocking a release? (y/n): n

Regression issue? (y/n): y
[If yes, add Regression label]

Priority: P2

### Steps to reproduce the issue

1. Select some layers for visualization in E&A
2. Using the pen tool, draw an AOI that is relatively small for the current zoom level
3. Click Run Analysis
4. Upon map recenter, you should not see the AOI drawn on the map anymore, although it still exists and is selected
5. If you can still see the AOI drawn on the map, try some other combination of AOI size and zoom level. Anecdotally, I can recreate this ~70% of the time

https://www.loom.com/share/39b8d68a331443629d86910f72272614?sid=20d4ea43-5a7a-48aa-bbe9-36b51489f2d7

aboydnw:
similar to #949

AliceR:
I faced the same issue with these steps:
1. load exploration page from fresh start
2. draw big area, leave in selected mode
3. click run analysis
4. -> drawn area disappears from map
5. click pen or anywhere on the map
6. -> new area appears 

Alternatively, in case the area remained:
7. draw another area next to it
8. click apply
9. -> new area disappears, old area is being displayed as selected
10. click pen or anywhere on the map
11. -> new area appears  

And finally, to add some more buggy state issues:
1. load exploration page from fresh start
2. draw and analyze some areas
3. delete all areas
4. draw new area, leave in selected mode
5. click apply
6. -> old areas appear instead of new one! (expected: new area remains)
7. click pen or anywhere on the map
8. -> new area appears 
9. click anywhere again
10. -> old area disappears

aboydnw:
Probably want to address with #1289

aboydnw:
I tested this in #1395 and could not reproduce the error anymore

I tested different zoom levels and locations on the map and everything seemed to work properly. I think we can close this on alongside #1395

Issue #949, opened 2024-05-10
Geometries sometimes do not render across page reloads


dzole0311:
There's an odd behavior that occurs sometimes across multiple page reloads on the exploration page where the AOIs do not seem to be shown on the map. First I suspected that this was caused by the introduction of the static_mode, but after some debugging, I reverted all of the changes made in [936
](https://github.com/NASA-IMPACT/veda-ui/pull/936/files) and the issue still seems to persist.

https://github.com/NASA-IMPACT/veda-ui/assets/13894251/90216904-8b5b-404d-a068-dec66536aa59

j08lue:
I also noticed this issue recently. 👍

aboydnw:
I think we hold off on this one unless it becomes more prevalent. Feels like a rabbit hole that is only worth going down if the severity and frequency is high

aboydnw:
I tested this in #1395 and it seems to work as expected

Before confirming the area, the AOI does not appear on reload. This seems as expected.

After confirming the area, I could not reproduce this error, either before clicking run analysis or after running analysis. Different zoom levels and locations on the map did not seem to have any effect, even though it did before. I think we can close when #1395 is merged.

PR #1395, opened 2025-01-22
feat: Simplified AOI Creation ✨

### LINKED ISSUE #1289: Implement Simplified AOI Creation in E&A
### Context

Followup on #1249 

Goal is to simplify the technical implementation and user experience for AOI creation.

User stories for the three input scenarios:

---

### **1. Draw an AOI**
**As a user**,
I want to draw a custom area of interest (AOI) on the map,
So that I can define a specific geographic area to use in my analysis.

**Acceptance Criteria:**
- [ ] A "Draw AOI" button allows users to enter a distinct drawing mode.
- [ ] While in drawing mode, users can draw a polygon on the map by clicking to define vertices and double-clicking to complete the shape.
- [ ] An option to edit the drawn AOI is available before leaving drawing mode.
- [ ] Once the AOI is completed, users can confirm the drawn polygon to make it the active AOI.
- [ ] Confirming the polygon as AOI exits drawing mode and updates the URL (hashed) for sharing with collaborators.
- [ ] The active AOI is displayed on the map and is ready to trigger the analysis.
- [ ] The drawing mode can be cancelled without saving changes, leaving any prior defined AOI unchanged.
- [ ] Users can clear the active AOI to revert to no defined area.

---

### **2. Upload a GeoJSON File**
**As a user**,
I want to upload a GeoJSON file that defines an AOI,
So that I can use predefined boundaries from external sources in my analysis.

**Acceptance Criteria:**  
- [ ] A "Upload AOI" button opens a file selection dialog allowing users to select a GeoJSON file.
- [ ] The system validates the uploaded file to ensure it contains a valid polygon or multipolygon geometry.
- [ ] An error message is displayed if the file is invalid, specifying what needs to be corrected.
- [ ] If valid, the AOI is displayed on the map as the active AOI for analysis.
- [ ] The uploaded AOI automatically updates the URL to include its (hashed) coordinates for sharing with collaborators.
- [ ] Users can clear the uploaded AOI to revert to no defined area.

---

### **3. Select a Pre-defined AOI**  
**As a user**,  
I want to select a pre-defined AOI, such as a state or national boundary,  
So that I can quickly use commonly recognized areas in my analysis.

**Acceptance Criteria:**
- [ ] A "Select AOI" dropdown menu lists available pre-defined AOIs, such as state or national boundaries, grouped by category.  [^1]
- [ ] Selecting an AOI highlights it on the map and makes it the active AOI for analysis.
- [ ] The selected AOI automatically updates the URL to reflect its parameters for sharing with collaborators.
- [ ] Users can clear the selected AOI to revert to no defined area or choose a different predefined AOI.
- [ ] The system ensures the selected AOI is ready for use in analysis.

[^1]: Ideally, a user can select the boundaries directly from the map. The menu can be replaced with map selection later in a separate step, as the user flow will remain the same.

-------

Related Issues:

- https://github.com/NASA-IMPACT/veda-ui/issues/949
- https://github.com/NASA-IMPACT/veda-ui/issues/950
- https://github.com/NASA-IMPACT/veda-ui/issues/1093
- https://github.com/NASA-IMPACT/veda-ui/issues/1258

AliceR:
**Related Ticket:** Close #1289
### Description of Changes
This update simplifies AOI management by removing the concept of multiple AOIs where only some are selected for analysis. Instead, there will always be a single AOI that serves as input for analysis. The AOI is now displayed using a regular LineLayer with a GeoJson Source (both react-map-gl components). 
There are three methods for creating an AOI: uploading a GeoJSON file, selecting a predefined boundary from a menu (or directly from the map in a future iteration), and using the Mapbox Draw Control. By leveraging the draw control solely for generating polygons, we eliminate the need for extensive custom code to sync with the library. This simplification addresses long-standing technical debt (close https://github.com/NASA-IMPACT/veda-ui/issues/710) and resolves many issues (fix https://github.com/NASA-IMPACT/veda-ui/issues/949, fix https://github.com/NASA-IMPACT/veda-ui/issues/1093, fix https://github.com/NASA-IMPACT/veda-ui/issues/1258, close https://github.com/NASA-IMPACT/veda-ui/issues/950).

A couple of things I want to do or consider before merging:

- Remove concept of "selected" or multiple aois. I left most of the logic around syncing with hashed urls untouched, but I think this would benefit from some cleanup, now what we have only ever one (Multi-)Polygon AOI.
- Rethink the analysis flow. Apply? Exit? Does this flow still makes sense? Should this Analysis Message rather be moved outside of the map to visually include the timeline (as it does)?
- [x] Upgrade mapbox and replace flyTo with fitBounds, now that they have various projections supported  🕳️ 🐇 lots of subsequent changes to our typings are required to upgrade mapbox > v3.5, and only in [v3.7](https://github.com/mapbox/mapbox-gl-js/releases/tag/v3.7.0) the fitBounds feature was improved for the alternative projections. https://github.com/NASA-IMPACT/veda-ui/pull/1400 
- Write tests 😬

### Validation / Testing
Make sure we can tick all the boxes listed in the user stories and acceptance criteria at #1289. 
Did I miss anything that is not listed?
Did I break any other places where the map components are used, e.g. compare?
Did I actually solve all the bugs listed above?

--- 
This is a big PR. Please review commit by commit - due to prettier formatting [many changes](https://github.com/NASA-IMPACT/veda-ui/pull/1395/commits/cd094af6d26bfaf809671d0d2bcb5d151b7c2af0) can be ignored in the review :) 
I am also happy to pair-review with anyone interested!


hanbyul-here:
amazing work 🥳 
I didn't get to the code review part yet, but I wanted to check in one behavior I am seeing. If I am reading right, we not don't support multiple polygons from draw. which I think is fine - but it is not very clear the drawn polygon is going to be removed when I am drawing the second polygon. ( I came back to this pr and read the changelog to see if that is an intended behavior.) do we need to give some differentiation, or visual signal that the existing polygon is going to be gone if there is a new one?

AliceR:
Thanks for pointing this out! I’ll admit I didn’t do a great job clarifying the expected behavior in the [spike](https://github.com/NASA-IMPACT/veda-ui/issues/1249) we did. Our goal right now is to keep the drawing functionality as simple as possible, as we don’t currently see significant value in supporting multiple polygons for scientific use cases. With that in mind, the behavior you’re seeing—where the previously drawn polygon is removed when a new one is started—is intentional.

When we talk about “multiple polygons,” it could mean different things:
A) A multi-polygon, where the total covered area is treated as a single input for analysis. Multi-polygons are supported by our analysis and can be uploaded if created in another tool, but they can’t currently be created via our drawing tool.
B) Multiple distinct polygons on the map, each producing a separate analysis result that can be compared, such as in a choropleth map.

Which of these approaches were you referring to?

If the need arises, it will be straightforward to expand the drawing functionality into a more robust tool, for example, to support multi-polygon creation. The current implementation is encapsulated, so we can iterate on the drawing tool without impacting other parts of the application. The output of the drawing mode is used directly, no matter how it’s created.

Changing the analysis flow to support multiple distinct polygons, up to eventually enabling choropleth mapping, would require some more thought—but it’s a very interesting direction to explore!

Let me know if you think we need to address this behavior further, perhaps by adding a visual signal or confirmation. It’s a great callout!

hanbyul-here:
Oh, I am sorry that I didn't make my point clear. I do think it is valid that we don't support multiple 'drawn polygons' as long as we support them through file upload and presets. What I was referring to is, this specific UX process that when I click on 'drawing' again. It is not clear that the old polygon will be deleted if my I draw a new polygon and I think some visual signal that the old polygon will be gone will help the user. 

Screenshot 2025-01-23 at 8 35 16 AM

AliceR:
I would like to hear @aboydnw 's and @faustoperez 's opinions here, but I just want to note that what you describe has happened before: when users were drawing but then selecting an area from the dropdown, the drawing would clear without warning.

aboydnw:
I think this is good enough for now, and we can iterate on the UX in the future, assuming we will make updates to E&A with our PI objective this quarter: https://github.com/NASA-IMPACT/veda-architecture/issues/556

faustoperez:
Great work Alice!

I agree with what Hanbyul said: that the intermediate step when drawing a second shape is not very clear. Also, we need to consolidate the positions for the `confirm area` + `delete area` as they belong to the same UI logic group 🤔

app/scripts/components/common/map/controls/hooks/use-themed-control.tsx
```diff
@@ -49,9 +48,10 @@ export default function useThemedControl(
         <ThemeProvider theme={theme}>{renderFn() as any}</ThemeProvider>
       );
     }
-  }, [renderFn, theme]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [theme]); // Only re-render if the theme changes. Adding renderFn here would cause infinite loop
```
hanbyul-here, line 52: I thought `renderFn` as a dependency here is what enables dynamic rendering of mapbox control. 

ex. Coords controls should show the dynamic value when panning the map. (The screenshot is from the current [earthdata.nasa.gov/dashboard](https://earthdata.nasa.gov/dashboard/exploration?datasets=%5B%7B%22id%22%3A%22lis-global-da-evap%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%2C%22colorMap%22%3A%22viridis%22%2C%22scale%22%3A%7B%22min%22%3A0%2C%22max%22%3A0.0001%7D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2021-12-01T05%3A00%3A00.000Z))
Screenshot 2025-01-28 at 9 34 54 AM 

But the coordinate control is not showing the dynamic value anymore from [the preview of this pr](https://deploy-preview-1395--veda-ui.netlify.app/exploration?datasets=%5B%7B%22id%22%3A%22casa-gfed-co2-flux-hr%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%2C%22colorMap%22%3A%22blues%22%2C%22scale%22%3A%7B%22min%22%3A0%2C%22max%22%3A0.3%7D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2017-12-01T05%3A00%3A00.000Z).
Screenshot 2025-01-28 at 9 38 00 AM


Can we rethink the approach here?

AliceR: Oh, good catch! You are right, the Coords Control should update. It does it at times, for me... 
The reason I removed `renderFn` from the dependency array was that I was running into infinite rendering loops that I could pin down to that component. 

I need to better understand `useThemedControl`, will investigate!

app/scripts/components/common/map/controls/hooks/use-draw-control.ts
```diff
@@ -0,0 +1,83 @@
...
+ * const [drawing, isValid] = useDrawControl({ mapboxMap, isDrawing });
+ */
+export const useDrawControl = ({
+  mapboxMap,
+  isDrawing,
+  styles
+}: DrawAOIProps): [Feature | Feature[] | null, boolean] => {
+  const [{ drawing, isValid }, setDrawing] = useState<DrawControlState>({
+    drawing: null,
+    isValid: false
```
hanbyul-here, line 34: Real nitpick: It could be because I am so used to the term 'validation of' geometry, but this variable name throws me slightly off. As far as I understand, this flag is an explicit variable to save whether the drawing feature is null or not, which I don't think my brain clicks to 'valid' status.

app/scripts/components/common/map/hooks/use-layer-interaction.ts
```diff
@@ -36,5 +35,6 @@ export default function useLayerInteraction({
       mapInstance.off('mouseenter', layerId, onPointsEnter);
       mapInstance.off('mouseleave', layerId, onPointsLeave);
     };
-  }, [layerId, mapInstance, onClick]);
-}
\ No newline at end of file
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [layerId]); // do not include mapInstance to avoid unintended re-renders
```
hanbyul-here, line 39: `onClick` of this line and  `onStyleUpdate` of https://github.com/NASA-IMPACT/veda-ui/pull/1395/files#diff-91147ba31615cf8c97badd4f2294836992b581298e966ecd9a4fb3039044a348R176

Are they safe to be removed?

hanbyul-here (commented):
Yay, thanks for tackling aoi issue again.

I understand that you didn't want to touch atom part too much that could mess the url, but I think we can at least remove the methods that are not used anymore for clarification ex. `selectedForEditingAtom`.  (I still owe a thought on this issue: https://github.com/NASA-IMPACT/veda-ui/issues/710 will comment about it later.)

app/scripts/components/common/map/controls/hooks/use-themed-control.tsx
AliceR: Resolved with https://github.com/NASA-IMPACT/veda-ui/pull/1414

app/scripts/components/common/map/utils.ts
```diff
@@ -215,7 +215,7 @@ export function resolveConfigFunctions(
 }
 
 export function toAoIid(drawId: string) {
-  return drawId.slice(-6);
+  return drawId.toString().slice(-6);
```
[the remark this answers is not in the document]
AliceR: The id can be a number when we use a geojson from different sources (i.e. upload). 

I don't know what the history of this function is, and why it was needed anyways. Do you know the reason we only want the last 6 letters of the id?

app/scripts/components/common/map/controls/hooks/use-aois.ts
```diff
@@ -1,71 +1,41 @@
...
+  aoiDeleteAllAtom,
   aoisUpdateGeometryAtom,
   isDrawingAtom
 } from '../aoi/atoms';
-import { SIMPLE_SELECT } from '../aoi';
 
 export default function useAois() {
   const features = useAtomValue(aoisFeaturesAtom);
 
+  const aoi: Feature | null = features.slice(1).reduce((acc, feature) => {
```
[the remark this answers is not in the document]
AliceR: Thats one of my leftover thoughts in the PR description: Remove concept of "selected" or multiple aois. I left most of the logic around syncing with hashed urls untouched, but I think this would benefit from some cleanup, now what we have only ever one (Multi-)Polygon AOI.

Do you think its worth investing the cleanup time now? I wanted to get feedback on the general approach first.

app/scripts/components/common/map/controls/hooks/use-draw-control.ts
```diff
@@ -0,0 +1,83 @@
...
+      }); // init drawControl
+
+      mapboxMap.addControl(drawControl.current);
+
+      mapboxMap.on('draw.create', (e: { features: Feature }) => {
+        // Fired when a feature is created.
+        // https://github.com/mapbox/mapbox-gl-draw/blob/main/docs/API.md#drawcreate
+        setDrawing({ drawing: e.features, isValid: true });
+      });
+      mapboxMap.on('draw.update', (e: { features: Feature }) => {
```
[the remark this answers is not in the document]
AliceR: Yes, `draw.update` is fired when we edit the aoi - which is possible while in drawing mode, before confirming the shape.

app/scripts/components/common/map/controls/hooks/use-draw-control.ts
AliceR: I am using this flag to enable the 'confirm area' button after a polygon has been drawn. Confirming is not possible while still drawing the shape.

app/scripts/components/common/map/hooks/use-layer-interaction.ts
AliceR: Can you explain why you think it should be included in the dependency array? How/when do those functions change? 🤔

hanbyul-here: my thought process is : The `onClick` prop is passed from both `vector-timeseries` and `raster-timeseries`. In vector-timeseries, the value of `onClick` changes based on minZoom. If `onClick` is not included in the dependency array of a hook, the hook may not reflect the updated value of onClick when minZoom changes.

However, the more important point is that we shouldn't have to track every dependency like this manually. The standard rule—at least as enforced by our ESLint configuration—is to include all necessary dependencies in the array, only making exceptions when necessary. But we also have to be flexible and ensure our tool reflects our needs - do you see the need to mend our es-lint configuration? or is this an exception that we need to make?

app/scripts/components/common/map/controls/hooks/use-draw-control.ts
hanbyul-here: Yah, I found that it is used there later. (I edited my comment later in case you saw the outdated one 😓 ) my comment is more about whether the variable can be named differently.

app/scripts/components/common/map/utils.ts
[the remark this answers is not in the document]
hanbyul-here: I do not know! But it sounds like type is wrong then!

app/scripts/components/common/map/hooks/use-layer-interaction.ts
AliceR: I disabled eslint as a shortcut to keep this PR from getting too bloated, but I’m happy to revisit those changes in a follow-up!

The alternative would have been rewriting even more sections of the code. Passing objects or functions as dependencies can be an anti-pattern—“If your Effect depends on an object or a function created during rendering, it might run too often.” ([React Docs](https://react.dev/reference/react/useEffect#removing-unnecessary-function-dependencies)).

I don’t think the definition of onClick actually changes when only its value updates, so I’m not sure it would even trigger an effect. Debugging what’s causing an update can be really tricky—how would I know about minZoom affecting things down the road?

app/scripts/components/common/map/utils.ts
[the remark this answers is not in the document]
AliceR: Not a type issue, according to the [GeoJSON Specification](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2) we have to account for both cases:

> If a Feature has a commonly used identifier, that identifier
> SHOULD be included as a member of the Feature object with the name
> "id", and the value of this member is either a JSON string or
> number.

And [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) only works on arrays and array-like objects, not numbers.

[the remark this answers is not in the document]
hanbyul-here: on line 217, drawId's type is explicitly defined as string though?

app/scripts/components/common/map/utils.ts
```diff
@@ -215,7 +215,7 @@ export function resolveConfigFunctions(
 }
 
 export function toAoIid(drawId: string) {
```
hanbyul-here, line 217:
```suggestion
export function toAoIid(drawId: string | number ) {
```

app/scripts/components/common/map/hooks/use-layer-interaction.ts
AliceR: It seems that `minZoom` is defined in the dataset parameters and won't dynamically change. The value passes several memoized `useCallback` or `useMemo` on the way to the `onClick`. The onClick handler itself is wrapped in a `useCallback`... I believe this is safe to remove, but please let me know if you find a problem!
Same with `onStyleUpdate` in `app/scripts/components/exploration/components/map/index.tsx`, this is wrapped in a `useCallback` as well... 
I see that there is a lot going on, and it is difficult to know what we are breaking. But I would take the chance here and prefer cleanup over fear of breaking things.

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  const onPresetConfirm = (features: Feature<Polygon>[]) => {
+    aoiDeleteAll(); // delete all previous AOIs but keep preset selection
+
+    updateAoi({ features });
+  };
+
+  const onTrashClick = () => {
+    resetAoi();
+  };
```
sandrahoang686, line 89: what is the point of having this? cant we just call `resetAoi()` directly?

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
```diff
@@ -0,0 +1,232 @@
...
+      )}
+
+      <CustomAoIModal
+        revealed={aoiModalRevealed}
+        onConfirm={onUploadConfirm}
+        onCloseClick={() => setAoIModalRevealed(false)}
+      />
+    </>
+  );
+}
```
sandrahoang686, line 205: ⛏️ 🧹  : Further cleanup can be done by breaking these out for example: drawingMode and nonDrawingMode

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  const control = useThemedControl(
+    () => <AoiControl mapboxMap={main} disableReason={disableReason} />,
+    {
+      position: 'top-left'
+    }
+  );
+
+  return control;
+}
```
sandrahoang686, line 232: ❓ : What is this wrapper abstraction and is it really needed or can it just be a part of the `AoiControl` component? If we keep it, can we give it a more helpful name like `AoiControlWithThemedControl`?

sandrahoang686 (commented):
Thanks for working on this 🙌🏼 ! I already think this is a significant improvement and it looks like even more that will come iteratively. I made a few comments/questions but none blocking. After Hanbyul's questions/concerns are addressed, I think its good to merge as a first iteration with next follow-ups which I wanted to ask if you have listed? Have the follow up tickets been created? Where do we go from here after this is merged?

_Another note:_ `eslint-disable-next-line react-hooks/exhaustive-deps` is used quite a bit in this PR for the dependency arrays which like we talked about this morning, tends to signify some type of memory leak or other unwanted side effects that can happen. But having this `eslint-disable...` as a quick solution if we only want to render something on initial only is understandable especially since we do it in other places in dashboard. This isn't ideal though and especially if we use it so much. Ideally, the workaround this is usually adding all needed dependencies but conditionally exiting in the useEffect right away before any logic is executed. Could we make a ticket to clean this up?

aboydnw (approved):
From a UX workflow standpoint, I think we can continue to refine this in the future. For now, I think it is an improvement. 

I received this feedback from Jeanné on the two different options:

> I liked that I’m asked to confirm my area, but beyond that if I change my mind and delete the area to start over, I’m not noticing any difference between the two

Which, matches my gut sense about science users. I personally like fewer buttons, but I see science users liking some more clear confirmation actions. 

In my opinion, we should go forward with this option and drop the experiment in https://github.com/NASA-IMPACT/veda-ui/pull/1406

I think there is something there to use in a next iteration of this flow, but with open questions on whether it should be delete, or something else, and this feedback from Jeanné, I'd propose moving forward with the flow in this PR (1395)

(obviously my review was just on the workflow and not the code itself)

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
AliceR: This wrapper is a leftover from old code, and I wasn’t sure why it was originally needed. I renamed it to 'Wrapper' in a [previous commit](https://github.com/NASA-IMPACT/veda-ui/pull/1395/commits/251897de81d108fd96e6dc243e4d17f5a29d25a2) for clarity, as it only serves to wrap the main component. Since it's used internally, I felt a generic name was appropriate, but I see how a more meaningful name could help in the react component tree. That said, I think the best approach is to refactor and remove the extra wrapper altogether, making the structure cleaner.

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
AliceR: I like that suggestion, this component is huge and getting difficult to read with the conditionals!

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
AliceR: This is originally a leftover artifact from old code, but I find it useful to keep event handlers in one place, for better organization and overview.
That said, I’m happy to inline it if you think it would improve clarity.

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
AliceR: Addressed with 9d0c45d0ffdfa9a38dcad0861b52f84a9888eace

app/scripts/components/common/map/controls/aoi/aoi-control.tsx
AliceR: Addressed with 027d5c5092d18b207478af9a9198dc650890980b

Would you extract even more? Personally I think this should do for now, we are going to rewrite a lot of the UI buttons anyways with the redesign.

app/scripts/components/common/map/controls/hooks/use-aois.ts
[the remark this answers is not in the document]
hanbyul-here: I asked since this code looks like a new addition for this PR (not carrying over the past code - or am I missing anything?) What case does this code serve right now?

app/scripts/components/common/map/hooks/use-layer-interaction.ts
hanbyul-here: Oki, then let's find a way of capturing the reasoning of excluding dependencies.

hanbyul-here (approved):
🙇 thanks for working on it, let's keep the discussion going but I do think we can ship this.

PR #1398, opened 2025-01-22
chore: Remove veda dependency

hanbyul-here:
**Related Ticket:** #1360 

This PR removes VEDA dependency from the registry build. Run `yarn buidlib`locally, see how less number of assets it builds to test it. (PageHeader and PageFooter were the root of the cause)

Screenshot 2025-01-22 at 3 33 43 PM

Compare it to : https://github.com/NASA-IMPACT/veda-ui/pull/1367

It is kind of difficult to think what kind of test we can do check this, but it will be nice to have some automatic checks - since it is much easier to get rid of veda dependency when you know what directly introduced it.


app/scripts/utils/use-feedback-modal.tsx
```diff
@@ -0,0 +1,20 @@
+import { useCallback } from 'react';
+import { useVedaUI } from '$context/veda-ui-provider';
+/**
+ * Hook to access the feedback modal.
+ */
+export function useFeedbackModal() {
+  const { feedbackModalRevealed, setFeedbackModalRevealed } = useVedaUI();
```
hanbyul-here, line 7: We cannot use layout provider, so I added these to VEDA UI Provider. but if there is any better place. willing to move this.

sandrahoang686 (changes requested):
I don't think VedaUIProvider is the best place, this provider has been a workaround for us to handle env vars and workarounds for framework related things (routing). It would be great if we could keep the usecase for it limited. 

I also think we dont need it as part of context and could just isolate to the hook. It looks like pretty simple logic where all its being used for is to just show/not show something so this could be entirely simplified. Really, we dont even need the hook but [created a PR to this branch here](https://github.com/NASA-IMPACT/veda-ui/pull/1399) to make said changes if you could take a look (this should also fix the tests because no more dep on provider).

sandrahoang686 (approved):

PR #1399, opened 2025-01-22
fix: Isolate changes to just hook and rename for general use

sandrahoang686:
**Changes:**
https://github.com/NASA-IMPACT/veda-ui/pull/1398#pullrequestreview-2568480514

Screenshot 2025-01-22 at 6 30 42 PM


hanbyul-here:
Yay thank you~ (just heads-up that there are many other places using useFeedbackModal from layout-root in case if you want to consolidate them later.)

Issue #1363, opened 2025-01-08
[v7] Review and Clean up Exploration types & Veda Types for Dataset Layers


sandrahoang686:
**Context:**
Related to comment from PR here => https://github.com/NASA-IMPACT/veda-ui/pull/1361#issuecomment-2578709718

We need to review the data types between the [/exploration/types.d.ts file](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/types.d.ts.ts) and the [/types/veda.ts file](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/types/veda.ts). There is duplication happening. We should also check the files that references these types. For example there is typecasting happening which looks to be masking an issues of how we may be defining these types or some mis-transformations/reconciling along the way like here => https://github.com/NASA-IMPACT/veda-ui/blob/93ae948ca395c4ff1c9882d7dd01a9912ec0e7f7/app/scripts/components/common/blocks/scrollytelling/index.tsx#L165-L167

aboydnw:
@sandrahoang686 is this ticket still valid? If so, how urgent is it that we address?

sandrahoang686:
@aboydnw we should keep it, its tech-debt but not urgent. Should be addressed at some point for sure though, maybe during a not so heavy sprint.

PR #1400, opened 2025-01-23
fix: Upgrade mapbox-gl

AliceR:
I am attempting to upgrade mapbox-gl, in order to use the fitBounds with alternative projections. Version [3.5](https://github.com/mapbox/mapbox-gl-js/releases/tag/v3.5.0) introduces breaking changes to the type system, which I was able to address in e227a9d7b7b79fcc40d17b6e93b73e1f4276f071. I have some issues left that I need help with, also need someone with better understanding of our veda typings to have a proper look at my changes. 
@sandrahoang686 maybe? 😊


dzole0311:
@AliceR I think the issue is a mismatch between the `ProjectionOptions` in `veda.ts` and `ProjectionOptions` in `index.d.ts` 🤔  Updating the one in the module definition file (`parcel-resolver-veda/index.d.ts`) seems to resolve the TS errors:

```
  export type ProjectionOptions = {
    id: mapboxgl.ProjectionSpecification['name'] | 'polarNorth' | 'polarSouth';
    parallels?: number[];
    center?: [number, number];
  };
```

AliceR:
Thanks @dzole0311 , this indeed fixed all the ts errors. I am still not sure if I solved all the issues correctly... And now I even wonder why we have this duplication between `app/scripts/types/veda.ts` and `parcel-resolver-veda/index.d.ts`... 🤔

sandrahoang686:
hey @AliceR 👋🏼 . Just saw your question in the description but what could I help out with specifically when it comes to typing? There is lots going on with it and much improvement in our codebase needed for it 😄 . A few things I can summarize right now though that might help out..
1. When it comes to casting, `(someType as otherType)` this generally isn't good practice because it means we are masking ill defined type interfaces or bad transformations or dont have the correct types. There is actually already a good amount of this happening in the veda-ui codebase which has been noted as tech-debt (a tech-debt ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/1363)). Looks like a good amount of casting is going on in this PR which we means we should probably look at the type definitions and clean it up
2. Like Hanbyul mentioned, we want to decouple from veda virtual modules so any types under `parser-resolver-veda/*` we wont want to use and should use `$types/veda`. Most types exist there already I believe, if not, we'd just want to copy them over
3. If we want types to be exposed to the NextJs instance to use, these types of have to defined in a `*.ts` file instead of a `*.d.ts` file. Context from this PR [here](https://github.com/NASA-IMPACT/veda-ui/pull/1223).

lmk if I could help assist in any of these areas :)

app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -206,7 +206,7 @@ function MapBlock(props: MapBlockProps) {
         parallels: projectionParallels
       });
       return {
-        ...projection,
```
AliceR, line 209: `Spread types may only be created from object types.ts(2698)`

app/scripts/components/common/map/style-generators/basemap.tsx
```diff
@@ -115,7 +123,7 @@ export function Basemap({
 
     updateStyle({
       generatorId: 'basemap',
-      sources: baseStyle.sources as Record<string, AnySourceImpl>,
+      sources: baseStyle.sources as Record<string, SourceSpecification>,
```
dzole0311, line 126: Probably fine without any casting?

`sources: baseStyle.sources`

app/scripts/components/common/map/style-generators/basemap.tsx
```diff
@@ -76,15 +82,17 @@ export function Basemap({
...
       GROUPS_BY_OPTION.boundaries,
-      baseStyle.metadata['mapbox:groups']
+      (baseStyle.metadata as Record<string, any>)['mapbox:groups']
     );
 
     const layers = baseStyle.layers.map((layer) => {
-      const layerGroup = (layer as Layer).metadata?.['mapbox:group'];
+      const layerGroup = (layer as LayerSpecification).metadata?.[
+        'mapbox:group'
+      ];
```
dzole0311, line 95: Same, probably fine without any casting?

`const layerGroup = layer.metadata?.['mapbox:group']`

app/scripts/components/common/map/style-generators/basemap.tsx
```diff
@@ -76,15 +82,17 @@ export function Basemap({
     // this id from the list of groups in the metadata section of the style.
     const labelsGroupIds = mapGroupNameToGroupId(
       GROUPS_BY_OPTION.labels,
-      baseStyle.metadata['mapbox:groups']
+      (baseStyle.metadata as Record<string, any>)['mapbox:groups']
     );
     const boundariesGroupIds = mapGroupNameToGroupId(
       GROUPS_BY_OPTION.boundaries,
-      baseStyle.metadata['mapbox:groups']
+      (baseStyle.metadata as Record<string, any>)['mapbox:groups']
```
dzole0311, line 89:
```suggestion
      baseStyle.metadata?.['mapbox:groups']
```

app/scripts/components/common/map/style-generators/basemap.tsx
```diff
@@ -76,15 +82,17 @@ export function Basemap({
     // this id from the list of groups in the metadata section of the style.
     const labelsGroupIds = mapGroupNameToGroupId(
       GROUPS_BY_OPTION.labels,
-      baseStyle.metadata['mapbox:groups']
+      (baseStyle.metadata as Record<string, any>)['mapbox:groups']
```
dzole0311, line 85:
```suggestion
      baseStyle.metadata?.['mapbox:groups']
```

app/scripts/components/common/map/styles.tsx
```diff
@@ -64,10 +68,12 @@ const generateStyle = (
     };
 
     const generatorLayers = generatorParams.layers.map((generatorLayer) => {
-      const metadata: ExtendedMetadata = generatorLayer.metadata ?? {};
-      metadata.generatorId = generatorId;
+      const metadata: ExtendedMetadata = {
+        ...(generatorLayer.metadata ?? {}),
```
dzole0311, line 72:
```suggestion
        ...generatorLayer.metadata
```

dzole0311: Oops, this will only work if we don't pass `unknown` to the `metadata?: ExtendedMetadata`.

PR #1406, opened 2025-01-27
[Do not merge this yet] Hide toolbar when an AOI is created

AliceR:
Based on our discussions around improving the user experience when creating a new AOI after a shape has already been drawn, this PR introduces an alternative user flow:
- Users can only create a new AOI (via selection, drawing, or uploading) after deleting the existing one.
- While an AOI is present, the toolbar for creating new AOIs will be hidden, reducing the risk of unintentional overwrites and clarifying the workflow.

This proposal is a variation (Version B ☺️) of the current approach in #1395 (Version A). It emphasizes deliberate action on the part of the user, ensuring that any new AOI creation starts from a clean slate.

Feedback on whether this alternative flow better aligns with user needs would be greatly appreciated! We could even run some A/B testing with both of the deploy preview links!

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/aoi-control.tsx
@@ -120,14 +120,18 @@ function AoiControl({
 
   return (
     <>
-      <Tip disabled={!disableReason} content={disableReason} placement='bottom'>
-        <div>
+      {!aoi && ( // only show the toolbar when there is no AOI
+        <Tip
+          disabled={!disableReason}
+          content={disableReason}
+          placement='bottom'
+        >
           <AnalysisToolbar
             visuallyDisabled={!!disableReason}
             size='small'
             data-tour='analysis-tour'
           >
-            {isDrawing ? (
+            {isDrawing ? ( // toolbar for drawing mode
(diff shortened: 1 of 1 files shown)


j08lue (changes requested):
I understand the intention to make it explicit to users that they need to delete the current AOI to be able to draw or select a new one.

However, what the user actually wants is more like "Start over" or "Draw or select a new AOI". Not delete. I don't anyone ever only wants "delete" - why would you delete if you do not want to draw or select a new AOI?

So I think the idea is good to make it clear that there can only be one AOI and if you want a different one, you are effectively starting over.

But the delete button will be a blocker, because users are looking to do something else - ultimately the draw/select UI again.

faustoperez (approved):
I agree with Jonas, the presence of the delete area button is still a bit wonky. I don't think we have the flow figured out yet. Looks good though, since this is one side of an A/B test

j08lue:
IMO changing the label on the "Delete" button to "Start over" might already address my main concern, for now. Then we can still revisit the whole controls situation.

AliceR:
Closed as discussed here: https://github.com/NASA-IMPACT/veda-ui/pull/1395#pullrequestreview-2585237595

Note: A redesign of the UI elements is coming soon, so we shouldn't get too caught up on this for now :)

Issue #1402, opened 2025-01-27
Fix Scrollytelling layout in Next instance


AliceR:
## Context

ScrollytellingBlock styling needs to be fixed to match existing, including ensuring all frames populate with the correct layer and basemap. You can see the problem from : https://veda-ui-next-test.netlify.app/stories/air-quality-and-covid-19

## Acceptance Criteria

- [x] Map is fixed and content scrolls over as intended

PR #1391, opened 2025-01-20
feat: Publish Veda UI to npm and workflow proposal

dzole0311:
**Related Ticket:** closes [#1369](https://github.com/NASA-IMPACT/veda-ui/issues/1369)

### Description of Changes
- Changed the registry in .npmrc from the private Verdaccio registry to the public NPM registry under the scope `teamimpact`
- [Published the library](https://www.npmjs.com/~teamimpact?activeTab=packages) under the `teamimpact` scope and updated the docs on how to set the necessary npm credentials
- Experimented with a new `README-lib.md`:
  - Added an experimental workflow to copy this file to the lib dist during the library build
  - This allows us to have a unique `README.md` for the library when its published on NPM, separate from the veda-ui submodule documentation
- I noticed the lib bundle size increased, largely due to the USWDS images and fonts being copied in the library build. There’s room for optimization, including removing unused assets or modules from the existing veda-ui (e.g. `air-quality-and-covid-19--story-cover.png` etc), and I made a tracking issue for that: #1392

### Notes & Questions About Changes

Proposed workflow for developing, testing and publishing the NPM registry:
 
- Workflow proposal for veda-ui NPM publishing and development / testing with consuming apps (Next.JS) is [documented here](https://docs.google.com/document/d/1JRvZybU08mSdZTTXPkJr5pqYgoB_5EYYNWixXlgUIlA/edit?usp=sharing). The proposal aligns and extends our current veda-ui publishing workflow with the bi-weekly automated releases

### Validation / Testing
1. Set the necessary npm auth credentials mentioned in the set up docs
2. dry-run publish a version to make sure everything you're correctly authenticated
3. Review the [Google doc](https://docs.google.com/document/d/1JRvZybU08mSdZTTXPkJr5pqYgoB_5EYYNWixXlgUIlA/edit?usp=sharing) proposal and share feedback

PR #1408, opened 2025-01-28
fix: Scrollytelling layout in Next instance

AliceR:
**Related Ticket:** #1402 

### Description of Changes
Provide a fallback value for top offset of ScrollyMapContainer: When `topOffset` is undefined it causes the css `top` property to be faulty, which makes the `position: sticky` not working.

Do we still need the whole `useSlidingStickyHeaderProps` logic with the USWDS header?

### Validation / Testing
You will to validate in the next-veda instance, making sure the map stays sticky on the top when scrolling down. 
Related PR on next-veda-ui: https://github.com/developmentseed/next-veda-ui/pull/41

### DIFF
--- a/app/scripts/components/common/blocks/scrollytelling/index.tsx
+++ b/app/scripts/components/common/blocks/scrollytelling/index.tsx
@@ -359,7 +359,7 @@ function Scrollytelling(props) {
 
   return (
     <>
-      <ScrollyMapContainer topOffset={topOffset}>
+      <ScrollyMapContainer topOffset={topOffset || 0}>
         {areLayersLoading && <MapLoading />}
 
         {/*


hanbyul-here:
I believe `useSlidingStickyHeaderProps` is not necessary for the new uswds header ( that hook was there becauase we showed/hid the header based on scroll direction, and we ditched that behavior for the new header). The deploy preview from the pr you linked - https://deploy-preview-41--veda-ui-next-test.netlify.app/stories/air-quality-and-covid-19 Scrollytelling block still doesn't work for me. Can you take a look?

AliceR:
Yeah, I noticed the deploy preview isn't updating the library correctly. I need to figure out how to make it use [this branch](fix#1402/scrollytelling-position-sticky). Still not an expert on our verdaccio setup — and [hopefully](https://github.com/NASA-IMPACT/veda-ui/pull/1391), I won't have to become one! 😅

dzole0311 (approved):
Tested in NextJS and works as expected. Thanks for fixing @AliceR

hanbyul-here:
👋 I see a bug that scrollytelling doesn't scroll away even when the scrollytelling section is done (from the next instance preview!) 
https://deploy-preview-41--veda-ui-next-test.netlify.app/stories/air-quality-and-covid-19
Screenshot 2025-01-29 at 8 49 08 AM

AliceR:
Hej @hanbyul-here , thanks for reporting! Do you think the bug was introduced by this change? Or is that a separate issue?

hanbyul-here:
This PR probably didn't introduce the bug, but if its intention was to handle https://github.com/NASA-IMPACT/veda-ui/issues/1402, I think it is the same issue. (Feel free to handle it in whichever way you prefer! I mainly wanted to flag what I found!)

AliceR:
I wasn't aware of this issue when working on #1402 . I created a new one and added your description here: https://github.com/NASA-IMPACT/veda-ui/issues/1418

PR #1414, opened 2025-01-29
fix: Replace root.render in useThemedControl with React Portal

AliceR:
**Related Conversation:** https://github.com/NASA-IMPACT/veda-ui/pull/1395#discussion_r1932291047

### Description of Changes
From the react documentation:
If your app is fully built with React, you shouldn’t need to create any more roots, or to call root.render again.

It is uncommon to call render multiple times. Usually, your components will update state instead.

When you want to render a piece of JSX in a different part of the DOM tree that isn’t a child of your component (for example, a modal or a tooltip), use createPortal instead of createRoot.

How to guide: https://18.react.dev/reference/react-dom/createPortal\#rendering-react-components-into-non-react-dom-nodes

We had to update all cases where useThemedControl was used to render the returned control.

### Validation / Testing
Ensure all map controls are working as expected and do not re-render infinitely.

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/aoi-control.tsx
+++ b/app/scripts/components/common/map/controls/aoi/aoi-control.tsx
@@ -221,12 +221,12 @@ export default function Wrapper({
 }) {
   const { main } = useMaps();
 
-  useThemedControl(
+  const control = useThemedControl(
     () => <AoiControl mapboxMap={main} disableReason={disableReason} />,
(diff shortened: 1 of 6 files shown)


hanbyul-here (approved):
Wahhh, There is official React documentation about implementing the map framework in React. I didn't know! Thank you so much for pointing it out and improving the current hook 🙇

PR #1438, opened 2025-02-04
fix: Empty story  description close #1424

### LINKED ISSUE #1424: [Bug] Catalog search with empty card description
### Context

URL: https://deploy-preview-705--ghg-demo.netlify.app/stories

Blocking a release? (y/n): n

Regression issue? (y/n): n

Priority: P3

### Steps to reproduce the issue

1. Go to stories catalog
2. Enter any search terms in search box
3. See error page

This error page shows here because there are some stories that have nothing entered into the cardDescription field (according to GHG team)
We should probably handle this case, since we split description and cardDescription

hanbyul-here:
**Related Ticket:** #1424 

### Description of Changes
Handle the case when description is empty
Add card description to search criteria
Separate out shared data structure between StoryData and Datasetdata

It might not be very intuitive that DatasetData requires description while StoryData doesn't. but it takes a bigger scope of change if we want to make description optional for DatasetData. (And we might not want it? See the problem of empty description in this comment: https://github.com/NASA-IMPACT/veda-ui/issues/1424#issuecomment-2634734891) 

### Validation / Testing
I removed `description` from life of water story for testing: https://deploy-preview-1438--veda-ui.netlify.app/stories?search=null

### DIFF
--- a/app/scripts/components/common/catalog/prepare-datasets.ts
+++ b/app/scripts/components/common/catalog/prepare-datasets.ts
@@ -24,15 +24,12 @@ export function prepareDatasets(
   options: FilterOptionsType
 ): StoryData[];
 export function prepareDatasets(
-  data: DatasetData[] | StoryData[],
+  data: (DatasetData | StoryData)[],
   options: FilterOptionsType
 ) {
   const { sortField, sortDir, search, taxonomies, filterLayers } = options;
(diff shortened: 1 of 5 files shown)


snmln (approved):
This looks good and cleans up of the previous typing!

PR #1446, opened 2025-02-07
Update sandbox.data.mdx

abarciauskas-bgse:
### Description of Changes
Removed CMIP6 kerchunk demo dataset as titiler-xarray no longer supports kerchunk references (at least not until someone asks for that feature back).

This dataset has already been removed from veda-config.


hanbyul-here:
thank you for handling this 🙇

Issue #1429, opened 2025-02-03
Review subpath routing pattern


AliceR:
Most deployments use subpaths (e.g., /dashboard). Address asset routing issues by establishing a robust subpath handling mechanism in the Next.js instance, replacing reliance on Parcel's baseurl.

snmln:
This appears to be functioning appropriately, moving this to closed

Issue #1415, opened 2025-01-29
[Bug] GHGcenter banner issues


snmln:
### GHGcenter banner icons not functioning correctly

Message context: The banner icons on the GHGC staging environment weren't loading properly when accessed through the /ghgcenter subpath. I suspect the issue might be related to the subpath configuration.
We attempted a quick [workaround](https://github.com/US-GHG-Center/veda-config-ghg/pull/711) by prefixing the paths with /ghgcenter, but this didn't fully resolve the issue—specifically, the US flag icon still didn't show up.

URL: https://github.com/US-GHG-Center/veda-config-ghg/pull/711

Blocking a release? (y/n): n

Regression issue? (y/n): 
[If yes, add Regression label]

Priority: P3

dzole0311:
@snmln Is this only happening when deployed to earth.gov/ghgcenter? Because if i open https://earth.gov/ghgcenter/img/icon-dot-gov.svg, I can see the icon 🤔 

I thought this is already fixed via https://github.com/NASA-IMPACT/veda-ui/pull/1378, preview with the configuration for GHGC with the banner icons showing here: https://github.com/US-GHG-Center/veda-config-ghg/pull/698

cc @slesaad

slesaad:
The files do get copied, but there are two issues:

1. like @snmln mentioned, we tried this workaround where we explicitly prefixed the path with `/ghgcenter` in here; but obviously that's not the best way to do it
2. The workaround worked for the guidance icons, but didn't for the us flag icon (it was still referring to `/img/...` instead of `/ghgcenter/img/..`) probably because of the icon path is hardcoded [here](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/banner/index.tsx#L111)

Issue #1389, opened 2025-01-20
Assess USWDS theming for legacy veda-config instances


dzole0311:
## Context

We have figured out how the USWDS theming will work for the new NextJS-based instances + the veda UI repo for local development.

We also need to assess whether and how USWDS-based theming for veda-ui components can be supported long-term within legacy veda-config instances, when veda-ui is used as a submodule (e.g. the existing `veda-config`).

## Acceptance Criteria

- [ ] Confirm whether USWDS theming is compatible with legacy veda-config instances
- [ ] Add additional configuration in veda-config if required to support USWDS theming
- [ ] Provide documentation detailing the theming setup process for legacy instances

## Related Tickets

[If applicable, link any tickets that are related]

PR #1447, opened 2025-02-10
fix: Make banner flag path and alt text configurable

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1415

### Description of Changes

Make the USWDS banner flag path and alt texts configurable. The USWDS assets are currently resolved incorrectly because they're using absolute paths (starting with '/') which are being resolved relative to earth.gov instead of earth.gov/ghgcenter. As an immediate fix, this PR makes the flag path configurable so we can adjust it based on the deployment context. 

- Path resolution could potentially affect other USWDS assets beyond the flag and in other components. We'll be tackling this via #1429 (subpath route handling) and #1389

### Validation / Testing
- Start veda-ui locally
- Change the `flagImgSrc` path in veda-config.js to `/img/icon-https.svg`
- Restart your server
- Verify that the flag img in the banner shows the new icon
- Change it back the original flag path
- Verify that the flag icon is shown

### DIFF
--- a/app/scripts/components/common/banner/index.tsx
+++ b/app/scripts/components/common/banner/index.tsx
@@ -27,6 +27,7 @@ interface BannerProps {
   headerActionText?: string;
   ariaLabel?: string;
   flagImgAlt?: string;
+  flagImgSrc?: string;
   leftGuidance?: GuidanceContent;
   rightGuidance?: GuidanceContent;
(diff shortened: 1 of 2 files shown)


snmln (approved):
Nice, functions well and meets all testing criteria!

Issue #1435, opened 2025-02-04
Release action follow-up


hanbyul-here:
- [x] Send action failure notice when a scheduled release fails
- [x] Investigate the discrepancy between the production branch and the feature branch: the feature branch succeeds at getting the recommended version from the conventional commit while the main branch fails. I looked into this a bit, and I think it is related to non-linear git history on main. (ex. v5.12.1 and https://github.com/NASA-IMPACT/veda-ui/commit/f146f4506f35f5cebbba842fc97b652358a1a469 will be bumped as a patch, even though the log is prefixed with `feat:`- more findings on the way . it seems it recognizes `feat(scope):` but not `feat:`) I am wondering if it would be better to write a simple version number based on the commit messages 🤔 - I did end up making our own simple plugin using existing tools: https://github.com/NASA-IMPACT/veda-ui/blob/main/recommended-bump/index.mjs )
- [x] Make a better dev environment? - It is challenging with the current plug-in we are using. But maybe..?!
- [ ] set up preview generation action for GHG instance : Since GHG is a separate organization, we need a different approach. I will separate this item

PR #1457, opened 2025-02-11
fix: version bump for release

hanbyul-here:
**This will be worth getting it merged before the scheduled release next week**

**Related Ticket:** #1435 

Problem: Our release hasn't bumped up the version properly even with the rightly formatted conventional commits

### Description of Changes
- Use `'conventional-changelog-conventionalcommits'` package directly to get a recommended bump - I made a small custom release-it plugin and ditched the plugin that I used before.
- I also put a debug flag to make debugging easier (I was editing this in and out to debug so far 😓 )

I am definitely missing something here - because as far as I can read, passing `preset: 'conventionalcommits'` with the plugin I used (release-it-conventionalcommit), should do the same (or more elaborated) job as what this PR is trying to do. But somehow, using a bit of a disassembled approach works while using the preset through the plugin is not working. 

Maybe something to note: using 'angular' preset works 🤔 - I wonder the big body of the commit has anything to the bug. (But angular convention doesn't support `!` exclamation mark as a breaking change)

### Validation / Testing
I diverged this branch from the commit f146f4506f35f5cebbba842fc97b652358a1a469 on purpose because this is the commit that includes `feat:,` which should recommend a minor bump. With the previous setup, the recommended version was patch, but the recommended bump should be minor with this setup. 

You can try this locally by `yarn release --debug`


recommended-bump/index.mjs
```diff
@@ -4,30 +4,24 @@ import semver from 'semver';
...
 // Detect the version number based on conventional commits
 // This does not handle pre-release
 class RecommendedBump extends Plugin {
   async getIncrementedVersion(options) {
+    // loadPreset('conventional') shows different result
     const bumper = new Bumper(process.cwd());
     const conventionalCommitPreset = await createPreset();
-    bumper.commits(null, conventionalCommitPreset.parser);
-
     const recommendation = await bumper.bump(conventionalCommitPreset.whatBump);
```
hanbyul-here, line 14: This line is necessary for the proper bump

dzole0311 (approved):

Issue #1456, opened 2025-02-11
[Bug] Categorical Colormaps Failing


acblackford:
### Context

Datasets with categorical colormap renders are not being displayed in the dashboard. Selecting predefined colormap names for the renders do work, however. This does not throw any errors, and allows for a preview to deploy when working on a config PR, but the dataset will not populate in a story visualization or in the exploration/dataset tabs. 

URL: 

Blocking a release? (y/n): y

Regression issue? (y/n):
[If yes, add Regression label]

Priority: P2

### Steps to reproduce the issue

1. Example sourceParams markdown (also issue in renders):
colormap: '{"6":"#1e90c9", "12":"#ab94d6", "18": "#5a1680", "24": "#d943d4"}'

does not populate the dataset.

Selecting a markdown with colormap name:

colormap_name: blues_r    

works as intended.

hanbyul-here:
Can you share the domplete configuration for this layer? We expect some flags for categorical legend work, and I'd like to check them.

anayeaye:
I have looked at a few different working and not working categorical colormaps. Here are some notes and examples:

**Notes**
There are a couple ways that categorical colormaps are handled: for simple mappings, the colormap can be defined in the sourceParams directly (example: geoglam). For more complicated colormaps that are blocked by the firewall, we have been creating custom named colormaps in the raster-api that are passed as colormap_name in the sourceParams (example: nlcd). In all working examples colormap or colormap_name the raster tiles for a categorical map do not work properly without a categorical legend in the mdx file--I can work up an example to share. 

**Examples**

**collection: geoglam working categorical colormap**

- categorical colormap
- working tile request https://openveda.cloud/api/raster/searches/5eed9169c72f068fa7a9cf9e779b04c3/tiles/WebMercatorQuad/1/0/0?assets=cog_default&colormap=%7B%221%22%3A+%5B120%2C+120%2C+120%5D%2C+%222%22%3A+%5B130%2C+65%2C+0%5D%2C+%223%22%3A+%5B66%2C+207%2C+56%5D%2C+%224%22%3A+%5B245%2C+239%2C+0%5D%2C+%225%22%3A+%5B241%2C+89%2C+32%5D%2C+%226%22%3A+%5B168%2C+0%2C+0%5D%2C+%227%22%3A+%5B0%2C+143%2C+201%5D%7D&bidx=1&unscale=false&resampling=nearest&max_size=1024&return_mask=true

[...]

PR #1462, opened 2025-02-12
fix: categorical legend

hanbyul-here:
**Related Ticket:** #1456 

### Description of Changes
Categorical legend often doesn't have a colormap name but custom colormap values. This PR makes colormap&&rescale combination valid sourceParam. I think I would still like to run it with back-end team re: what makes sourceParams valid. 

We need to consolidate where we are subbing colormap and rescale value. The current dataflow seems to be scattered, making it difficult to pin down where the source of the data is. IMO, it should happen while reconciling the render data. 

Also it is interesting... that [GeoGLAM dataset](https://earthdata.nasa.gov/dashboard/exploration?datasets=%5B%7B%22id%22%3A%22geoglam%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2025-01-01T05%3A00%3A00.000Z) has been displayed just fine - which is another categorical dataset with colormap value. It just didn't get caught in any of our error Logic - which probably means that we can do better 😬 

### Validation / Testing
Check campfire nlcd data from the preview of https://github.com/NASA-IMPACT/veda-config/pull/669
direct url: https://deploy-preview-669--visex.netlify.app/exploration?datasets=%5B%7B%22id%22%3A%22campfire-nlcd%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2019-01-01T05%3A00%3A00.000Z

### DIFF
--- a/app/scripts/components/exploration/data-utils-no-faux-module.ts
+++ b/app/scripts/components/exploration/data-utils-no-faux-module.ts
@@ -155,8 +155,13 @@ export function resolveLayerTemporalExtent(
   }
 }
 
+// What is a valid source params to make the dataset render as expected?
 const hasValidSourceParams = (params) => {
-  return params && 'colormap_name' in params && 'rescale' in params;
+  return (
+    params &&
+    ('colormap_name' in params || 'colormap' in params) &&
+    'rescale' in params
+  );
 };
 
 /**
--- a/app/scripts/components/exploration/data-utils-no-faux-module.ts
+++ b/app/scripts/components/exploration/data-utils-no-faux-module.ts
@@ -211,35 +211,33 @@ export function formatRenderExtensionData(renderData): SourceParameters {
   };
 }
 
+export type SourceParametersWithLayerId = SourceParameters & {
+  layerId: string;
+};
 // renderParams precedence
-// 1. if the layer requires specific asset first,
-// 1-1. If so, check if renderExtension has that asset as a name space.
-// 1-1-1. If so, use the data.
-// 1-1-2. If not, fallback to user defined source parameters
-// 2. if renderExtension data exists
-// 2-1. If so, use the render Extension
-// 3. return user defined source parameters (which can be undefined)
+// 1. Check if render extension has a layer specific namespace from render extension
+// 1-1. If so, return the layer specific render data
+// 2. Check if user-defined source parameters haave assets defined
+// 2-1. If so, return user defined source parameters
+// 3. Check if render extension has dashboard namespace
+// 3-1. If so, use the render Extension data
+// 4. return user defined source parameters (which can be an empty object)
 export function resolveRenderParams(
-  datasetSourceParams: Record<string, any> | undefined,
+  datasetSourceParams: SourceParametersWithLayerId,
   queryDataRenders: Record<string, any> | undefined
-): SourceParameters | undefined {
-  // @NOTE: Render extension might not have separate namespaces for each asset yet. (2025 Feb)
-  // Fallback to manual source parameters when assets are specified
-  if (datasetSourceParams?.assets) {
-    if (queryDataRenders && queryDataRenders[datasetSourceParams.assets])
-      return formatRenderExtensionData(
-        queryDataRenders[datasetSourceParams.assets]
-      );
-    return datasetSourceParams;
-  }
-
-  // return render extension data
-  if (queryDataRenders && queryDataRenders[RENDER_KEY]) {
-    const renderParams = queryDataRenders[RENDER_KEY];
-    return formatRenderExtensionData(renderParams);
-  }
-  // return user defined source params (which can be undefined)
-  return datasetSourceParams;
+): SourceParameters {
+  const { layerId, ...rest } = datasetSourceParams;
+  // 1. Check if render extension has a layer specific namespace from render extension
+  if (queryDataRenders && queryDataRenders[layerId])
+    return formatRenderExtensionData(queryDataRenders[layerId]);
+  // 2. (TO DEPRECATE Once all the assets specific source parameters are namespaced with render extension)
+  // Return user defined source parameters if there is one
+  if (datasetSourceParams?.assets) return rest;
+  // 3. Return dashboard namespace render extension data
+  if (queryDataRenders && queryDataRenders[RENDER_KEY])
+    return formatRenderExtensionData(queryDataRenders[RENDER_KEY]);
+  // 4. return user defined source params (which can be an empty object)
+  return rest;
 }
 
 export function getTimeDensityStartDate(date: Date, timeDensity: TimeDensity) {
--- a/app/scripts/components/exploration/data-utils-no-faux-module.ts
+++ b/app/scripts/components/exploration/data-utils-no-faux-module.ts
@@ -24,7 +24,8 @@ import {
   VedaDatum,
   DatasetData,
   DatasetLayerType,
-  ParentDatset
+  ParentDatset,
+  SourceParameters
 } from '$types/veda';
 
 export function getParentDataset(data: DatasetData): ParentDatset {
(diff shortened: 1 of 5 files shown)


dzole0311 (approved):

anayeaye:
I am curious about cases where 
- there are no sourceParams (and none are needed) such as the planet data that is displayed with blue tarp detections https://github.com/NASA-IMPACT/veda-config/blob/develop/datasets/ps_blue_tarp_detections.ej.data.mdx?plain=1#L49
- there are valid sourceParams that have a colormap_name but no rescale (adding rescale would not work) such as soil-texture https://github.com/NASA-IMPACT/veda-config/blob/develop/datasets/soil-texture.data.mdx?plain=1#L28

hanbyul-here:
@anayeaye  
Regarding Blue Tarp, this dataset has source param from render extension - https://openveda.cloud/api/stac/collections/blue-tarp-detection  
```
      "rescale": [
        [0, 400]
      ],
      "resampling": "cubic_spline",
      "colormap_name": "reds"
  ```

But if the mdx configuration and render extension are both empty, the dashboard will make requests without any sourceParams.

Re: soil texture data - I think this is the same with geoglam data - it should have thrown an error since it doesn't have a valid sourceparams (at least for the dashboard) - but we have a bug that we are not throwing an error with some specific combination- user configured source param is not valid + render param exists + render param is not valid. This is a bug, but factoring all these cases, there doesn't seem to be any definite rule that we can tell whether the source is valid or not. We can skip any validation process if that reflects the actual rendering of the dataset more correctly?

**Comment by hanbyul-here** [comment]
Not directly related but kind of related - do all the datasetes have `dashboard` namespace now? (Can we expect `dashboard` namespace for all the datasets on VEDA?) Asking since we had to depend on user-configured data to retrieve render data ex. https://earth.gov/ghgcenter/api/stac/collections/micasa-carbonflux-monthgrid-v1 If I remember correctly, this dataset render extension did not have `dashboard` namespace before but now it does!

**Comment by hanbyul-here** [comment]
Based on the discussion with @anayeaye @smohiudd and @dzole0311  I made several more changes 
- Removed validation
- Changed the precedence of resolving sourceParams. The order is like this: 
// 1. if the layer requires specific asset,
// 1-1. If so, check if renderExtension has that asset as a name space.
// 1-1-1. If so, use the renderExtension data.
// 1-1-2. If not, fallback to user-defined source parameters
// 2. if renderExtension data exists
// 2-1. If so, use the render Extension
// 3. return user defined source parameters (which can be undefined)

@dzole0311 I will appreciate if you can review this one more time 🙇

**Comment by anayeaye** [comment]
> // 1-1. If so, check if renderExtension has that asset as a name space.

I think it could be problematic to mix item asset keys with the renders namespace visualization object keys because renders could include multiple assets in source parameters. Can we make a 1:1 mapping of `<dataset-name>.data.mdx layer.id` ([example](https://github.com/NASA-IMPACT/veda-config/blob/develop/datasets/so2.data.mdx?plain=1#L24)) to the renders object keys?

In this case the logic might be 
1. if layer.id in renders keys, use renders[layer.id], 
2. else if renders["dashboard"] exists, use that
3. else use mdx sourceParams

@hanbyul-here @smohiudd @dzole0311

**Comment by hanbyul-here** [comment]
thanks for the feedback @anayeaye  - I think eventually we should make an attribute something like `render_namespace` for mdx files, but I am going with layer id for now. Considering some datasets don't have namespaces needed, I made the steps like below.

// 1. Check if render extension has a layer specific namespace from render extension
// 1-1. If so, return the layer specific render data
// 2. Check if user-defined source parameters haave assets defined
// 2-1. If so, return user defined source parameters
// 3. Check if render extension has dashboard namespace
// 3-1. If so, use the render Extension data
// 4. return user defined source parameters (which can be an empty object)

**Comment by dzole0311** [comment]
Tested with a few datasets and works as outlined.

Just to confirm: is there a plan to audit which datasets still rely on [step 2 here](https://github.com/NASA-IMPACT/veda-ui/pull/1462/files#diff-3ef5907984dd01769eba7ea9935db817f59bce65779772a63dfac40ede3a8511R233) and track the progress of moving all of them to backend renders? It sounds like the step can only be safely removed once we have confidence that no datasets rely on mdx settings anymore. 

Is there maybe a way to log when step 2 is used in the UI (e.g, logging the dataset / layer to an endpoint), so we have visibility into which datasets/layers are missing render extensions?

**Comment by anayeaye** [comment]
@hanbyul-here @dzole0311 this change looks like it would support all existing mdx configurations and would incrementally start using the renders info from STAC 😎. Right now there are no custom namespaced renders configs but the majority of the production collections do have a dashboard namespace. 

> 1. Check if render extension has a layer specific namespace from render extension

We do not have a task in our backlog to update VEDA collections with namespaced renders configuration but we could encourage namespaced renders going forward and then plan to update the existing collections (might need to make this an objective for the next PI if it is the plan).


> 2. Check if user-defined source parameters haave assets defined

I'm a little unclear what has assets/ has no assets is used for. Is it because contributors were not previously required to set an asset in the mdx file because veda-ui assumed there would be an asset with the key `cog_default`? I think [that is what is happening](https://github.com/search?q=repo%3ANASA-IMPACT%2Fveda-ui%20cog_default&type=code) but I don't know what this check is doing.

cc: @smohiudd

**Comment by hanbyul-here** [comment]
@anayeaye  - yah, it does fallback to cog_default if asset is not defined. But if an asset is defined, it usually means that the asset requires a different visualization than default (dashboard name spaced one)  - such as soil texture data example (https://openveda.cloud/api/stac/collections/soil-texture). So this is a work around, not perfect, but can't think of any other way now. I think we should have a PI objective, something like making the back-end the source of truth for dashboard rendering and work together!

File: app/scripts/components/exploration/data-utils-no-faux-module.ts [hunk 1]
```diff
@@ -201,48 +199,45 @@ function flattenAndCalculateMinMax(rescale: number[]): [number, number] {
...
+  layerId: string;
+};
+// renderParams precedence
+// 1. Check if render extension has a layer specific namespace from render extension
+// 1-1. If so, return the layer specific render data
+// 2. Check if user-defined source parameters haave assets defined
+// 2-1. If so, return user defined source parameters
+// 3. Check if render extension has dashboard namespace
+// 3-1. If so, use the render Extension data
+// 4. return user defined source parameters (which can be an empty object)
```
slesaad, line 224: Sorry, I'm just now seeing this @hanbyul-here 

I'd recommend that the user-defined `sourceParams` be # 1 in precedence.

The renders that the stac catalog provides are the data provider's recommendation. However, the user who is working on adding the dataset to the dashboard should be able to customize how the data is visualized rather than forcing them to use whatever is recommended.

hanbyul-here: What you suggested makes a lot of sense. This approach mainly comes from our short-term goal of delegating as much as possible to the STAC endpoint while keeping MDX metadata as lightweight as possible.

From my understanding, at least for the dashboard instance, data providers also edit the data. This means they are duplicating efforts—first ingesting the data and then writing MDX metadata. Is there any scenario where authors are meaningfully overwriting sourceParameters? For example, in the recent bug with multiple bidx, authors were overwriting the source parameters with asset_bidx, but ultimately, this was just a rewrite of bidx values.

The best approach might be to allow users to overwrite only specific attributes rather than the entire source parameter object - so they don't have to scratch their heads when they just want to see the data in a different colormap.

PR #1465, opened 2025-02-21
fix: display infoDescription for layer modal

hanbyul-here:
**Related Ticket:** Reported by GHG team that the dashboard is unable to load any kind of info description of the layer

image (1)

### Description of Changes

I think `infoDescription` was somehow missed from one of the changes. I did not trace down the changesets, but my guess is that the data catalog only needs the id and name of the parent dataset, while e&a needs infoDescription, and we somehow ended up with a way of getting the parent dataset for data catalog.  I consolidated them to be one type (which means that data catalog parent dataset will have an unnecessary attribute - which I think is better than diverging `ParentDataset` type throughout the app.)  This PR picks it up again

### DIFF
--- a/app/scripts/components/common/catalog/catalog-content.tsx
+++ b/app/scripts/components/common/catalog/catalog-content.tsx
@@ -21,6 +21,7 @@ import {
 import { OptionItem } from '$components/common/form/checkable-filter';
 import { Pill } from '$styles/pill';
 import { usePreviousValue } from '$utils/use-effect-previous';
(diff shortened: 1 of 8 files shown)


snmln (approved):
Discussed with Hanybul via slack original issue was associated with infoDescription not being provided on initial load. These changes appear to fix the issue.

paridhi-parajuli:
Tested the branch for veda-config-ghg, this seems to fix the issue!

PR #1466, opened 2025-02-21
fix: Left Align Banner and Guidance Content

snmln:
### Description of Changes
Left aligning banner content based on`ENABLE_USWDS_PAGE_HEADER`. If header is set to FALSE the margin-auto css will be removed. 

Brian and Alex requested that the banner content be left aligned 
Screenshot 2025-02-21 at 1 09 40 PM

https://deploy-preview-165--earth-information-center.netlify.app/

### Validation / Testing
Check netlify preview for left aligned content when .env `ENABLE_USWDS_PAGE_HEADER` is set to FALSE

### DIFF
--- a/app/scripts/components/common/banner/banner.scss
+++ b/app/scripts/components/common/banner/banner.scss
@@ -4,6 +4,9 @@
   top: 3px;
 }
 
-.veda_one_padding .usa-banner__inner {
-  margin: 0;
+.veda_one_padding {
+  & .usa-banner__inner,
+  & .usa-banner__content {
+    margin: 0;
+  }
 }
--- a/app/scripts/components/common/banner/banner.scss
+++ b/app/scripts/components/common/banner/banner.scss
@@ -1,3 +1,6 @@
 .usa-banner__button:after {
-    top: 3px;
+  top: 3px;
+}
+.usa-banner__inner {
+  justify-content: center;
 }
--- a/app/scripts/components/common/banner/banner.scss
+++ b/app/scripts/components/common/banner/banner.scss
@@ -1,6 +1,9 @@
+@use '$styles/veda-ui-theme-vars.scss' as themeVars;
+
 .usa-banner__button:after {
   top: 3px;
 }
-.usa-banner__inner {
-  justify-content: center;
+
+.veda_one_padding .usa-banner__inner {
+  margin: 0;
 }
--- a/app/scripts/components/common/banner/banner.scss
+++ b/app/scripts/components/common/banner/banner.scss
@@ -4,7 +4,7 @@
   top: 3px;
 }
 
-.veda_one_padding {
+.banner--left-aligned {
   & .usa-banner__inner,
   & .usa-banner__content {
     margin: 0;
--- a/app/scripts/components/common/banner/index.tsx
+++ b/app/scripts/components/common/banner/index.tsx
@@ -9,6 +9,7 @@ import {
   USWDSBannerGuidance,
   USWDSMediaBlockBody
 } from '$uswds';
+import { checkEnvFlag } from '$utils/utils';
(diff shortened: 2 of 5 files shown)


snmln:
@dzole0311 I noticed that this [line for the guidances](https://github.com/NASA-IMPACT/veda-ui/blob/04c85c2e1679637d0539fae2d101ed371c967ecc/parcel-resolver-veda/index.js#L143C1-L144C1) is overriding the veda.config settings for `defaultIsOpen`. I want to confirm my understanding, are instances supposed to change the settings in the parcel resolver or in veda.config to have the guidances correctly display?

app/scripts/components/common/banner/index.tsx
```diff
@@ -91,16 +92,15 @@ export default function Banner({
...
   contentId = 'gov-banner-content'
 }: BannerProps) {
   const [isOpen, setIsOpen] = useState(defaultIsOpen);
-
   const leftContent = {
     ...DEFAULT_GUIDANCE.left,
     ...leftGuidance
   } as GuidanceContent;
-
+  const uswdsHeaderActive = checkEnvFlag(process.env.ENABLE_USWDS_PAGE_HEADER);
```
dzole0311, line 103: Why do we need a check for the header?

Can we simply add a new setting for the banner in the veda.config.js (e.g. `isLeftAligned`), that adds the new class if it's set to true?

app/scripts/components/common/banner/banner.scss
```diff
@@ -1,3 +1,12 @@
+@use '$styles/veda-ui-theme-vars.scss' as themeVars;
```
dzole0311, line 1: Is this import needed?

app/scripts/components/common/banner/banner.scss
```diff
@@ -1,3 +1,12 @@
...
+
 .usa-banner__button:after {
-    top: 3px;
-}
\ No newline at end of file
+  top: 3px;
+}
+
+.veda_one_padding {
```
dzole0311, line 7: What does one padding refer to here? 

```suggestion
.banner--left-aligned {
```

dzole0311 (changes requested):

dzole0311:
> @dzole0311 I noticed that this [line for the guidances](https://github.com/NASA-IMPACT/veda-ui/blob/04c85c2e1679637d0539fae2d101ed371c967ecc/parcel-resolver-veda/index.js#L143C1-L144C1) is overriding the [veda.config settings for](https://github.com/NASA-IMPACT/veda-ui/blob/04c85c2e1679637d0539fae2d101ed371c967ecc/mock/veda.config.js#L138C1-L139C1) `defaultIsOpen`. I want to confirm my understanding, are instances supposed to change the settings in the parcel resolver or in veda.config to have the guidances correctly display?

Good catch, ideally via the veda config. Could we add that small change so that the resolver simply resolves the `defaultIsOpen` prop from the veda config and passes it down to the banner (e.g. same like the `flagImgSrc`)?

app/scripts/components/common/banner/index.tsx
snmln: We could. The original intention was to base this left alignment implementation off of what menu is in use. The veda 1 menu stretches full width while the veda 2 USWDS menu does not stretch full width. So the new banner updates create a misalignment when paired with the veda 1 menu design. I think for the time being we should stick to strictly applying left alignment only when the veda 1 menu is in use to avoid potential alignment issues with any veda 2 menu implementations.

app/scripts/components/common/banner/banner.scss
snmln: No, that was a carry over from some initial exploration with the ticket.

PR #1467, opened 2025-02-24
refactor: Remove USWDS assets from the lib bundle

dzole0311:
**Related Ticket:**  https://github.com/NASA-IMPACT/veda-ui/issues/1389
**Related veda-config PR:** https://github.com/NASA-IMPACT/veda-config/pull/720

### Description of Changes
- Removed USWDS assets from the lib bundle to reduce size and prevent duplication
- Created placeholder `theme.css` file with a warning about its purpose. The file will be populated only by the veda-config build process
- Replaced the DevSeed UI button element in the DatasetMenu with one from USWDS, as it was conflicting with the PageHero styles

This PR is part of a coordinated update with https://github.com/NASA-IMPACT/veda-config/pull/720 to implement USWDS theming for the current generation of VEDA instances

### Validation / Testing

Most of the testing is done via the veda-config PR. Here we only need to quickly verify:

- Run `yarn serve` in the `veda-ui` repo and confirm the app loads with the styling set in the `uswds-theme.scss`


app/index.html
```diff
@@ -101,13 +101,16 @@
     });
   </script>
 
+  <!-- Theme CSS - only populated by legacy VEDA-config instances, not used in standalone VEDA UI mode -->
+  <link rel="stylesheet" href="../static/styles/theme.css">
```
hanbyul-here, line 105: oh it looks like parcel adds cache busting hash anyway even if the css was included in this way. noice

app/scripts/components/data-catalog/dataset-menu.tsx
```diff
@@ -14,18 +14,14 @@ import {
   CollecticonEllipsisVertical,
   CollecticonPage
 } from '@devseed-ui/collecticons';
-import { Button } from '@devseed-ui/button';
+import { Button } from '@trussworks/react-uswds';
```
sandrahoang686, line 17: Instead of directly importing, can you import from here instead https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/uswds/button.tsx#L4

sandrahoang686:
Nice stuff! I followed the instructions and ran locally. Validated that updating the `$theme-color-x` in the themes file in veda-config were reflected in the Header component ✅ . 

The only thing is that I noticed the divider with the secondary nav looked off.
Screenshot 2025-03-03 at 11 05 23 AM

hanbyul-here:
I pushed a tiny style fix - since USWDS tries to give bold outlines to all the possible focusable elements, the search bar looks a bit weird while being focused. (We would need to replace it to use actual uswds search bar at aome point: https://designsystem.digital.gov/components/search/) Since the parent component handles the focus style anyway, I hid the focus ring for the search input for the data catalog.

Without disabling focus style of search box: 
Screenshot 2025-03-03 at 5 43 04 PM

hanbyul-here (approved):
To recap: 
- VEDA dashboard builds duplicated uswds css, but not for production build. 
- VEDA dashboard swaps the theme file with the UI theme file for every build. 

let's roll back unncessary customization (https://github.com/NASA-IMPACT/veda-config/pull/720/files) but I think this is in a good place to be merged. Thanks again for working on it 🙇

+ I think this might be a breaking change?

PR #1469, opened 2025-02-24
chore: Consolidate where to handle parentDatasets

hanbyul-here:
**Related Ticket:** While working on https://github.com/NASA-IMPACT/veda-ui/pull/1465, I realized that the logic to find parent data sets is scattered, and it isn't easy to track them down. I believe we put this logic on the browser side because when we first put it, parent dataset was not necessary outside of the E&A page. Now that we have other places that use parentDataset, I think making it a part of the data generation process makes sense at this point. 

Page for manual test - infolayer modal on e&A page: https://earthdata.nasa.gov/dashboard/exploration?datasets=%5B%7B%22id%22%3A%22lis-global-da-evap%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%2C%22colorMap%22%3A%22viridis%22%2C%22scale%22%3A%7B%22min%22%3A0%2C%22max%22%3A0.0001%7D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2021-12-01T05%3A00%3A00.000Z

This needs to go with next-veda-ui pr : https://github.com/NASA-IMPACT/next-veda-ui/compare/generate-parentdataset?expand=1


parcel-resolver-veda/stringify-yml-func.js
```diff
@@ -38,7 +46,15 @@ const md = markdownit();
  * @returns string
  */
 function stringifyYmlWithFns(data, filePath) {
-  const jsonVal = JSON.stringify(data, (k, v) => {
+  const mdxData = {
+    ...data,
+    layers: data.layers?.map((l) => ({
+      ...l,
+      parentDataset: getParentDataset(data)
```
AliceR, line 53:
```suggestion
      parentDataset: data.id
```

parcel-resolver-veda/stringify-yml-func.js
```diff
@@ -3,6 +3,14 @@ const path = require('path');
...
 const md = markdownit();
 
+function getParentDataset(data) {
+  return {
+    id: data.id,
+    name: data.name,
+    infoDescription: data.infoDescription
+  };
+}
+
```
AliceR, line 13:
```suggestion
```

parcel-resolver-veda/stringify-yml-func.js
AliceR: We would need to find the places where this is being used in the UI!

hanbyul-here: Yah this would require more changes but more I think this is the right approach.

parcel-resolver-veda/stringify-yml-func.js
hanbyul-here: I incorporated this change into my pr 🙇

hanbyul-here:
The changes become more than I wanted 😓 
Most of them are lint/ path fix. To list changes that are worth attention
- MDX parses parentDataset's id when parsing the layers: https://github.com/NASA-IMPACT/veda-ui/pull/1469/files#diff-0c9ec10e02d1b3fa24c44e3c8cb40da57189d09c9fcd83a3a3d18bc541d92eccR4
- The components look up parent dataset using the id only when needed using the hook: https://github.com/NASA-IMPACT/veda-ui/pull/1469/files#diff-69ae90b3919bc9b69907030460631d634d1592a178fccb0f0247dec1cadb6db2R1
- Removed unnecessary logic
- renamed  `exploration/data-util-no-faux-module` to `utils/data-util` - this was overdue clean up. (I probably should have separated this PR but well well ),  Moved not exploration specific utils to `utils/data-util`

app/scripts/utils/data-utils.ts
```diff
@@ -0,0 +1,24 @@
+import { VedaData, VedaDatum, DatasetData } from '$types/veda';
+
+export const findParentDataset = ({
+  datasets,
+  datasetId
+}: {
+  datasets: DatasetData[];
+  datasetId: string;
+}): DatasetData | undefined => {
+  const parentDatset = datasets.find(
```
AliceR, line 10: Could we incorporate this one line `datasets.find(...)` directly into the `useParentDataset` hook? I don't see a reason why we need to keep it as a separate utils function. Is this even used elsewhere? Seems you deleted the places where it was used in data-catalog.

AliceR (approved):
Thanks for cleaning up @hanbyul-here !
One comment regarding possible further simplification below, but this looks good to go.

app/scripts/utils/data-utils.ts
hanbyul-here: Yes! great idea.

app/scripts/utils/data-utils.ts
```diff
@@ -0,0 +1,24 @@
+import { VedaData, VedaDatum, DatasetData } from '$types/veda';
+
+export function findParentDatasetFromLayer({
```
hanbyul-here, line 3: @AliceR  - heads-up that I added this util that is used both for `data-catalog` and `data-catalog-legacy`. (It turns out finding parent dataset on fly is necessary for filtering logic on E&A - this logic is so convoluted at this point but I will not include the change of this logic in this pr.)

hanbyul-here: I will merge this since the pr is pretty big and I am worried about the conflict. lmk if you have any concern about this!

PR #1473, opened 2025-02-24
fix: changelog generation

hanbyul-here:
**Related Ticket:** #1435


dzole0311 (approved):
I guess to be tested with the next release?

PR #1474, opened 2025-02-25
ci: Add cache, and failure notice for release action

hanbyul-here:
**Related Ticket:** #1435

### Description of Changes
- Add caching and failure notice ( to Slack) for release action
- Slightly improve release note formatting

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,9 +46,6 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - weekindex
-    strategy:
-      matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -46,6 +46,9 @@ jobs:
     runs-on: ubuntu-latest
     needs:
       - weekindex
+    strategy:
+      matrix:
+        repo: ['nasa-impact/veda-config', 'US-GHG-Center/veda-config-ghg']
```
hanbyul-here, line 51: I don't think this will work - because dispatching an event to GHG-Center will require a different auth token.

snmln (approved):

Issue #1453, opened 2025-02-11
Implement Flag Layout Card from USWDS for DataCatalog Component


sandrahoang686:
Implement the Flag Layout Card from USWDS for DataCatalog Component (TODO: Confirm with @faustoperez  if this is the right card from USWDS).

Create the common attributes interface type between all cards... Since we are starting out with this card first, we should find the common attributes across different card types to use.

### Implementation Approach

- The base Card type will be the USWDS Card
- We will expose explicitly the cards from the veda-ui for usage (e.g. <StoryCard />)
- Naming of the card components is based on the entity type (e.g. <StoryCard />), matching the naming in the Figma designs

USWDS Cards: https://designsystem.digital.gov/components/card/
Figma Card Designs: https://www.figma.com/design/PG7AhRloW6qJd9U00VdQ77/VEDA-Template-Instance?node-id=996-8300&p=f&t=kVGohMEac8eFA6jQ-0

![Image](https://github.com/user-attachments/assets/c5bd44d5-9231-4b5e-8845-55677de9d4eb)

PR #1479, opened 2025-02-26
feat!: Add FlagCard & DefaultCard and implement FlagCard in DataCatalog

sandrahoang686:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1453
**Related PRs:** https://github.com/NASA-IMPACT/next-veda-ui/pull/52

https://deploy-preview-1479--veda-ui.netlify.app/data-catalog
https://deploy-preview-1479--veda-ui.netlify.app/sandbox/cards

### Description of Changes
* Added a `BaseCard` component where `FlagCard` and `DefaultCard` extend off of
   * BaseCard Attributes to be used across cards
* Created Fork of `CatalogContent` where we can start migrating components over to USWDS and renamed old one to `CatalogContentLegacy`. 
   * This is FF
* Updated ** - Exposed `CatalogContent` instead of `CatalogView`, updates reflected in connected NextJs PR
* Added `Tags` component from USWDS
* Added draft documentation of the components we are starting to expose from the library with props, I created this for now because we gotta start documenting this somewhere but we can figure out where and how best to document later
* Added both card types to `/sandbox/cards` because right now we dont have a storybook(esque) implemented but right now, it looked like a fine enough place to showcase

**What this doesn't do:**
* DataSelectorModal isn't updated but uses the old view, @faustoperez will work on the designs for that so a separate ticket will follow
* Card Accessibility states to be implemented as part of this ticket https://github.com/NASA-IMPACT/veda-ui/issues/1481

* `FlagCard` and `DefaultCard` can be exposed as is, if exposed as is, they dont have the wrapping routing logic wrapped around it which makes it behave as a button. If the wrapping routing logic is wanted. The `Card` Component which is our wrapper to support current routing logic could be exposed and used where type `FLAG-LAYOUT` would be passed in. Users can decide.

### Validation / Testing
* Validate that the Feature Flagging works `ENABLE_USWDS_DATA_CATALOG`
* Validate when enabled...
   * DataCatalog Cards look correct
   * Entire card behaves like a button
   * The Card routes correctly
   * Validate that the DataSelectorModal is still the same as originally since nothing should have changed with its look or behavior

_____
  Self Validation Loom Walkthrough (recommended to watch @1.5x)
https://www.loom.com/share/c52aa4227651422dbac9554a744f2aa8?sid=3ac71fca-189f-42ce-856c-0104ee6703c7


faustoperez:
Follow-up design ticket: https://github.com/NASA-IMPACT/veda-ui/issues/1485

dzole0311 (commented):
It's really nice to see the Loom videos again @sandrahoang686!

faustoperez (changes requested):
Great progress, Sandra! Here's my feedback:

Please review the hover + focus states for the different cards here: https://www.figma.com/design/5mclPTReHcRIzKbJm8YA6a/VEDA---USWDS?node-id=335-1412&t=njksqAoXJOF2b1xG-1

The subcomponents we used to design the different versions are summarised here (if this helps):
DUMMY CARD

Please also be mindful of the general styling (correct typeface, accent colors, paddings, etc.) -> I'm not sure if this applies to this particular PR or if it should be addressed elsewhere.

Thank you!

app/scripts/components/common/card/index.tsx
```diff
@@ -191,7 +193,26 @@ export default function CardComponent(
           tagLabels={data.tagLabels}
         />
       );
+    } else if (
+      cardType === CardType.FLAGLAYOUT &&
+      data.imgSrc &&
+      data.imgAlt &&
+      data.title &&
+      data.description &&
+      data.footerContent
```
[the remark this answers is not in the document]
sandrahoang686: yeah because `BaseCardComponentProps` from the OG card component has these attr as optional but for the Flag card they are required so i'm doing this check for ts sake

app/scripts/components/common/catalog/index.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import styled from 'styled-components';
 import { themeVal } from '@devseed-ui/theme-provider';
 import CatalogContent from './catalog-content';
+import CatalogContentLegacy from './catalog-content-legacy';
```
hanbyul-here, line 5: It will be worth lazy loading catalog content so the page script doesn't have both of the components? https://react.dev/reference/react/lazy

app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
```diff
@@ -11,7 +11,7 @@ import { TimelineDataset } from '../../types.d.ts';
 
 import RenderModalHeader from './header';
 import ModalFooterRender from './footer';
-import CatalogContent from '$components/common/catalog/catalog-content';
+import CatalogContent from '$components/common/catalog/catalog-content-legacy';
```
hanbyul-here, line 14: (mainly for curiosity and plan for e&a page) - can we use the new catalog content here?

docs/components/FEATURE_COMPONENTS.md
```diff
@@ -0,0 +1,16 @@
+## DRAFT LOG OF COMPONENTS AND PROPS
```
hanbyul-here, line 1: (if you are looking for feedback)
feature_components might not be self-explanatory for readers. Either to explain what it is, or better to merge the components documentation and make a section for composite components. Also, just fyi that we can automate this with jsdoc syntax !

hanbyul-here:
The general pattern looks good. But as @faustoperez, it looks like we can do some more polishings. Can we list out the items so we can work on it? @faustoperez ? Let's see the LOE of the items and decide whether to put changes in this PR or to make a follow up. 

One thing I found is that you can click outside of the card because a tag doesn't match the size of the card

Screenshot 2025-03-04 at 5 34 27 PM

app/scripts/index.ts
```diff
@@ -39,6 +38,9 @@ import {
 
 // Include only the custom stylings for the VEDA components into the library build
 import './styles/veda-components.scss';
+import CatalogContent from '$components/common/catalog/catalog-content';
+import DefaultCard from '$components/common/card/uswds-cards/default-card';
```
hanbyul-here, line 42: 🤔 (Please correct me if I am reading it wrong) as far as I can read, this is a straigt up USWDS card - if we are not adding anything in the middle, do we need to export these? What do you think is the benefit of using this card compared to using the card directly from USWDS?

app/scripts/components/common/card/index.tsx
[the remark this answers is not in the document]
AliceR: Could we instead mark them as required in `FlagCardProps`?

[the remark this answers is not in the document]
sandrahoang686: they are

app/scripts/components/exploration/components/dataset-selector-modal/index.tsx
sandrahoang686: that will be a separate issue. I'm not going to worry about this in this PR because of the selecting differences.

app/scripts/index.ts
sandrahoang686: "straight up USWDS" card is maybe not the best way to phrase or look at it. These are extensions off the BaseCard [here](https://github.com/NASA-IMPACT/veda-ui/blob/8f12a962176a734321a173e9463d3d6951e77f9a/app/scripts/components/common/card/uswds-cards/index.tsx) which does compose the USWDS card elements together so there kindof is a "middle". I've created `DefaultCard` to maintain explicitness but the similarities are very close when compared to FlagCard. The only difference is we are passing in the type for the user and I define a default gridLayout value which can be overwritten.

app/scripts/index.ts
```diff
@@ -39,6 +38,9 @@ import {
 
 // Include only the custom stylings for the VEDA components into the library build
 import './styles/veda-components.scss';
+import CatalogContent from '$components/common/catalog/catalog-content';
```
hanbyul-here, line 41: Also heads-up that this is a breaking change (`CatalogView` is not available anymore). We want to be careful about introducing breaking change moving forward. Is this renaming worth the breaking change? if so, mark this as a breaking change!

app/scripts/components/common/card/uswds-cards/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {
+  USWDSCardComponent,
+  USWDSCardBody,
+  USWDSCardFooter,
+  USWDSCardHeader,
+  USWDSCardMedia
+} from '$uswds';
+import './styles.scss';
+
+// @TODO: Implement overrideable Card Acessibility States
```
hanbyul-here, line 12: What does this mean?

docs/components/FEATURE_COMPONENTS.md
sandrahoang686: yeah! I just added this to start documenting somewhere lol but lmk if you got a better place right now to move it. Just wanted to start documenting props somewhere for our expose components

app/scripts/index.ts
sandrahoang686: Yeah this is breaking change for NextJs. And this isn't a rename actually, I expose CatalogContent instead of CatalogView and describe why above. I actually still dont know how we want to address breaking changes for nextJs but not for legacy instances. I guess we should just proceed to still mark? Marking!

app/scripts/components/common/card/uswds-cards/index.tsx
sandrahoang686: Hover states will be addressed in separate ticket

sandrahoang686:
@dzole0311 @hanbyul-here @faustoperez this is ready for a re-review! Please focus on just the `FlagCard` component. I'm not adressing the entire theme of the app in this PR, that is out of scope. @faustoperez related to Focus color, that will be addressed in the separate ticket to implement accessibility states. I've also updated the uswds theme mapping to point to `public-sans` instead of `baseFontFamily` which I think we kept for a specific reason in the past?? Maybe because of GHG but i'm not sure, either way i've updated the mapping but @dzole0311 can you lmk if we shouldn't do that for whatever reason?

app/scripts/index.ts
sandrahoang686: Yeah I was really trying to find that balance of flexibility but then also providing structure for our veda-ui uswds extensions. I believe this strikes the balance well but if things change, it should be easy enough to pivot :).

docs/components/FEATURE_COMPONENTS.md
AliceR: Just noting that the legacy Card component uses jsdocs already, see line 120ff in `app/scripts/components/common/card/index.tsx` https://github.com/NASA-IMPACT/veda-ui/pull/1479/files/8f12a962176a734321a173e9463d3d6951e77f9a#diff-f936775422e987f9214c5bbdc155e139c0404494103a14f1da0792e2825beef3L120

I prefer jsdocs due to IntelliSense tooltips and the proximity of docs to code (making it much easier to maintain).

app/scripts/components/common/card/uswds-cards/styles.scss
```diff
@@ -0,0 +1,9 @@
+@use '$styles/veda-ui-theme-vars.scss' as themeVars;
+
+.usa-card {
+  margin-bottom: 0;
```
AliceR, line 4: We need to remove the left and right margin as well, to have the `<a>` tag match the card itself. How about 
```suggestion
  margin: 0;
```

AliceR: Ah wait, the left and right margin are set on the `usa-card__container`, not the usa-card itself.

app/scripts/components/common/card/index.tsx
```diff
@@ -191,7 +195,26 @@ export default function CardComponent(
           tagLabels={data.tagLabels}
         />
       );
+    } else if (
```
AliceR, line 198: 💭  I am wondering why you decided to expand this logic here, instead of moving this to a `-legacy` file and creating our new cards in a new place? It could be easier to deprecate in the future if we don't have old and new cards intertwined...

dzole0311:
> I've also updated the uswds theme mapping to point to `public-sans` instead of `baseFontFamily` which I think we kept for a specific reason in the past?? Maybe because of GHG but i'm not sure, either way i've updated the mapping but @dzole0311 can you lmk if we shouldn't do that for whatever reason?

@sandrahoang686 Yeah, we used it so that GHGC can pass their own font via a root variable. I think we can safely remove it now and use the new USWDS theming pipeline I've added to set the font from a uswds-theme file in GHGc.

app/scripts/components/common/card/index.tsx
sandrahoang686: This is the controller/wrapper file that I spent this [PR](https://github.com/NASA-IMPACT/veda-ui/pull/1407) to clean for breaking out to scale. We will need this controller file for legacy and modern code to do our wrapping logic for routing. This file is clean enough to add onto so it is fine but ideally for better pattern practices, NextJs instances should use the card directly and do their own wrapping routing logic. But we have it here as an option.

app/scripts/index.ts
hanbyul-here: It is great that we are not afraid of making a breaking change. The other side to consider is when building a library is that we offer compatibility. (ex. when a user is updating the ui library to get the new features, they will be surprised that they will need to rename components.) For example, we can export CatalogContent but still can export CatalogView to support the old way of syntaxes. (sudo; `CatalogView = CatalogContent; export {CatalogContent, CatalogView}`. This is something that we need to keep in mind going forward.

hanbyul-here (approved):
thanks for handling all the comments (just to be clear. I only checked if my comments were handled! so some might be still in wip, but leaving approval here to mark that I checked your update)

dzole0311 (approved):

faustoperez:
> @dzole0311 @hanbyul-here @faustoperez this is ready for a re-review! Please focus on just the FlagCard component. I'm not adressing the entire theme of the app in this PR, that is out of scope. @faustoperez related to Focus color, that will be addressed in the separate ticket to implement accessibility states. I've also updated the uswds theme mapping to point to public-sans instead of baseFontFamily which I think we kept for a specific reason in the past?? Maybe because of GHG but i'm not sure, either way i've updated the mapping but @dzole0311 can you lmk if we shouldn't do that for whatever reason?

Looking good Sandra! Hope this feedback helps:

- The Public Sans font doesn't seem to be loading :/
- Please also be mindful of the text sizes and colors in the card, and the paddings and margins of the content type tag, also the color of the icon and the tag border color.
- The margins on the focus state are off, also no border-radius and no underlining of the title, though you mention that it will be addressed in a separate ticket.

**Implementation:**
image

**Design:**
image

sandrahoang686:
@faustoperez thanks for reviewing again! Those comments are helpful when they are specific, i'm going to merge this to not hold but i've created the Card accessibility states ticket here https://github.com/NASA-IMPACT/veda-ui/issues/1501 🙇🏼‍♀️. When it comes to small styling tweaks, can you include what is blocking for you in the future as part of the comment as long as specifics of what is off and what it actually needs to be?

Issue #1482, opened 2025-02-27
[Bug] Clicking on a topic/source tag from a story page doesn't populate the "Filter" button


slesaad:
### Context

URL: https://earth.gov/ghgcenter/stories?taxonomy={"Topics":["methane"]}

Blocking a release? (y/n): n

Regression issue? (y/n): n
[If yes, add Regression label]

Priority: P2

### Steps to reproduce the issue

1. Go to this story: https://earth.gov/ghgcenter/stories/modeling-natural-methane-emissions
2. Click on the "Methane" topic
3. See how even though the stories are filtered by the topic, the "Filter" button isn't populated like it normally does

![Image](https://github.com/user-attachments/assets/7c118a3f-3e4b-42e9-a968-422366bed41f)

### Other details

It looks like in the stories page, the URL looks like this: 

```
https://earth.gov/ghgcenter/stories?taxonomy={"Topics":"methane"}
```

, but coming from the story page, it looks like this:

```
https://earth.gov/ghgcenter/stories?taxonomy={"Topics":["methane"]}
```

PR #1483, opened 2025-02-27
fix: Topic and Source dropdown in stories tab

vidhi04bhatia:
**Related Ticket:** _(https://github.com/NASA-IMPACT/veda-ui/issues/1482)_

### Description of Changes
_Topic and Source field in Stories Tab on the dashboard under Browser do not retain the values selected on single stories_

### Validation / Testing
_Validate topics and source dropdown values after selecting topic/source from singe stories_

### DIFF
--- a/app/scripts/components/common/browse-controls/index.tsx
+++ b/app/scripts/components/common/browse-controls/index.tsx
@@ -132,14 +132,16 @@ export default styled(BrowseControls)`
 interface DropdownOptionsProps {
   size: ButtonProps['size'];
   items: FilterOption[];
-  currentId?: string;
+  currentId?: string | string[];
   onChange: (value: FilterOption['id']) => void;
   prefix: string;
 }
 
 function DropdownOptions(props: DropdownOptionsProps) {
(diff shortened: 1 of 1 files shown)


hanbyul-here:
Thanks for your contribution, can you share a bit more about what you found while fixing this issue? such as, Is `currentId` always `string[]` or it is actually `string | string[]` ?

AliceR:
It seemed strange to me that the id could be both a string and an array, but we care about the first item only, anyways. Made the alternative suggestion [here](https://github.com/NASA-IMPACT/veda-ui/pull/1499) to encode the id as directly string, removing the wrapping brackets.

hanbyul-here:
My understanding is @AliceR 's pr fixed the issue. Are there any more items that this PR handles?

vidhi04bhatia:
@hanbyul-here No, that was all in this PR.

hanbyul-here:
Okay, since Alice's pr is merged to main, I will close this one. Thanks for your contribution again @vidhi04bhatia !

Issue #1489, opened 2025-03-04
Prepare Sandbox for USWDS


hanbyul-here:
We have a sandbox page that we can use for USWDS integration. It is a little forgotten at this point, revamp it. 

Stretch goal: make it use library version of veda-ui

PR #1490, opened 2025-03-05
Refactor: Sandbox clean up, split between legacy and USWDS

AliceR:
**Related Ticket:** [#1489](https://github.com/NASA-IMPACT/veda-ui/issues/1489)

### Description of Changes
Splitting the main section of the sandbox into two galleries, one for USWDS and one for the legacy components. 
Add USWDS colors page - help wanted! How can I access the scss theme variables in javascript/react to render the swatches?
Add pagination component split into presentational component and interaction container. I am hoping this will allow us to render the pagination server-side, while updating based on user interaction on the client. Using the 'use client' directive accordingly on the client component. Looking for opinions here!

Not sure how to approach using the library version of veda-ui. Maybe we should do that within next-veda-ui?

What do you all think about using storybook to render (and develop, and document!) components in isolation? I am thinking that would be the way to go, but I am also open to keep this sandbox setup if thats what people prefer.

### Validation / Testing
Access http://localhost:9000/sandbox to check out the current status


sandrahoang686:
Thanks for this work 🙌🏼 . I'm wondering can we split this PR though? Separate PR for sandbox cleanup and then separate PR for adding pagination? That way we can keep separate the actual purpose of the PR and it would be easier to review too.

AliceR:
@sandrahoang686 I removed the commits from this branch and added them to a separate one, PR here: https://github.com/NASA-IMPACT/veda-ui/pull/1492
This way we can keep the conversations about sandbox and pagination apart. Note that I am adding the pagination to the sandbox, hence use this branch as a base.

sandrahoang686 (approved):
thanks for the cleanup!

PR #1502, opened 2025-03-10
fix: Move UI library seed global style to component level, export veda specific global style to be used with UIlibrary provider

hanbyul-here:
**Related Ticket:** #1503 

### Description of Changes
Nav height (specifically, the bottom row) varies depending on which page a user is on in Next instance: https://veda-ui-next-test.netlify.app/exploration?datasets=%5B%5D&taxonomy=%7B%7D (Navigate to any other page than homepage, the bottom row becomes taller.) 
This is because devseed ui library injects the default global line height: https://github.com/developmentseed/ui-library-seed/blob/96807c534c46521ce781717aa09d20ae1134e99b/packages/theme-provider/src/theme.js#L59  - and the homepage does not use ui library theme - therefore it has a different global style value. In this pr, I am passing the global line-height value as inherent and moving the global line height value to the top level of each component. 

I am exporting GlobalStyle, and using it for the pages that use devseed ui library components (https://github.com/NASA-IMPACT/next-veda-ui/pull/56) - most of the globalstyle is just css variables that ui seed library needs, which I think should have no conflict with uswds. 

### Validation / Testing
There are two things to validate.
- The components don't look different inside of VEDA UI instance. 
- USWDS Header doesn't change heights in Next instance. 
- But this change will decrease the height of USWDS nav bottom row - Tagging @faustoperez to check

### DIFF
--- a/app/scripts/components/common/catalog/catalog-content.tsx
+++ b/app/scripts/components/common/catalog/catalog-content.tsx
@@ -33,6 +33,7 @@ import { usePreviousValue } from '$utils/use-effect-previous';
 
 import { useVedaUI } from '$context/veda-ui-provider';
 import { findParentDatasetFromLayer } from '$utils/data-utils';
+import { globalStyleCSSBlock } from '$styles/theme';
 
 const EXCLUSIVE_SOURCE_WARNING =
   'Can only be analyzed with layers from the same source';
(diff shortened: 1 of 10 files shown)


hanbyul-here:
@snmln  - I unassigned you since I will make more changes. I will assign you again when ready 🙇

hanbyul-here:
@snmln this pr is ready for a review 🙇

hanbyul-here:
@snmln Thanks for checking it thoroughly 🙇 Do you mind sharing where you found the gap of dimension (ex. which element on which page)? I will merge this, but also will check if I missed anything.

snmln:
Of course! Specifically the height of the drop down nav. This is an image from the preview link above. 

Screenshot 2025-03-17 at 10 13 32 AM

hanbyul-here:
Yes! that is expected change, thanks @snmln !

Issue #1281, opened 2024-12-01
Full Page Widget


aboydnw:
### Context

Followup to #1261 
Ultimately we want every widget to allow for this functionality. Open to a discussion on whether that's feasible, or if we need to wait for implementing updated designs like we see with #1280 

### Acceptance Criteria
- [ ] Users can select to open a widget in a full page and see the page designed in #1261 
- [ ] Users can exit the full page, and it functions similar to browser back, returning user to the story/dataset overview, scrolled to the same position as before

In the future, we anticipate widgets will become a more integrated part of the Dashboard. So, we might have a widget catalog (similar to data or story catalog) where users can see all widgets configured across the different content items on the Dashboard. As part of this, a particular widget (i.e. line chart showing CO2 levels at a particular sensor for last 50 years) might be referenced across multiple content items and we want to allow users to traverse those content items. This is out of scope for this ticket, but FYI for where we think we will go in the future.

aboydnw:
@AliceR I went ahead and removed these two AC and put in #1508 just to move things along more quickly

> - [ ] (Ideally) Users can share the widget page directly with other people via URL 
> - [ ] (Ideally) If somebody visits a URL that has been shared with them, when they exit the full screen view then they are viewing the story or dataset overview that the widget is configured on

aboydnw:
To add some more details, I think this ticket would be considered complete if we can show our existing widgets with this full screen frame, allowing users to go full screen and back, with no additional functionality or changes to the widgets.

Widgets in scope:
- Map
- Map comparison
- Chart

Widgets out of scope
- Image/video (we should decide if we want this to function the same way
- Scrollytelling
- HTML tables (not really a widget right now, but calling it out anyways)

Issue #1508, opened 2025-03-13
Share widgets


aboydnw:
### Context

Widgets should become standalone items that can be shared directly with others. So, we'll need a way for people to copy a URL (via share button or other) and then determine what happens when somebody opens that URL.

### Acceptance Criteria

- [ ] Users can share the widget page directly with other people via URL
- [ ] If somebody visits a URL that has been shared with them, they see the widget in full page view (possibly its own page?)
- [ ] (TBD?) When they exit the full screen view then they are viewing the story or dataset overview that the widget is configured on

Issue #1261, opened 2024-11-18
[Design] Widget full page


aboydnw:
### Context

With a widget now as a defined content type in VEDA, users should be able to see a widget and share a widget with others. Long term, we envision having a widget catalog, and widgets that are shared across multiple stories/datasets. In the shorter term, we likely see just adding the ability for users to expand a widget to full screen mode from a story or dataset overview. 

So, we'd like some designs that are compatible with the long term, and also easy to implement in the short term. Imagine a user clicking an "expand" button on a story, seeing that widget full screen, being able to close it and return back to the story or share directly with somebody else. If somebody opens a link to a widget that has been shared with them, then they'd land on the full screen view and if they close it then ideally they see the story that the widget is on. (proposed workflow)

### Acceptance Criteria
- [x] Figma file outlining full screen view for the widget
- [x] Designs outlining the "expand" (name TBD) action that the user sees in a story/dataset overview page in order to open the full screen view, as well as any commentary on animation between the two

faustoperez:
Here are the designs for the widget and the full-page widget.

Figma link 👉 https://www.figma.com/design/PG7AhRloW6qJd9U00VdQ77/VEDA-Template-Instance?node-id=502-1205&t=juGRrUo41ZVT2DSG-1

The close icon is the same as the browser back. Users can access the full widget page from multiple sources: catalog, story, dataset overview, other widget pages, etc.

They feature these new VEDA/USWDS components which have been added to the library:

- Widget
- Card
- Segmented button
- Icon button

![Image](https://github.com/user-attachments/assets/baa725da-0422-4fd3-8088-01764a389e14)
![Image](https://github.com/user-attachments/assets/db570fe3-075e-4c63-9a73-e9014002e30a)

PR #1520, opened 2025-03-17
feat: Add widget container components for full page and full screen view

AliceR:
**Related Ticket:** contribute to #1281 

### Description of Changes
Adding two different components, a fullscreen widget using the [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API), and a full page widget using a model overlay with a micro animation. Both containers consist of a header with the button to toggle, and the content rendering the components children.

To show case the components I add some examples to the sandbox page. 

I did not yet add the container to our mdx content provider, I will do this in a separate PR to keep to scope small enough to easily review.

### Validation / Testing
- Explore examples on https://deploy-preview-1520--veda-ui.netlify.app/sandbox/
- Open and close full screen or full page mode

### DIFF
--- a/app/scripts/components/common/widget/fullpage-modal.tsx
+++ b/app/scripts/components/common/widget/fullpage-modal.tsx
@@ -35,25 +35,19 @@ const FullpageModalWidget = ({
   return (
     <>
       {!isFullPage && (
-        <div // widget-container
-          className='width-full height-full border border-base-light radius-md'
-        >
+        <div className='width-full height-full border border-base-light radius-md'>
           <Header
             isExpanded={isFullPage}
             heading={heading}
             handleExpansion={() => setIsFullPage(true)}
           />
-          <div // widget-content
-            className='padding-2'
-          >
-            {children}
-          </div>
+          <div className='padding-2'>{children}</div>
         </div>
       )}
 
       <AnimatePresence>
         {isFullPage && (
-          <motion.div // fullpage-overlay
+          <motion.div
             initial={{ opacity: 0, scale: 0.9 }}
             animate={{ opacity: 1, scale: 1 }}
             exit={{ opacity: 0, scale: 0.9 }}
@@ -65,11 +59,7 @@ const FullpageModalWidget = ({
               heading={heading}
               handleExpansion={() => setIsFullPage(false)}
             />
-            <div // widget-content
-              className='padding-2'
-            >
-              {children}
-            </div>
+            <div className='padding-2'>{children}</div>
           </motion.div>
         )}
       </AnimatePresence>
--- a/app/scripts/components/common/widget/fullpage-modal.tsx
+++ b/app/scripts/components/common/widget/fullpage-modal.tsx
@@ -1,0 +1,80 @@
+import React, { ReactNode, useState } from 'react';
+import { motion, AnimatePresence } from 'framer-motion';
+import Header from './header';
+
+interface FullpageModalWidgetProps {
+  heading: string;
+  children: ReactNode;
+}
+
+/**
+ * A React component that toggles the display of modal content between a compact
+ *  widget view and a full-page overlay.
+ *
+ * This component renders a container that initially shows a widget with a
+ * header and content. When the header's expansion handler is activated, the
+ * component transitions to a full-page mode using animated effects provided
+ * by Framer Motion's AnimatePresence and motion.div.
+ *
+ * @param {string} heading - The title text displayed in the header of the
+ * modal.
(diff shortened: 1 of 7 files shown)


app/scripts/components/common/widget/fullpage-modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+const FullpageModalWidget = ({
+  heading,
+  children
+}: FullpageModalWidgetProps) => {
+  const [isFullPage, setIsFullPage] = useState(false);
+
+  return (
+    <>
+      {!isFullPage && (
+        <div // widget-container
```
dzole0311, line 38: Obsolete comment

app/scripts/components/common/widget/fullpage-modal.tsx
```diff
@@ -0,0 +1,80 @@
...
+            className='padding-2'
+          >
+            {children}
+          </div>
+        </div>
+      )}
+
+      <AnimatePresence>
+        {isFullPage && (
+          <motion.div // fullpage-overlay
```
dzole0311, line 56: Same, can we make sure to clean up comments like these please? I think they add nothing useful to the dev experience.

app/scripts/components/common/widget/fullscreen-api.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { ReactNode, useRef, useState } from 'react';
+import Header from './header';
+
+/**
+ * Renders a widget component with fullscreen capability.
+ *
+ * This component wraps its children with a container that can be toggled into
+ * fullscreen mode.
+ * The fullscreen state is controlled via the [Fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API), and the component
```
dzole0311, line 9: This is great! Fwiw I think the overall wording in the ticket's AC might be a bit misleading:

> - Users can select to open a widget in a full page and see the page designed in https://github.com/NASA-IMPACT/veda-ui/issues/1261
> - Users can exit the full page, and it functions similar to browser back, returning user to the story/dataset overview, scrolled to the same position as before

I think what is meant is that we'll either:

1. open the widget in a modal (the second approach implemented in the PR + with extra url params so that it is shareable), or
2. a new page showing the widget, plus the "Appears In" section

app/scripts/components/common/widget/header.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  const updateButtonVisibility = useCallback(() => {
+    if (containerRef.current) {
+      setShowText(containerRef.current.offsetWidth > 450);
+    }
+  }, []);
+
+  useLayoutEffect(() => {
+    updateButtonVisibility(); // ensures that the button text visibility is set correctly when the component first renders.
```
dzole0311, line 40: Is this comment needed? I find the method's name pretty self explanatory

app/scripts/components/common/widget/fullpage-modal.tsx
AliceR: oh, ok - I liked them and made sure they are all nicely formatted :D 
But yeah, probably a <div> doesn't need additional explanation

app/scripts/components/common/widget/header.tsx
AliceR: In my first implementation I did not call the function and I was wondering why the button text did not show up in fullscreen mode. Realized this is not a resize but actually a re-render, so this fixed it.
Future me will wonder why I put it there, so I guess, future me would need it :)

app/scripts/components/common/widget/fullscreen-api.tsx
AliceR: We made the sharable part a separate ticket: https://github.com/NASA-IMPACT/veda-ui/issues/1508

I think there is more discussion to be had on how to approach the new page and routing situation when the url should include a stable reference to the standalone widget. But since I don't see that conflicting with the approaches implemented here, but just an extension in the functionality, I went ahead with just the simplest versions of a fullscreen/fullpage view. Since this is just in sandbox, we can explore it from a design perspective and better inform future development.

dzole0311 (approved):

faustoperez:
@AliceR the functionality works great, we'll need to review the implementation styles for the title and the button.

- Title is `h4`
- Button size is `small` -> button sizes [here](https://www.figma.com/design/5mclPTReHcRIzKbJm8YA6a/VEDA---USWDS?node-id=241-1004&t=WQqNJ7KbvtdNLHTh-1).

Love the cat gifs :)

Issue #1524, opened 2025-03-18
[Bug] Search in the Data Catalog throws an error


dzole0311:
### Context

URL: It's not on prod yet because veda-config has not been bumped to the latest UI version. Earliest deployment preview I could find: https://deploy-preview-1469--veda-ui.netlify.app/data-catalog

Blocking a release? (y/n): y

Regression issue? (y/n): y
[If yes, add Regression label]

Priority: P1

### Steps to reproduce the issue

1. Checkout the main branch
2. Go to Data Catalog
3. Search for something (e.g type 'CASA')
4. Notice how the page breaks

A git bisect shows that the issue might have been introduced [here](https://github.com/NASA-IMPACT/veda-ui/pull/1469)

aboydnw:
I agree this is blocking any uptake of the latest veda-ui version

PR #1526, opened 2025-03-18
Find parent dataset using layer id for search

hanbyul-here:
**Related Ticket:** #1524 

### Description of Changes
Finding parentDataset using layer Id

thanks for cathcing it 🙇 @dzole0311 

### Validation / Testing
Go to data catalog, and search for anything.


dzole0311 (approved):
Tested and works as before. I noticed that using the search with the new card doesn't highlight the searched term in the card, but that's a different issue.

PR #1535, opened 2025-03-19
fix: type generation

### LINKED ISSUE #1440: Set ts-check on next-veda
We want to have ts-check as part of the build pipeline.

### LINKED ISSUE #1533: Rearrange exported components/types

hanbyul-here:
**Related Ticket:** Close #1440  Close  #1533 

### Description of Changes
There are several changes here
- Change named function that are later exported by name (To work around Parcel's multiple export defaults problem: https://github.com/parcel-bundler/parcel/issues/6998)

to describe the problem in a simpler way,
The file like below
```
function ComponentA() {}
export default ComponentA
```
gets transpiled into in type definition (index.d.ts)
```
export default ComponentA;
```
So when importing the components from the library with named export, ts gives an error that there is no `ComponentA` (because it is default export) 

if we inline default export like below
```
export default function ComponentA
```
We can avoid this problem. 

- Split component imports into several files under `app/scripts/libs` 

I wonder if we can part away from Parcel's default type generation, or make the type generation pipeline better 🤔 mainly because I don't seem to have full control over it. 

type fix on this pr is not perfect - ex. the type files till has this import syntaxes on top of the file
```
import MDXImage, { Caption } from "$components/common/blocks/images";
import CompareImage from "$components/common/blocks/images/compare";
import Figure from "$components/common/blocks/figure";
import { ContentBlockProse as Prose } from "$styles/content-block";
```
which should not happen. I will separate this ticket. 

[This article](https://jakearchibald.com/2021/export-default-thing-vs-thing-as-default/) helped me understand how different ways of exports can behave differently. ex. I learned that the export default expression is different from the export default statement. (the previous one exports reference while the later one exports values. but I feel like that info doesn't help me solve this exact problem

### Validation / Testing
This change will get rid of ts error from https://github.com/NASA-IMPACT/next-veda-ui/blob/main/app/lib/index.ts . You won't be able to find multiple `export default` from built `lib/index.d.ts`. Everything else should work as before.


sandrahoang686 (approved):
Thanks for doing this tedious work. Also good knowledge to know which I didn't know before 👍🏼 .

PR #1538, opened 2025-03-20
chore: Add tests for the widget components

AliceR:
**Related Ticket:** close [#1522](https://github.com/NASA-IMPACT/veda-ui/issues/1522)

### Description of Changes
Adding tests for both the widgets on the sandbox page (with playwright) and testing the actual rendering with jest/testing-library.
Also added a few data-testid attributes to make it easier to select elements in tests, which works for both playwright and jest

There's now a snapshot test for the widget. If any changes accidentally affect this component, the test will catch it. Developers will need to review the changes and update the snapshot if they're intentional.

### Validation / Testing
You can run the tests by using `yarn test` for unit tests and `yarn test:e2e` for end-to-end tests.

### DIFF
--- a/app/scripts/components/common/widget/__snapshots__/widget.test.tsx.snap
+++ b/app/scripts/components/common/widget/__snapshots__/widget.test.tsx.snap
@@ -1,0 +1,45 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`FullscreenWidget matches snapshot 1`] = `
+<div
+  class="width-full height-full border border-base-light radius-md"
+  data-testid="widget-container"
+>
+  <div
+    class="padding-2 display-flex flex-justify space-between flex-align-center"
+    data-testid="widget-header"
+  >
+    <h6>
+      Test Fullscreen Widget
+    </h6>
+    <button
+      class="usa-button usa-button--outline margin-0 margin-left-3"
+      data-testid="widget-toggle"
+      type="button"
+    >
+      <svg
+        aria-label="Enter fullscreen"
+        class="usa-icon"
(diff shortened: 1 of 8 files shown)


AliceR:
Thanks @dzole0311 ! Looks like we did not update the tests after [refactoring the cards](#1479) - I wonder why our CI did not catch it?

Will work on fixing the tests in a separate branch...

dzole0311:
I've tested the widgets and all tests are passing. Since the entire widgets concept is still a work in progress, I think keeping them simpler for now is the right approach.

> I wonder why our CI didn't catch this issue?

Because of our switch to automated releases, which bypassed the e2e testing phase. [This is the doc](https://docs.google.com/document/d/1JGxibnxKz0MxD9DSvBDED2gCvee2K7Rz45T7KEO3zq8/edit?tab=t.0) where we discussed this concern. Here's what the previous workflow was:

1. Open a release-branch PR
2. The CI runs e2e tests specifically on that release branch
3. Merge only after all tests pass

Since the switch to the automated workflow:

1. Webhook automatically bumps the veda-ui version, commits it & publishes a git tag
2. No CI pipeline runs with the e2e tests

Some ideas I can think of:

1. Run e2e tests on every PR. The drawbacks is longer build times and probably increased test flakiness as we add more complexity and more e2e tests

2. Move e2e testing to veda-config instead of veda-ui
- Have e2e tests run on the release preview PR [created by the webhook](https://developmentseed.slack.com/archives/C03LGTM2SSH/p1742823482989679) with the updated veda-ui version
- This is similar to what the GHGC instance (veda-config-ghg) has

dzole0311 (approved):

Issue #1540, opened 2025-03-20
[Bug] render issues


slesaad:
### Context

URL: https://visex.netlify.app/stories/la-fires

Blocking a release? (y/n): y

Regression issue? (y/n): y
[If yes, add Regression label]

Priority: P1

### Description

>... trying to get the LA Fires data story ready for release, and the 3-band Planet imagery was visualizing correctly in the PR, but once approved and merged to develop (alongside some new ui changes, it appears) this no longer displays correctly.

slack thread: (https://teamdsig.slack.com/archives/C03LGTM2SSH/p1742505295057749?thread_ts=1742504923.354309&cid=C03LGTM2SSH)

Working PR link: https://github.com/NASA-IMPACT/veda-config/pull/752

Not working since merge to develop: https://visex.netlify.app/stories/la-fires

Data: Planet TrueColor Satellite Imagery

The working preview is making api requests using the `sourceParams` in data.mdx file, the non-working one is making request based on the stac render params (veda-ui version was upgraded from v5.13.1 -> v6.1.0)

![Image](https://github.com/user-attachments/assets/300203e8-423d-4e26-b075-fac6b3f7e937) (working)

![Image](https://github.com/user-attachments/assets/70d92066-eb15-4993-b54d-0da62a543dca) (not working)

However, the render params are working in the [stac browser](https://openveda.cloud/stac/collections/la-fires-planet?.language=en)

The difference is the stac browser is sending the query params as: (working)

```yaml
bidx: 1
bidx: 2
bidx: 3
assets: cog_default
```

but the dashboard is sending it as (not working with error "Input [bidx] should be a valid integer, unable to parse string as an integer")

```yaml
bidx: 1,2,3
assets: cog_default
```

regardless, the dashboard should prioritize the data.mdx `sourceParams` before the the stac render params.

So the two issues that need to be fixed:
1. Priority order of the source query params should be: mdx `sourceParams` > stac renders
2. Looks like this has already been reported in https://github.com/NASA-IMPACT/veda-ui/issues/1537

sandrahoang686:
**Note:** Regression seems to be linked to this PR [here](https://github.com/NASA-IMPACT/veda-ui/pull/1462). Currently investigating.

snmln:
the issue appears to be a layered check in some recent work. I have a solution in place to solve the issue but we will need to discuss a better solution next week with the team as a whole. 

https://github.com/NASA-IMPACT/veda-ui/pull/1543

[...]

PR #1544, opened 2025-03-21
fix: block-map render issue

snmln:
**Related Ticket:** [Bug ticket](https://github.com/NASA-IMPACT/veda-ui/issues/1540)

### Description of Changes
added additional layer to check to prioritize mdx data. THIS SOLUTION WILL NEED REVISITED

Veda-config release preview ticket: https://github.com/NASA-IMPACT/veda-config/pull/808

### DIFF
--- a/app/scripts/components/exploration/data-utils.ts
+++ b/app/scripts/components/exploration/data-utils.ts
@@ -198,8 +198,13 @@ export function resolveRenderParams(
   // Return user defined source parameters if there is one
   if (datasetSourceParams?.assets) return rest;
   // 3. Return dashboard namespace render extension data
-  if (queryDataRenders && queryDataRenders[RENDER_KEY])
+  if (
+    queryDataRenders &&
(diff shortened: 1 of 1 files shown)


app/scripts/components/exploration/data-utils.ts
```diff
@@ -198,8 +198,13 @@ export function resolveRenderParams(
   // Return user defined source parameters if there is one
   if (datasetSourceParams?.assets) return rest;
   // 3. Return dashboard namespace render extension data
-  if (queryDataRenders && queryDataRenders[RENDER_KEY])
+  if (
+    queryDataRenders &&
+    queryDataRenders[RENDER_KEY] &&
+    !Array.isArray(queryDataRenders[RENDER_KEY].bidx)
```
hanbyul-here: I made a small change because bidx is always array (even with one bidx.) - I changed it to return user defined sourceparams when asset_bidx is defined, which also should be revisited.

PR #1546, opened 2025-03-24
fix: playwright tests

AliceR:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1538#issuecomment-2740302710

### Description of Changes
Apparently we forgot to maintain our playwright test suite, this updates the specs so that they pass again.

We should update our CI pipeline to run playwright on open PR's!

### Validation / Testing
Run `yarn test:e2e`

### DIFF
--- a/test/playwright/pages/aboutPage.ts
+++ b/test/playwright/pages/aboutPage.ts
@@ -6,6 +6,8 @@ export default class AboutPage {
 
   constructor(page: Page) {
     this.page = page;
-    this.aboutParagraph = this.page.getByText("The VEDA Dashboard is one of several user interfaces in the VEDA project");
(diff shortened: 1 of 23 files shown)


dzole0311 (approved):
All tests pass.

Screenshot 2025-03-25 at 14 20 52

hanbyul-here (approved):
Yayy thanks for fixing e2e tests.

I would like to note that it took me more than 700 seconds to run e2e test. This long run-time is usually a reason that I don't tend to think that it is worth putting e2e test, because it will time out quite often that I will start not caring about the test result. 

Considering @dzole0311 took 10% of time, my case must be an exceptional one but I also would like to know if there is any way that we can decrease the variance of the runtime for e2e testing.

Issue #1484, opened 2025-02-28
[Bug] Legend Blocks Not Populating in Scrollytelling


acblackford:
### Context

All datasets used in a Scrollytelling Block must have a legend or once you get to the first dataset without a legend, any dataset after that point will not populate a legend even if they have one defined in the mdx. This was noted when going from data with gradient or categorical colormaps to satellite imagery (which typically is not shown with a legend on the site) and back to the gradient or categorical data visualizations. Quick fix was to add legend items for satellite data with a 0,0,0,0 RGBA call in the colorbar.  

URL: most recently noted on [blizzards](https://www.earthdata.nasa.gov/dashboard/stories/blizzards) and [tornadoes](https://www.earthdata.nasa.gov/dashboard/stories/2024-tornadoes) stories. 

Blocking a release? (y/n): n

Regression issue? (y/n): ?
[If yes, add Regression label]

Priority: P3

### Steps to reproduce the issue

1. [Step 1 to reproduce]
2. [Step 2 to reproduce]

PR #1550, opened 2025-03-24
fix: Legend populated after enountering a datset without it

vidhi04bhatia:
**Related Ticket:** #1484 

### Description of Changes
The current .mdx file contains a temporary fix where the legend properties are set to a transparent color to avoid issues with scrollytelling. We will update this to set legends that do not need a color swatch (e.g., categorical or graphical types) to type: text and remove unnecessary properties like stops, color, and label.

#### Before (Temporary Fix):
```
legend:
  type: categorical
  stops:
    - color: "rgba(0, 0, 0, 0)"  # using transparent for now to fix the issues with scrollytelling
      label: "Imagery"
```
#### After (Updated Configuration):
```
legend:
  type: text
```
Would it be better if we could just remove the swatch and keep the legend title instead of removing the legend all together. This would help users who are accessing the story to know what imagery they are look at on the map. Removing the legends entirely could create a sense of unknowingness with what is shown on the map. (implemented in this pull request)

### Validation / Testing
1. In the .mdx file 
- Find the legend configuration for any layers that were previously set as type: categorical or type: graphical.
- Update type to text for legends that do not require a color swatch.
- Remove unnecessary properties like stops, color, and label.
3. Go to https://www.earthdata.nasa.gov/dashboard/stories/blizzards or https://www.earthdata.nasa.gov/dashboard/stories/2024-tornadoes
4. Go to Scrollytelling section, and notice that the legends are populated even after we scroll through block that does not have a legend. 

> [!NOTE]  
> The use case for this legend type is for satellite imagery, where we'd still want the information of what is being shown, but don't need to show colors

### DIFF
--- a/parcel-resolver-veda/index.d.ts
+++ b/parcel-resolver-veda/index.d.ts
@@ -1,3 +1,5 @@
+// import { LayerLegendText } from '$types/veda';
+
 declare module 'veda' {
   import * as dateFns from 'date-fns';
   import { MDXModule } from 'mdx/types';
@@ -72,7 +74,7 @@ declare module 'veda' {
     basemapId?: 'dark' | 'light' | 'satellite' | 'topo';
     type: DatasetLayerType;
     compare: DatasetLayerCompareSTAC | DatasetLayerCompareInternal | null;
-    legend?: LayerLegendCategorical | LayerLegendGradient;
+    legend?: LayerLegendCategorical | LayerLegendGradient | LayerLegendText;
     analysis?: {
       metrics: string[];
       exclude: boolean;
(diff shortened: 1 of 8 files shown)


parcel-resolver-veda/index.d.ts
```diff
@@ -1,3 +1,5 @@
+// import { LayerLegendText } from '$types/veda';
+
```
slesaad, line 2: remove this comment?

slesaad:
can you add an example in the mock dataset, so we can validate it in the deploy preview?

slesaad (approved):
Looks good to me. But would be nice to also be reviewed by @hanbyul-here or someone from the ui team.

AliceR (approved):
Hi @vidhi04bhatia - Thank you for improving the legend! I agree with you that it makes sense to keep the legend to display a title for those satellite layers. It surely does look better than having the transparent `rgba(0, 0, 0, 0)` bar in those cases.
I tried the new category with a couple of stories in veda-config and it all looks good to me!

Issue #1506, opened 2025-03-12
earth.gov Themes Cards / topic cards


aboydnw:
### Context

Themes will be shown using this simple vertical card. Below is a screenshot of the [earth.gov design](https://www.figma.com/design/zkctHQ9OkISqhPqCzPSULj/Earth.gov-Design-System?node-id=356-3787&m=dev) and the similar design from our [template instance](https://www.figma.com/design/PG7AhRloW6qJd9U00VdQ77/VEDA-Template-Instance?node-id=996-8300&p=f&t=Q0aXfLnUezbTzEd9-0). The goal is to have a single card style in veda-ui that can be used across instances (assuming these designs are similar enough). Fausto and Esther working on unifying the design, so subject to change.

earth.gov: 

![Image](https://github.com/user-attachments/assets/85919407-bda8-4629-b93d-c7e580d2290e)

template instance:

![Image](https://github.com/user-attachments/assets/777a88f9-51d5-4af0-b121-01eed6faee34)

### Acceptance Criteria
- [ ] Card layout matches design
- [ ] Card hover and selected states match designs

We will worry about how the information on these cards gets defined in other implementation tickets. For example, earth.gov might use these cards to provide links to stories. So we might pull story titles from mdx files. But there will be other uses of the card that might not pull from any mdx files and will be manually defined.

PR #1552, opened 2025-03-25
feat: Add a Topic card

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1506

### Description of Changes

- Implemented new Topic card component
- Organized sandbox by separating legacy cards from USWDS components
- The card is responsive and adapts to the grid layout
- Added two display variants: full-background topic card and bordered topic card

I thought long about naming conventions for our card components and found it difficult coming up with a truly generic name for this case. I considered "MediaCard", but it could practically apply to several other cards as well if you look from a usage (dev) perspective. Also, what would we call the non-full background variant for the Topic Card if the fullBg is called MediaCard?

I ultimately chose "TopicCard" to follow a use-case based naming approach rather than describing visual structure. I think this semantic approach more clearly communicates the component's purpose rather than just its appearance, and it's unique to VEDA's new content types concept. This will also scale well as we develop the other card types for stories, widgets and other content types across the platform. Happy to hear other opinions here.

### Validation / Testing

- Review the sandbox usage of the card https://deploy-preview-1552--veda-ui.netlify.app/sandbox/uswds-cards
- Verify proper spacing, typography and interaction matches the design specs

Implementation:

Screenshot 2025-03-25 at 13 50 19

Design:

Screenshot 2025-03-25 at 13 50 35


sandrahoang686 (changes requested):
Hey @dzole0311, thanks for tackling! I had some thoughts but I think it might've just been easier for me to PR to yours. Could you take a look at some of my thoughts and implementation here? https://github.com/NASA-IMPACT/veda-ui/pull/1553

AliceR (commented):
Great job on the structure of this PR! I only have a couple of minor nitpick comments below.

I haven’t yet done a design review or fully considered the broader implications of the card components. Good thoughts on the card naming... this is maybe something to even discuss on the design side already 🤔 

Have you thought about adding a few tests? As the complexity grows, having tests in place would be really helpful for documenting the logic and ensuring everything works as expected - especially when we refactor in the future 😅.

app/scripts/components/common/card/uswds-cards/base-card.tsx
```diff
@@ -0,0 +1,61 @@
...
+  isCardFocusable
+}: BaseCardProps) {
+  return (
+    <USWDSCardComponent
+      id={id}
+      layout={layout}
+      gridLayout={gridLayout}
+      style={style}
+      className={className}
+      tabIndex={isCardFocusable ? 0 : undefined}
```
dzole0311, line 33: I've added a new prop `isCardFocusable` to the `BaseCard` that supplies a `tabIndex` value to the USWDS card. We'll have some cards whose whole card area should be clickable, and it should have a correct border outline (with the proper border radius applied).

If the `tabIndex` is not set in the card like this, then the focus outline won't be properly applied around the cards height/width and will lack the correct border radius.

sandrahoang686 (approved):
🙇🏼‍♀️

Issue #1463, opened 2025-02-18
Support additional time density values


anayeaye:
## What
Currently day, month, and year time steps are supported as time density values. We need to support new intervals
hour
half-hour / 30-min

## Related pyarc2stac
https://github.com/orgs/NASA-IMPACT/projects/51?pane=issue&itemId=98240073&issue=NASA-IMPACT%7Cpyarc2stac%7C21

## Current need
`GPM IMERG Final Precipitation (mm/hr)` items are periodically spaced in 30 minute intervals. Currently this collection has to be configured as non periodic to represent this higher temporal resolution data but the summaries object is excessive.
https://dev.openveda.cloud/stac/collections/gpm_imerg-cyclone-beryl

anayeaye:
cc @slesaad, I think we could expand the supported dashboard:time_density intervals for dashboard:is_periodic but regardless of how we decide to handle time steps in the dashboard extension, this 30 minute collection should probably adopt the same pattern that is implemented for the pyarc2stac collections

Also @smohiudd maybe we should go ahead and make the dashboard extension official?

aboydnw:
similar ticket: #1050

PR #1560, opened 2025-03-27
feat: Support custom time interval

slesaad:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1463

### Description of Changes
To support custom time steps (not just 1 day, 1 month, 1 year), we're adding a `dashboard:time_interval` key to STAC collections. This is in the [ISO time duration format](https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm). 

This PR adds support for this new key while still being backward compatible.

Adds a function to generate valid dates based on the start/end/and the various time intervals.

N/A

### Validation / Testing
1. Make sure the generated dates in the timeline are correct for datasets with or without `dashboard:time_interval` value and that the layers load correctly

This preview from [this other PR](https://github.com/NASA-IMPACT/veda-ui/pull/1565) shows that it works for a P7D dataset: https://deploy-preview-1565--veda-ui.netlify.app/exploration?search=&datasets=%5B%7B%22id%22%3A%22esi_4wk_layer%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%2C%22scale%22%3A%7B%22min%22%3A0%2C%22max%22%3A1%7D%7D%7D%5D&taxonomy=%7B%7D&date=2001-01-23T06%3A00%3A00.000Z&dateRange=

### DIFF
--- a/app/scripts/components/exploration/data-utils.ts
+++ b/app/scripts/components/exploration/data-utils.ts
@@ -1,6 +1,9 @@
+import add from 'date-fns/add';
 import eachMonthOfInterval from 'date-fns/eachMonthOfInterval';
 import eachDayOfInterval from 'date-fns/eachDayOfInterval';
 import eachYearOfInterval from 'date-fns/eachYearOfInterval';
+import isBefore from 'date-fns/isBefore';
+import isEqual from 'date-fns/isEqual';
 import startOfDay from 'date-fns/startOfDay';
 import startOfMonth from 'date-fns/startOfMonth';
 import startOfYear from 'date-fns/startOfYear';
@@ -90,7 +93,7 @@ export function resolveLayerTemporalExtent(
   datasetId: string,
   datasetData: StacDatasetData
 ): Date[] {
-  const { domain, isPeriodic, timeDensity } = datasetData;
+  const { domain, isPeriodic, timeInterval } = datasetData;
 
   if (!domain || domain.length === 0) {
     throw new Error(`Invalid domain on dataset [${datasetId}]`);
@@ -98,26 +101,54 @@ export function resolveLayerTemporalExtent(
 
   if (!isPeriodic) return domain.map((d) => utcString2userTzDate(d));
 
-  switch (timeDensity) {
-    case TimeDensity.YEAR:
-      return eachYearOfInterval({
-        start: utcString2userTzDate(domain[0]),
-        end: utcString2userTzDate(domain[domain.length - 1])
-      });
-    case TimeDensity.MONTH:
-      return eachMonthOfInterval({
-        start: utcString2userTzDate(domain[0]),
-        end: utcString2userTzDate(domain[domain.length - 1])
-      });
-    case TimeDensity.DAY:
-      return eachDayOfInterval({
-        start: utcString2userTzDate(domain[0]),
-        end: utcString2userTzDate(domain[domain.length - 1])
-      });
+  const start = utcString2userTzDate(domain[0]);
+  const end = utcString2userTzDate(domain[domain.length - 1]);
+
+  if (!timeInterval) {
+    throw new Error(
+      `Missing time interval for periodic dataset [${datasetId}]`
+    );
+  }
+
+  const intervalDuration = timeInterval.toUpperCase();
+
+  switch (intervalDuration) {
+    case 'P1Y': // Every year
+      return eachYearOfInterval({ start, end });
+    case 'P1M': // Every month
+      return eachMonthOfInterval({ start, end });
+    case 'P1D': // Every day
+      return eachDayOfInterval({ start, end });
     default:
+      if (intervalDuration.startsWith('P')) {
+        return generateDates(start, end, intervalDuration);
+      }
       throw new Error(
-        `Invalid time density [${timeDensity}] on dataset [${datasetId}]`
+        `Unsupported time interval [${timeInterval}] on dataset [${datasetId}]`
       );
+  }
+
+  function generateDates(
+    start: Date,
+    end: Date,
+    interval: string
+  ): Date[] {
+    const match = interval.match(
+      /P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?/
+    );
+    if (!match) throw new Error("Invalid ISO duration");
+
+    const [, years, months, weeks, days, hours, minutes, seconds] = match.map(v => (v ? parseInt(v) : 0));
+
+    let currentDate = start;
+    const validDates: Date[] = [];
+
+    while (isBefore(currentDate, end) || isEqual(currentDate, end)) {
+        validDates.push(currentDate);
+        currentDate = add(currentDate, { years, months, weeks, days, hours, minutes, seconds });
+    }
+
+    return validDates;
   }
 }
 
--- a/app/scripts/components/exploration/data-utils.ts
+++ b/app/scripts/components/exploration/data-utils.ts
@@ -101,12 +101,6 @@ export function resolveLayerTemporalExtent(
   const start = utcString2userTzDate(domain[0]);
   const end = utcString2userTzDate(domain[domain.length - 1]);
 
-  if (!timeInterval) {
-    throw new Error(
-      `Missing time interval for periodic dataset [${datasetId}]`
-    );
-  }
-
   const intervalDuration = timeInterval.toUpperCase();
 
   if (intervalDuration.startsWith('P')) {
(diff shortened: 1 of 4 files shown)


app/scripts/components/exploration/data-utils.ts
```diff
@@ -90,35 +90,78 @@ export function resolveLayerTemporalExtent(
...
-        `Invalid time density [${timeDensity}] on dataset [${datasetId}]`
-      );
+  const start = utcString2userTzDate(domain[0]);
+  const end = utcString2userTzDate(domain[domain.length - 1]);
+
+  if (!timeInterval) {
+    throw new Error(
+      `Missing time interval for periodic dataset [${datasetId}]`
+    );
+  }
```
sandrahoang686, line 108: i'd move above line 101 and throw right away before having to do that utc date conversion

app/scripts/components/exploration/data-utils.ts
```diff
@@ -90,35 +90,78 @@ export function resolveLayerTemporalExtent(
...
+  const intervalDuration = timeInterval.toUpperCase();
+
+  if (intervalDuration.startsWith('P')) {
+    return generateDates(start, end, intervalDuration);
+  } else {
+    throw new Error(
+      `Unsupported time interval [${timeInterval}] on dataset [${datasetId}]`
+    );
+  }
+}
```
sandrahoang686, line 119: unfortunately none of these good informative error messages will see light and be ignored due to the error being caught and a different error message will be used instead of the original :(. https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/hooks/use-stac-metadata-datasets.ts#L73-L80. To make this PR more useful, can you update to make the general error message and the actual error message you've added to this logic display? And we need to come up with a better error handling system in veda-ui ...

sandrahoang686 (approved):
manually validated. other than some comments, lgtm.

app/scripts/components/exploration/data-utils.ts
```diff
@@ -90,35 +90,78 @@ export function resolveLayerTemporalExtent(
...
+  if (!match) throw new Error('Invalid ISO duration');
+
+  const [, years, months, weeks, days, hours, minutes, seconds] = match.map(
+    (v) => (v ? parseInt(v) : 0)
+  );
+
+  let currentDate = start;
+  let validDates: Date[] = [];
+
+  while (isBefore(currentDate, end) || isEqual(currentDate, end)) {
```
hanbyul-here, line 152: In theory, it should be always `isEqual(currentDate, end)` if an appropriate `time-interval` is provided and `endDate` is correct?

app/scripts/components/exploration/data-utils.ts
```diff
@@ -90,35 +90,78 @@ export function resolveLayerTemporalExtent(
...
+ * @example
+ * ```typescript
+ * const start = new Date('2023-01-01');
+ * const end = new Date('2023-01-10');
+ * const interval = 'P1D'; // 1 day interval
+ * const dates = generateDates(start, end, interval);
+ * console.log(dates); // [2023-01-01, 2023-01-02, ..., 2023-01-10]
+ * ```
+ */
+function generateDates(start: Date, end: Date, interval: string): Date[] {
```
hanbyul-here, line 139: I think this is a great candidate for unit tests (and we already have a test file for these utils !

app/scripts/components/exploration/hooks/use-stac-metadata-datasets.ts
```diff
@@ -94,9 +94,13 @@ async function fetchStacDatasetById(
...
   );
 
+  const timeDensity =
+    data['dashboard:time_density'] || time_density || TimeDensity.DAY;
   const commonTimeseriesParams = {
     isPeriodic: !!data['dashboard:is_periodic'],
-    timeDensity: data['dashboard:time_density'] || TimeDensity.DAY
+    timeDensity: timeDensity,
+    timeInterval:
+      data['dashboard:time_interval'] || `P1${timeDensity[0].toUpperCase()}`
```
hanbyul-here, line 103: Is the plan to deprecate `dashboard:time_density`?

app/scripts/components/exploration/data-utils.ts
```diff
@@ -90,35 +90,78 @@ export function resolveLayerTemporalExtent(
...
-        end: utcString2userTzDate(domain[domain.length - 1])
-      });
-    default:
-      throw new Error(
-        `Invalid time density [${timeDensity}] on dataset [${datasetId}]`
-      );
+  const start = utcString2userTzDate(domain[0]);
+  const end = utcString2userTzDate(domain[domain.length - 1]);
+
+  if (!timeInterval) {
```
hanbyul-here, line 104: per type definition: https://github.com/NASA-IMPACT/veda-ui/pull/1560/files#diff-a84645a045108f129bae11ffbb6d29f9b1c8744ae7a4f63b4db872211394889bR19 (And considering we are subbing this already in https://github.com/NASA-IMPACT/veda-ui/pull/1560/files#diff-b0d6bd8ff29016fd81054f35bfcc97dd5b50077310a81f6043a99240cb43a4eaR103 ) I don't think there will be ever a case that timeInterval is missing.

app/scripts/components/exploration/data-utils.ts
```diff
@@ -90,35 +90,78 @@ export function resolveLayerTemporalExtent(
...
+ * ```typescript
+ * const start = new Date('2023-01-01');
+ * const end = new Date('2023-01-10');
+ * const interval = 'P1D'; // 1 day interval
+ * const dates = generateDates(start, end, interval);
+ * console.log(dates); // [2023-01-01, 2023-01-02, ..., 2023-01-10]
+ * ```
+ */
+function generateDates(start: Date, end: Date, interval: string): Date[] {
+  const match = interval.match(
```
hanbyul-here, line 140: I am surprised that date-fns doesn't offer any out-of-box solution for parsing isoduration 🤔

hanbyul-here (commented):
I have some questions and it would be great to have some tests for the new date handling method. 

This should not be a blocker, but to leave a note: I think we are actually resolving the temporal layer in two places, first in `fetchStacDatasetById` and then in `resolveLayerTemporalExtent`. `fetchStacDatasetById` is doing much more than fetching the data (introduced by precedent handlings of timestamps for different datasets). We probably want to move this logic to somewhere like`reconcileQueryDataWithDataset` in data-utils.

app/scripts/components/exploration/data-utils.ts
slesaad: true

app/scripts/components/exploration/hooks/use-stac-metadata-datasets.ts
slesaad: > Is the plan to deprecate `dashboard:time_density`?

yes, but incrementally.

Hopefully this takes care of that https://github.com/NASA-IMPACT/veda-ui/pull/1560/commits/937d7cd7fb84bfa3d6d13bf351aa4e0098f3c41d

app/scripts/components/exploration/data-utils.ts
slesaad: i'm not sure i understand this question. do you mean the ending condition of the loop will be that?
if so, yes, it should be.
but we still need to `isBefore` condition for all the other dates.

app/scripts/components/exploration/data-utils.ts
slesaad: i know 😭

app/scripts/components/exploration/data-utils.ts
slesaad: absolutely! i added some tests [here](https://github.com/NASA-IMPACT/veda-ui/pull/1560/commits/98d8cc57d6afba8e56b427dee2b8b0fa34cdb5db).

i am having some trouble with some of the logic tho, so i'd appreciate your input.

the test fails at these two lines (because "month" is kinda ambiguous, and date-fns handles it a little weird):

1. https://github.com/NASA-IMPACT/veda-ui/pull/1560/commits/98d8cc57d6afba8e56b427dee2b8b0fa34cdb5db#diff-a3254c24152bd7254aa58ddf1399a606c62ee6a460a20391eb882a564f50e43cR144

2. https://github.com/NASA-IMPACT/veda-ui/pull/1560/commits/98d8cc57d6afba8e56b427dee2b8b0fa34cdb5db#diff-a3254c24152bd7254aa58ddf1399a606c62ee6a460a20391eb882a564f50e43cR314

not sure what the course of action should be here?
should we only allow the API to have day based interval (and omit month because it can be ambiguous? what about year)?

hanbyul-here: Is this still a problem? I am not seeing any test failing.

hanbyul-here (approved):
One issue worth being aware is that the date that a user sees, won't be the date of the data. (Data request will fetch the most relevant date, but UI shows the selected date - related part: https://github.com/NASA-IMPACT/veda-ui/blob/955fdaf9af3c06b3ea92106bc9163c731b1911b0/app/scripts/components/exploration/components/map/layer.tsx#L31) As far as I can see, this doesn't seem to be a regression?(non-periodic data is working in this way anyway?) - Anyway, it will be worth having a ticket to handle this. 

Screenshot 2025-04-15 at 10 49 29 AM

app/scripts/components/exploration/data-utils.ts
slesaad: apparently not lol.

PR #1563, opened 2025-03-27
fix: Use colormap from legend when no colormap is provided in sourceParams

slesaad:
### Issues
1. when no colormap is provided, the `raster` data type/titiler uses greyscale; but the UI shows the default colormap `viridis`
2. when no colormap is provided because non `raster` data type doesn't use the colormap `sourceParams`, it still uses the default `viridis` colormap instead of reading from the `legend`

### Description of Changes
When no colormap is provided, the colormap is generated from the `legend` property instead of using a default colormap, which doesn't always make sense.

should we always allow "raster" data type to be configurable?

### Validation / Testing
[preview link](https://deploy-preview-1563--veda-ui.netlify.app/exploration?search=&datasets=%5B%7B%22id%22%3A%22casa-gfed-co2-flux%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&date=2017-12-01T06%3A00%3A00.000Z)
[mdx file](https://github.com/NASA-IMPACT/veda-ui/blob/fefe3ed70ad08924565620b40c724d8353e02b23/mock/datasets/data-from-ghg.data.mdx?plain=1#L46-L85)

### DIFF
--- a/app/scripts/components/common/map/layer-legend.tsx
+++ b/app/scripts/components/common/map/layer-legend.tsx
@@ -347,22 +347,21 @@ export const LayerGradientGraphic = (props: LayerLegendGradient) => {
   );
 };
 
-export const LayerGradientColormapGraphic = (props: Omit<LayerLegendGradient, 'stops' | 'type'>) => {
-  const { colorMap = DEFAULT_COLORMAP, ...otherProps } = props;
-  const colormapResult = findColormapByName(colorMap);
+export const LayerGradientColormapGraphic = (props: Omit<LayerLegendGradient, 'type'>) => {
+  const { colorMap, stops: defaultStops, ...otherProps } = props;
 
-  const { foundColorMap, isReversed } = colormapResult;
-  const stops = Object.values(foundColorMap)
-  .filter(value => Array.isArray(value) && value.length === 4)
-  .map((value) => {
-    return `rgba(${(value as number[]).join(',')})`;
-  });
+  const processedStops = React.useMemo(() => {
+    if (!colorMap) return defaultStops;
 
-  const processedStops = isReversed
-  ? stops.reduceRight((acc, stop) => [...acc, stop], [])
-  : stops;
+    const { foundColorMap, isReversed } = findColormapByName(colorMap);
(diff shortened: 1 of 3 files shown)


sandrahoang686:
The original PR where we default to colormap `viridis` [here](https://github.com/NASA-IMPACT/veda-ui/pull/1117).

dzole0311 (approved):
Thanks @slesaad! 

I think the intended behavior was that if no colormap is provided in sourceParams or from the STAC query, we’d default to `viridis` when requesting tiles, like:

`https://earth.gov/ghgcenter/api/raster/searches/3b46d29dcb49fd2eb8e4a65b5dcde805/tiles/WebMercatorQuad/1/1/1?assets=npp&rescale=0%2C0.3&colormap_name=viridis`

It looks like the `colormap_name` query param (along with the default) is currently missing, so titiler responded with the default grayscale.

> should we always allow "raster" data type to be configurable?

Could you clarify what kind of configurability you’re thinking about? On the configuration topic, I'm still a bit unclear on where we want to draw the line between using backend-rendered props as the source of truth versus enabling mdx/user-level overrides. It feels like something we need a broader alignment with the data services team ideally including you as well.

I was also re-reading an earlier thread between you and @hanbyul-here [in PR #1462](https://github.com/NASA-IMPACT/veda-ui/pull/1462#discussion_r2009458837) and this suggestion by Hanbyul might be a good point of discussion:

> The best approach might be to allow users to overwrite only specific attributes rather than the entire source parameter object - so they don't have to scratch their heads when they just want to see the data in a different colormap.

Issue #1523, opened 2025-03-18
Use widget component to wrap existing components in stories


AliceR:
### Context

Following up on #1281 this ticket aims to put the full screen view into production instances, wrapped around existing widgets.

### Acceptance Criteria
- [ ] Map, Map Comparison, and Charts are all wrapped in the widget frame developed in #1281 
- [ ] Users can expand the widget to see in a full screen view

The assumption here is that the widgets themselves won't function any differently, including in full screen view. Changes to functionality and additions to the full screen view will come later.

aboydnw:
@AliceR I added some details, let me know if you had a different understanding of this issue.

Do you have a good sense on LOE here? I think it might be a nice break point for the work you did on #1281 because then it's out in the world providing value, and we can get feedback on it. But, I'm worried that it's going to be a lot of work and cause layout issues across stories/dataset overview pages. 

So, if you feel good about it then I'd say we pull it in. Otherwise, I'd rather shift focus to other things than get caught playing whack-a-mole with styling/layout issues with these components.

(is the whack-a-mole reference uniquely an American thing? 😅 )

AliceR:
Thats great @aboydnw, thanks for refining the ticket! 
One more useful piece of information would be some guidance on where exactly to place the fullpage widget. Is there a specific story you have in mind that we can test it on? Maybe you can list the one or several places where you would want to see it in action?

aboydnw:
At the end of this story, I was imagining every use of the map, map comparison, and chart widgets have the full screen functionality. I guess I wasn't seeing this as a new "widget" but just as a step towards redefining what a widget is in VEDA.

PR #1564, opened 2025-03-27
feat: Wrap components with Widgets

AliceR:
**Related Ticket:** #1523

### Description of Changes
This change exports the Fullpage Modal Widgets as `<Widget />` from the library and adds it to the MDX Content components. The `<Widget />` is then used in some .mdx datasets and stories to wrap around the Map block components.

Due to the constraints of our current MDX parser, I couldn't wrap the entire `<Figure />` component with `<Widget />`, which would have been the preferred approach. Instead, as a workaround, I placed `<Widget />` inside `<Figure />`, wrapping all of its content. While this approach works, it introduces an issue with the absolutely positioned FigureAttribution (the small "i" icon in the top-right corner), which requires a relatively positioned parent. To address this, I added an extra `<div className="position-relative">`.

**Note**: I intentionally did not add styles to the <Widget /> component just to accommodate the way <Figure /> is structured.

Additionally, I updated the header tag from h6 to h4, following design instructions.

### Questions About Changes
1. Widget Composition:

- Should the Widget be an optional wrapper that MDX creators can choose to add?
- Or should the Widget be a fixed element within certain components (e.g., <Figure />) so that it always gets included when those elements are used?

2. MDX Parser Adjustments:

- If we go with the first option (making Widget optional), should we modify our MDX parser to allow <Widget /> as a child element?
- Or, should <Widget /> wrap only the content inside <Figure /> instead of the entire <Figure /> component anyways? 

3. Future Direction:

- Are we keeping <Figure /> and other legacy MDX components long-term, or are they expected to be replaced with USWDS components eventually?

Would love feedback on these decisions before finalizing the approach!

### Validation / Testing
See how widgets appear on https://deploy-preview-1564--veda-ui.netlify.app/data-catalog/nighttime-lights or https://deploy-preview-1564--veda-ui.netlify.app/stories/air-quality-and-covid-19

### DIFF
--- a/app/scripts/components/common/mdx-content.js
+++ b/app/scripts/components/common/mdx-content.js
@@ -20,6 +20,7 @@ import {
 } from '$components/common/blocks/lazy-components';
 import { NotebookConnectCalloutBlock } from '$components/common/notebook-connect';
 import SmartLink, { CustomLink } from '$components/common/smart-link';
+import { Widget } from '$libs';
 
 function MdxContent(props) {
   const pageMdx = useMdxPageLoader(props.loader);
@@ -46,7 +47,8 @@ function MdxContent(props) {
           Link: SmartLink,
           a: CustomLink,
           Table: LazyTable,
-          Embed: LazyEmbed
+          Embed: LazyEmbed,
+          Widget
         }}
       >
         <pageMdx.MdxContent {...(props.throughProps || {})} />
--- a/app/scripts/components/common/widget/__snapshots__/widget.test.tsx.snap
+++ b/app/scripts/components/common/widget/__snapshots__/widget.test.tsx.snap
@@ -9,9 +9,9 @@ exports[`FullscreenWidget matches snapshot 1`] = `
     class="padding-2 display-flex flex-justify space-between flex-align-center"
     data-testid="widget-header"
   >
-    <h6>
+    <h4>
       Test Fullscreen Widget
-    </h6>
+    </h4>
     <button
       class="usa-button usa-button--outline margin-0 margin-left-3"
       data-testid="widget-toggle"
--- a/app/scripts/components/common/widget/header.tsx
+++ b/app/scripts/components/common/widget/header.tsx
@@ -48,7 +48,7 @@ const Header = ({
       className='padding-2 display-flex flex-justify space-between flex-align-center'
       data-testid='widget-header'
     >
-      {isExpanded ? <h1>{heading}</h1> : <h6>{heading}</h6>}
+      {isExpanded ? <h1>{heading}</h1> : <h4>{heading}</h4>}
 
       <USWDSButton
         type='button'
--- a/app/scripts/libs/index.ts
+++ b/app/scripts/libs/index.ts
@@ -61,7 +61,8 @@ export {
   PageHeader,
   PageFooter,
   DefaultCard,
-  FlagCard
+  FlagCard,
+  Widget
 } from './uswds-components';
 
 export { PageHero, LegacyGlobalStyles };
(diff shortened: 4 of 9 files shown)


dzole0311 (approved):
Thanks @AliceR! Approving the PR and adding just a few thoughts:

- Pressing Esc could close the widget modal for better UX?
- Noticed in the designs: should the "Exit fullscreen" button be hidden while in fullscreen? Although that might not make sense design-wise if we go the modal approach
- Looking into the whole widget concept again, more I have the feeling that it might work better as its own page (e.g. routing to /widget/widget-page on "Enter fullscreen"). That also fits the idea of tools (fire tool) being standalone pages with consistent topbar branding and it’d help with SEO too (cc @aboydnw @faustoperez )

> Are we keeping other legacy MDX components long-term, or are they expected to be replaced with USWDS components eventually?

Do you mean here components like `ScrollytellingBlock`, `Chapter`, `Map`? I think they’ll need to be updated to match USWDS styles.

AliceR:
Tagging @faustoperez for thoughts on UX and design. 
I generally agree with the approach to route to `/widget/widget-page`, but I think it can be incrementally added, as we set up the concept of standalone widgets/tools.

aboydnw:
I agree that the URL changes can be incrementally added. We can add those to the Widgets initiative in the backlog.

I have two questions/comments:
1. When I view this on a large screen, the map stretches horizontally but not proportionally in the vertical space. So, I'm left with a lot of whitespace. This is not so much an issue on a normal laptop screen, though, so if it's not a quick fix then probably not worth our time at the moment.
2. How are titles defined? Is this a new field in the mdx file? I think the titles I see in the preview might a sticking point for people on live instances, especially GHG and earth.gov. I hadn't thought about this as new information until now, but we'll likely need a way to have more readable titles than the layer id and dates.

image

AliceR:
Thanks for the feedback! Regarding your points:

1. The map component has a fixed height. We might want to consider improving our map component by making it fill the available space. Not related to the widget though!
2. Titles are passed in as a prop to the widget component. I just put there some random combination of info I found on the map, but If you want, you can make a suggestion here: https://github.com/NASA-IMPACT/veda-ui/pull/1564/files#diff-85157275f2a02b73d742bc0fd1a381c37ed263ac68ca4c3ccf132fa82ae4dd62R218
Try to comment on that line, and click on the icon that says "Add a suggestion, <Ctrl+g>" (or actually press <Ctrl+g> 😄), and replace the heading="Your suggestion here!". 

So, it is clear that this feature will require more thought eventually, to match what @faustoperez has in mind for the concepts of widgets. I'll merge this PR now, knowing that it does not actually changes any public pages - but just offers a new component that we can build on later.

faustoperez:
> Tagging @faustoperez for thoughts on UX and design.
> I generally agree with the approach to route to /widget/widget-page, but I think it can be incrementally added, as we set up the concept of standalone widgets/tools.

Adding the URL changes incrementally also make sense to me.

When we do it, we need to thing about on how which button/label we use to close the modal, as the widget page could be accessed from any point as it is not a modal anymore. Maybe an X primary outline icon button would work in all cases, working as a browser `back` button.

Maybe we could also reuse the tool header component. These are the two options we could consider, what do y'all think? @AliceR @aboydnw 

image

aboydnw:
@faustoperez I might be biased by my belief that, for the foreseeable future, the most common pathway for users to come to this page will be through a story. So, in that context, I think that the second option would less confusing to users than if they saw the full header with potentially a different nav item highlighted. The second option feels more like a "I just entered full screen mode" whereas the first option feels more like a "I just went to a new page" type of thing.

I do like the larger title in the first option, though.

Issue #1525, opened 2025-03-18
Update release cadence to weekly


aboydnw:
## Context

Idea is to use the same release process, but update the cadence to be weekly on Monday

## Acceptance Criteria

- [ ] Given when then
- [ ] Given when then

## Related Tickets

[If applicable, link any tickets that are related]

PR #1572, opened 2025-04-01
ci: Release action follow-up

hanbyul-here:
**Related Ticket:** #1525
- [x] Delete v6.2.0

### Description of Changes
- Removing the condition that was blocking the weekly release 
- Simplify Regex (We don't need merge.. and initial commit for our change logs)
- Edit notification text 

**This change should be accompanied with repo settings** 
- It seems that requiring staus check makes force-push fail. To allow github action to bypass this rule, we need to use Ruleset (Settings > Rules > Rulesets), not the branch protection rule. (I already went through this process.)

### Validation / Testing
I applied the same ruleset for this branch, and tested the release action: https://github.com/NASA-IMPACT/veda-ui/releases/tag/v6.2.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,4 +79,4 @@ jobs:
               - type: "section"
                 text:
                   type: "mrkdwn"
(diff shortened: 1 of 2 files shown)


dzole0311 (approved):
Thanks @hanbyul-here

Issue #1578, opened 2025-04-02
[Bug] Non-VEDA apps under same domain incorrectly treated as internal routes


dzole0311:
### Context

Some apps are hosted under the same domain as VEDA but are not part of the VEDA project, for example, veda-config-eic is hosted under earth.gov vs. the Mobile Climate Mapper at earth.gov/mobile-climate-mapper.

For cases like these, we aim to:

1. Avoid showing an external link badge in the cards for these apps (achieved via a `hideExternalLinkBadge` prop)

2. We want VEDA UI not to treat their URLs as internal routes

There has been a regression on point 2, so the link to VEDA UI now treats these URLs as internal.

<img width="835" alt="Image" src="https://github.com/user-attachments/assets/45535fb2-8dea-4a35-a07d-9b192a9dd30b" />

URL: 

Blocking a release? (y/n): 

Regression issue? (y/n): y
[If yes, add Regression label]

Priority: 

### Steps to reproduce the issue

1. [Step 1 to reproduce]
2. [Step 2 to reproduce]

PR #1579, opened 2025-04-02
fix: use utils.isExternalLink logic for same-domain external apps

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1578

### Description of Changes

See issue above for context.

- Refactored to use `utils.isExternalLink`
- Removed the previous util from url.ts

### Validation / Testing

1. Pull branch locally in veda-ui
2. Go to mock/stories/external-link-example.stories.mdx
3. Change the url prop from to https://developmentseed.org to http://localhost:9000/stories 
4. Click on the card, it should open the url and in a new tab
5. Switch to the main branch
6. Test the same
7. Once you do step 4, notice how the link is broken (it appends the url to the current url)

### DIFF
--- a/app/scripts/context/veda-ui-provider.tsx
+++ b/app/scripts/context/veda-ui-provider.tsx
@@ -1,6 +1,6 @@
 import React, { createContext, ReactNode, useContext } from 'react';
 import { DATASETS_PATH, STORIES_PATH } from '$utils/routes';
-import { isExternalLink } from '$utils/url';
+import * as utils from '$utils/utils';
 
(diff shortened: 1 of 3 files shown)


AliceR (approved):
Would be nice to have a test for the url / internal vs. external logic, this would also help in documenting the expected behavior.

dzole0311:
@AliceR Added a very basic test for `isExternalLink`, but mainly for the comment explaining why we only check for protocol, and not domain

PR #1592, opened 2025-04-07
refactor: Clean up USWDS directory and imports

AliceR:
**Related Ticket:** no ticket

### Description of Changes
As part of a bigger cleanup initiative I would like to propose that we move the /uswds directory outside of the current components/common/... structure. This is essentially just a renaming of files, and updating the imports that were not previously referencing the $uswds alias. The reason bring just general cleanup and better separation.

There should be no changes to any rendering or build processes! It is just an organizational change.

💖 For easier review, please inspect the changes commit by commit!!

### Validation / Testing
- [ ] Validate that the change does not affect development nor build of the library!


vgeorge:
@AliceR This was not introduced in this PR, but I'm curious about why do we need to rename components from USWDS instead of just using the directly. In the `uswds` folder all the components are just passing props to a prefixed `USWDS*` component.

sandrahoang686 (approved):
I'm fine with this and have no strong opinions on if we move the `uswds` dir outside of the `common` dir. But I did tag @hanbyul-here too in case she might have a stronger opinion.

hanbyul-here:
@vgeorge  - If I remember correctly (please correct me if I am wrong) , the initial intention of having mid-layer was that even if we change the library in the future again, the end component doesn't have to change import syntax. If we look at the current context, let's say we want to change Button (which is from ui library), if the Button was encapsulated in the separate file, we can do whatever prop change in that midlayer and pass it to the component from the new library. The initial thought is documented in this PR: https://github.com/NASA-IMPACT/veda-ui/pull/1031#issuecomment-2245278125

@AliceR Since it is a PR that mainly changes the paths, I am guessing what you are looking for from review is an opinion . It is a bit difficult to form an opinion on 'cleaning'. Can you list more of the benefits we can get from this folder structure? What is the bigger initiative that we should foresee from this change?

AliceR:
@hanbyul-here Thanks for the question! The reason for moving the uswds folder is to better organize our project and improve maintainability. Right now, it’s mixed with legacy files and newer components, making it harder to navigate and find the right pieces. By separating it out, we clarify the distinction between system dependencies and domain-specific components, which improves discoverability and reduces cognitive load.

This is just a small first step—future changes will include auditing legacy files, refining the folder structure, and potentially documenting conventions for better consistency moving forward.

Happy to discuss further if needed!

vgeorge (approved):

Issue #1370, opened 2025-01-13
Documentation website


vgeorge:
## Context

Currently, this component library doesn't have a dedicated documentation website, relying on Markdown files referenced in the README.

I believe a public documentation website would make it easier for users to find guides, APIs, and examples. At the same time, we should aim to minimize additional effort in maintaining documentation.

I looked into [documentation.js](https://documentation.js.org/) and [Storybook](https://storybook.js.org/) and suggest using the latter for this project:

- Offers a sandboxed environment to demo and test components in isolation
- [Generates documentation](https://storybook.js.org/docs/writing-docs/autodocs) from JSDoc comments automatically
- [Integrates with Playwright](https://storybook.js.org/docs/writing-tests/import-stories-in-tests/stories-in-end-to-end-tests#with-playwright)
- Already used in [Devseed UI](https://ui.ds.io), leveraging existing team knowledge

Looking forward to hearing thoughts from the team.

## Related Tickets

- #1200
- #1263
- #1197 
- #1368

hanbyul-here:
👋 Thanks for your thought. This would be def step-up from what we have now. 

I can't think of any objection to take more automated approach, and it would help us to keep the documentation in comment more up to date. Also more component based documentation will be necessary as we move veda-ui to registry. There are some elements that I hope to get more a more specific picture. 

There are many non components specific documents sprinkled all over VEDA. (Two examples : https://github.com/NASA-IMPACT/veda-docs, https://github.com/NASA-IMPACT/veda-ui/tree/main/docs)  - Is this documentation will be the central point of VEDA UI? 

Storybook does offer really nice automated features, but as far as I am aware, we still need to put some efforts to make the example work etc. (Please correct me if I am wrong) And I am a concerned about diverging our work even into more items when we already have lots on our plates. Can you give us a rough estimation on the creation/maintenance of Storybook?

vgeorge:
@hanbyul-here thanks for the feedback. Regarding the different documentation sources, it might be helpful to focus on enhancing the existing documentation in this repository, surfacing the markdown files you pointed to and, at a minimum, adding the components' API. I don't think we need the examples if they are too involved to implement.

The veda-docs repository appears to have a broader scope, and it could be beneficial to keep it separate while referencing relevant documentation from this repository as needed.

[...]

PR #1593, opened 2025-04-07
chore: add storybook

AliceR:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1370

### Description of Changes
This is how I imagine our storybook setup to work. Note how we can run the storybook within our project on vite, independent from the legacy build tools.

Find more information here: [core/README.md](https://github.com/NASA-IMPACT/veda-ui/blob/86a1dd5ab8b12ba730b02f93c21f7aa8cc081fe0/core/README.md)

I need to figure out the best way to include our styles into the storybook. Help is welcome!

Next step: Set up testing for the components in storybook, including a coverage report.

### Validation / Testing
- [ ] Run `yarn core:install` and `yarn core:storybook`


vgeorge:
@AliceR the setup of Storybook seems good, I would just suggest using `/storybook` instead of `/core`. This PR contributes to the [documentation website ticket](https://github.com/NASA-IMPACT/veda-ui/issues/1370), where there is a broader discussion on how much we need it and the level of effort to maintain it. It would nice to explore how we could integrate our current Playwright implementation with Storybook, if we have time.

AliceR:
I like the suggestion @vgeorge . 

The reason I went with 'core' was a bigger picture that I had, but only in my mind.
I am thinking that we could migrate the refactored veda-ui components into core/src/components, where we set up a baseline for code quality, test coverage and documentation. 

This is called **The Strangler fig pattern**: Keep the existing project structure with the root package.json serving the legacy application, while still having the core components in a separate directory with its own isolated dependencies.

Benefits

- Clean slate: Modern tooling without legacy constraints
- Isolation: Prevents legacy issues from affecting new components
- Migration path: Creates a clear pathway for eventual full migration
- Documentation: Storybook becomes your "north star" for component design

Potential Challenges

- Duplication: Some code and dependencies might be duplicated
- Integration complexity: Components need to work in both environments
- Build complexity: Managing two build systems in one repo
- Consistency: Ensuring styles/themes match between old and new

I acknowledge that this is a bit too big of a conversation for now, so renaming the directory to storybook and keep it simple for now is the safer way forward :)

AliceR:
TODO:
- [ ] Add ADR about proposed change

vgeorge:
@AliceR, I gave this another try and pushed some changes. I hope it doesn’t break anything for you. I agree that the approach you proposed is better.

### Changes added:

* Downgraded React version to match the version used in veda-ui
* Upgraded Storybook to v9
* Fixed USWDS setup inside Storybook
* Scaffolded Button, Card, and Banner components (only Button is rendering correctly so far)

### Suggested next steps:

* Add a story for a component that uses devseed-ui for setup
* Update `core/README.md` (I think it’s too verbose now)
* Decide if we rename the folder `storybook`, keep it as `core`, or something else

I think this is close to being ready to merge. While I don’t believe we can use it as a documentation website just yet, it’s already useful for viewing components in isolation and could serve as a solid base for migrating from Devseed UI to USWDS with more confidence.

I’ll hold off on adding further changes until we get feedback from the team.

cc @sandrahoang686 @hanbyul-here @snmln @dzole0311 @aboydnw

core/src/stories/uswds-card.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+    </Card>
+  )
+};
+
+export const WithMedia: Story = {
+  render: () => (
+    <Card>
+      <CardMedia>
+        <img
+          src='https://via.placeholder.com/300x200/0071bc/ffffff?text=USWDS+Card'
```
dzole0311, line 40: The placeholder image doesn't seem to work for me. Can we use local placeholders to avoid depending on network calls in the stories?

core/src/stories/uswds-banner.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const Default: Story = {
+  render: () => (
+    <Banner>
+      <BannerHeader
+        isOpen={false}
+        flagImg={<BannerFlag />}
+        headerText='An official website of the United States government'
+        headerActionText="Here's how you know"
+      />
+      <BannerContent isOpen={false}>
```
dzole0311, line 30: I don’t think we should catalog USWDS components in our Storybook, since we already require react-uswds (which has its own Storybook) as a dependency of veda UI. We should only document veda UI components that may or may not wrap or build on USWDS (e.g. the [components listed here](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/libs/index.ts))

dzole0311 (changes requested):

core/src/stories/uswds-button.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+  title: 'Components/Button',
+  component: USWDSButton,
+  tags: ['autodocs']
+};
+
+export default meta;
+type Story = StoryObj<typeof USWDSButton>;
+
+export const Default: Story = {
+  render: () => <USWDSButton type='button'>Click Me</USWDSButton>
```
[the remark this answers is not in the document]
vgeorge: I removed the USWDS components, as suggested. But nice to know about this pattern.

core/src/stories/uswds-banner.stories.tsx
vgeorge: I agree and removed USWDS components.

dzole0311 (approved):
Tested the set up locally, looks good to me! 👍

PR #1594, opened 2025-04-07
ci: Throw error for unused vars

hanbyul-here:
I think unused var is worth throwing an error. Changing eslint config for that rule.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -102,7 +102,7 @@
         "@typescript-eslint/no-invalid-void-type": "warn",
         "@typescript-eslint/no-non-null-assertion": "warn",
         "@typescript-eslint/no-unused-expressions": "warn",
-        "@typescript-eslint/no-unused-vars": "warn"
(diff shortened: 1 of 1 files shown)


hanbyul-here:
🤔 It just struck me that our precommit hook lint and githubaction lint works differently.

AliceR:
We had ESLint issues set to warnings to avoid the overhead of cleaning up the entire codebase. At some point, ESLint was (accidentally?) disabled altogether—see the discussion here when re-enabling it: https://github.com/NASA-IMPACT/veda-ui/pull/1219.

It might be worth considering a clearer separation between legacy code (which we only touch when necessary) and refactored code. We could then enable lint errors for the refactored parts and enforce stricter hygiene standards there—without burdening ourselves with fixing legacy code unless needed.

And yes, setting to 'error' should definitely fail the hook before it even reaches the upstream branch! :)

Issue #1545, opened 2025-03-24
Consolidate Raster based layers


hanbyul-here:
Zarr, CMR uses RasterPaintLayer, which was derived from https://github.com/NASA-IMPACT/veda-ui/issues/1000. 
Eventually, thoses two layers and RasterTimeseries needs to draw raster layer, but RasterTimesereis layer has its own logic for layer rendering now. Consolidating all the layers to use RasterPaintLayer will fix several issues. 

1. https://github.com/NASA-IMPACT/veda-ui/issues/1537 - query parameter parsing logics for titiler parameters. 
2. Only RasterTimeseries is accepting the callback that we need to scrollytelling. (ex. https://github.com/NASA-IMPACT/veda-config/pull/493 in this PR, Derecho story uses `GPM_3IMERGDF.v07` data, which is CMR type)

PR #1595, opened 2025-04-08
feat: Separate Marker Layer from RasterTimeSeries

hanbyul-here:
**Related Ticket:** #1545 

### Description of Changes
Separate the Marker Layer from the RasterTimeSeries. For more context: Raster Timeseries shows markers for the dataset that have minimum zoom set up higher than the current Zoom to indicate to users where the data is. (ex.[ NightLight data](https://deploy-preview-1595--veda-ui.netlify.app/exploration?datasets=%5B%7B%22id%22%3A%22nightlights-hd-monthly%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%2C%22colorMap%22%3A%22inferno%22%2C%22scale%22%3A%7B%22min%22%3A0%2C%22max%22%3A255%7D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2022-02-28T15%3A00%3A00.000Z): ). This PR separates these Marker Layers so that RatserTimeSeries can eventually consist of RasterPaintLayer + MarkerLayer.  I believe the Marker Layer is uncoupled enough to be used outside of the RasterTimeSeries if needed.

- Move useRequestStatus (renamed from useLayerStatus) to the hooks file 
- did some lint cleaning 🧹 

### Validation / Testing
- RasterTimeseries data should work as before. (Whether it has marker layers or not.)

### DIFF
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -502,11 +502,13 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
   useFitBbox(!!isPositionSet, bounds, layerBounds);
 
   return (
-    <PointsLayer
-      id={id}
-      points={points}
-      zoomExtent={zoomExtent}
-      onPointsClick={onPointsClick}
-    />
+    points && (
+      <PointsLayer
+        id={id}
+        points={points}
+        zoomExtent={zoomExtent}
+        onPointsClick={onPointsClick}
+      />
+    )
   );
 }
--- a/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-timeseries.tsx
@@ -1,11 +1,5 @@
 /* eslint-disable react-hooks/exhaustive-deps */
-import React, {
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState
-} from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import qs from 'qs';
 import {
   LayerSpecification,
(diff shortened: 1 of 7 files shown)


app/scripts/components/common/map/style-generators/raster-timeseries.tsx
```diff
@@ -613,5 +501,14 @@ export function RasterTimeseries(props: RasterTimeseriesProps) {
...
+  return (
+    points && (
+      <PointsLayer
+        id={id}
+        points={points}
+        zoomExtent={zoomExtent}
+        onPointsClick={onPointsClick}
+      />
+    )
+  );
```
AliceR, line 513: Can you explain why the points are returned from the raster-timeseries layer?

I may be missing some context, but I think the decision of whether to display points or raster based on zoom level would make more sense at a higher architectural level, before rendering either.
It seems confusing to have one layer returned from another, and leaves the responsibility unclear. A cleaner approach would be to have a decision layer that selects the appropriate visualization method (points or raster) based on zoom level, then routes to either a dedicated points layer or raster layer.

AliceR (commented):
I appreciate the clear separation of concerns between these two layers - it's a thoughtful architectural decision that improves maintainability. 
The comment below is just some food for thought regarding the decision point for rendering...

I noticed a gap where none of the layers are rendering... is that intentional? Since you set ` maxzoom: minZoom,` in the point layer, I wouldn't expect a gap.

AliceR (approved):

app/scripts/components/common/map/style-generators/raster-timeseries.tsx
hanbyul-here: The points are there to indicate where the raster datasets are. When a user is on a lower zoom than the dataset's minimum zoom, the points will show up. From this context that points data is derived from raster timeseries data, I think it makes sense that points are accompanied by raster paint layer, and raster-timeseries is that decision layer to render both of those layers. It would make sense to make the minZoom/maxZoom more explicit as you suggested, and I think it would be on JSX level change (in pseudo code :  `points && map.minZoom < ... && <Points ... ` than having one more layer on top of raster timeseries. Maybe I am looking at this code too long that I am missing something? In any case, let's make it better little by little 🔨

hanbyul-here:
@AliceR that gap has been a mystery for me too!! I shamefully never tired to figure out why that is happening :[

PR #1644, opened 2025-04-30
feat: add wmts connection

ram-sharan25:
Related Ticket: [link](https://github.com/US-GHG-Center/ghgc-architecture/issues/563)

## Description of Changes
•	Add support for rendering WMTS layers using NASA GIBS data in the VEDA UI.  
•	Addition of OMI_Nitrogen_Dioxide_Tropo_Column as a test data.

## Notes & Questions About Changes
 [Google Doc – WMTS Integration Task](https://docs.google.com/document/d/1z54ZHSfs1PGdmL4Muu-NSAzUWnxCisyr5Mx83IKDt4M/edit?tab=t.0#heading=h.xg3fmyyygrbw)

## Validation / Testing
Veda Data Exploration  [OMI_Nitrogen_Dioxide_Tropo_Column](https://deploy-preview-1644--veda-ui.netlify.app/exploration?search=&datasets=%5B%7B%22id%22%3A%22OMI_Nitrogen_Dioxide_Tropo_Column%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&date=2015-09-09T05%3A00%3A00.000Z)

### DIFF
--- a/app/scripts/components/common/map/style-generators/hooks.ts
+++ b/app/scripts/components/common/map/style-generators/hooks.ts
@@ -35,11 +35,14 @@ interface Link {
   type: string;
   title: string;
   'wms:layers': string[];
-  'wms:styles': string[];
+  'wms:styles'?: string[];
+  'wmts:dimensions'?: string[];
+  'wmts:servers'?: string[];
 }
 
-interface WMSResponseData {
+interface WebMapResponseData {
   links: Link[];
+  tilematrixset?: string;
   extent: {
     spatial: {
       bbox: [number, number, number, number][];
@@ -47,24 +50,6 @@ interface WMSResponseData {
   };
 }
 
-interface WMTSLink {
-  href: string;
-  rel: string;
-  type: string;
-  title: string;
-  'wmts:layers': string[];
-  'wmts:dimensions': string[];
-  'wmts:servers': string[];
-}
-
-interface WMTSResponseData {
-  links: WMTSLink[];
-  extent: {
-    spatial: {
-      bbox: [number, number, number, number][];
-    };
-  };
-}
 
 export function useZarr({
   id,
@@ -178,10 +163,15 @@ export function useCMR({
 
   return assetUrl;
 }
-
-export function useWMS({ id, stacCol, stacApiEndpointToUse, onStatusChange }) {
+export function useWebMapService({
+  id,
+  stacCol,
+  stacApiEndpointToUse,
+  onStatusChange
+}) {
   const defaultBounds: [number, number, number, number] = [-180, -90, 180, 90];
-  const [wmsUrl, setWmsUrl] = useState('');
+  const [url, setUrl] = useState<string[]>();
+  const [tilematrixSet, setTilematrixSet] = useState<string>('');
   const [bounds, setBounds] =
     useState<[number, number, number, number]>(defaultBounds);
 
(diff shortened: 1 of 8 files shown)


slesaad:
i wonder if we can consolidate wms and wmts types since they have a lot in common. or do you think it makes sense to keep them separate @sandrahoang686 cc @hanbyul-here

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
```diff
@@ -73,14 +77,22 @@ export function RasterPaintLayer(props: RasterPaintLayerProps) {
       async function run() {
         // Create a modified version of the parameters
         const tileParamsAsString = formatTitilerParameter(updatedTileParams);
-        const tileUrl = `${tileApiEndpoint}?${tileParamsAsString}`;
+        const tileUrl = Array.isArray(tileApiEndpoint)
```
hanbyul-here, line 80: I think you should be able to handle this on parent level, by passing `sourceParamFormatter`. It will be best if raster-paint-layer can stay agnostic to any layer.

hanbyul-here:
I don't have much grasp about the difference between wms and wmts. What is the main difference?

slesaad:
> I don't have much grasp about the difference between wms and wmts. What is the main difference?

The main difference is that wmts is tiled while wms is an individual image. Both are OGC standard and thus we'd wanna support them both.

hanbyul-here:
I think there is a high potential to have one layer for both wms and wmts, considering the main difference between them are how tile URLs are structured (`sourceParamFormatter` should offer enough flexibility, I think) @ram-sharan25 can you try consolidating those layers and let us know if you face any challenge?

ram-sharan25:
I’ve added the individual components in this commit. Although they share structural similarities, their tile parameters and source formatting logic differ significantly, which makes maintaining them as separate components a cleaner and more maintainable solution.

In parallel, I’ve also implemented a common base module with lightweight wrappers for each layer type, available in a separate branch for review:
🔗 [Combined version](https://github.com/NASA-IMPACT/veda-ui/commit/c94ab77b37bfbde77501f38c207e880c5874141c)

@hanbyul-here  @slesaad Please let me know if the modular approach seems preferable, or if you would recommend proceeding with the original standalone version.

app/scripts/components/common/map/style-generators/hooks.ts
```diff
@@ -206,6 +225,59 @@ export function useWMS({ id, stacCol, stacApiEndpointToUse, onStatusChange }) {
...
+        });
+        const bounds = data.extent.spatial.bbox[0];
+        setBounds(bounds);
+        const wmtsLink = data.links.find((l) => l.rel === 'wmts');
+
+        if (!wmtsLink || !wmtsLink.href) {
+          throw new Error('WMTS link is missing or malformed');
+        }
+        const primaryHref = wmtsLink.href;
+        const alternateServers: string[] = wmtsLink['href:servers'] ?? [];
```
hanbyul-here, line 253: If I am reading the code right,  `useWMS` and `useWMTS` hooks appear to differ only in how they handle `alternateServers`. This makes me think we could consolidate them into a single hook.
Here's the expected behavior:

WMS layers: Since wmsLink lacks href:servers, it would return an empty array [], causing the hook to return only [primaryHref] (We will need to adjust wms-timeseries to accept wmsurl as string[])
WMTS layers: These would return both [primaryHref, alternateServers] since the alternate servers exist

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
```diff
@@ -139,7 +140,7 @@ export function RasterPaintLayer(props: RasterPaintLayerProps) {
             });
         } catch (e) {
           // eslint-disable-next-line no-console
-          console.error(e);
+          console.error('hello', e);
```
hanbyul-here, line 143:
```suggestion
          console.error(e);
```

app/scripts/components/common/map/style-generators/wms-timeseries.tsx
```diff
@@ -64,28 +64,27 @@ export function WMSTimeseries(props: MapLayerWMSProps) {
...
-      }}
+      tileApiEndpoint={wmsUrl as string}
+      tileParams={tileParams}
+      generatorPrefix='wms'
+      onStatusChange={changeStatus}
+      metadataFormatter={(_, tileParamsAsString) => ({
+        wmsTileUrl: `${wmsUrl}?${tileParamsAsString}`
+      })}
+      sourceParamFormatter={(url) => ({
+        tiles: [url],
```
hanbyul-here, line 85: url seems to be already string[]. Does it need to be wrapped as an array again?

app/scripts/components/common/map/style-generators/wmts-timeseries.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <RasterPaintLayer
+      {...props}
+      id={id}
+      tileApiEndpoint={primaryUrl}
+      tileParams={tileParams}
+      generatorPrefix='wmts'
+      onStatusChange={changeStatus}
+      metadataFormatter={(_, tileParamsAsString) => ({
+        wmtsTileUrl: `${wmtsUrl}?${tileParamsAsString}`
```
hanbyul-here, line 64: wmtsUrl is string[], should it be primaryUrl? fyi, this is called to load the tileurl for GIS. 

Screenshot 2025-06-09 at 12 36 13 PM

app/scripts/components/common/map/style-generators/hooks.ts
```diff
@@ -159,10 +163,15 @@ export function useCMR({
...
-export function useWMS({ id, stacCol, stacApiEndpointToUse, onStatusChange }) {
+export function useWebMapService({
+  id,
+  stacCol,
+  stacApiEndpointToUse,
+  onStatusChange
+}) {
   const defaultBounds: [number, number, number, number] = [-180, -90, 180, 90];
-  const [wmsUrl, setWmsUrl] = useState('');
+  const [url, setUrl] = useState<string[]>();
```
hanbyul-here, line 173: I am seeing the wrong use of this value as `string`, not `string[]` as it is intended. I wonder if it would be helpful to rename this variable as plural? (`urls`)

hanbyul-here (changes requested):
Some of the string[] and string values were used inappropriately.

app/scripts/components/common/map/style-generators/wms-timeseries.tsx
ram-sharan25: this url is the tileurl from the raster-painter-layer which is transformed and passed here as string.

hanbyul-here: Ah you are right. It can be named a bit differently, so it isn't confusing!

ram-sharan25:
@hanbyul-here  I changed the name of the variable. Let me know if there is anything I missed there .

hanbyul-here:
@ram-sharan25 looks good to me!

Issue #1233, opened 2024-10-31
earth.gov standard US gov banner


aboydnw:
### Context

[Banner component](https://designsystem.digital.gov/components/banner/) matching implementation in current earth.gov and GHGC. Seen at the top of the designs [here](https://www.figma.com/proto/2Vd0SDhFOTBJaYioJPr0Ei/Earth.gov?page-id=2321%3A441&node-id=2432-25820&viewport=834%2C-707%2C0.14&t=SkFmR5uIi8MuA1DP-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2432%3A25820&show-proto-sidebar=1). Should match all standard text and functionality from USWDS.

### Acceptance Criteria
- [ ] Banner component is on NextJS earth.gov instance, matching USWDS

sandrahoang686:
@aboydnw do we want to close this out? I think this existed to see how USWDS could be used in the template instance but we instead chose to implement pageheader to start.

aboydnw:
Is the banner component and site alert component something that we can easily include in any instances? for example if we make a NextJS instance for earth.gov, can we already use that component or is it important to somehow include it in the template instance?

sandrahoang686:
@aboydnw Looking at the veda-ui code, it looks like the banner could be exposed and used mostly without having to touch to much logic. Might need to expose some props to make it more flexible though. So maybe we keep this ticket for whenever we have an modern instance with the need? or can just close it and open it back up when there are more requirements.

aboydnw:
Okay, I'll just shift it to be for the earth.gov redesign, because we'll need it then and I need to start creating a backlog for that effort :) thanks @sandrahoang686

aboydnw:
related ticket to be aware of: https://github.com/NASA-IMPACT/veda-ui/issues/1574

Issue #1573, opened 2025-04-01
Fix USWDS spacing overrides


dzole0311:
## Context

In few places, we override USWDS default spacing (margins/paddings), which can unintentionally break layout alignment in the USWDS grid system. I observed these after bumping the veda-ui v6.1.1 and v6.2.0 in EIC and GHGC.

So far, I've noticed:

1. [usa-card](https://github.com/NASA-IMPACT/veda-ui/blob/3dd419a0b1e62958b6dcd4279f249ff64a9e39b1/app/scripts/components/common/card/uswds-cards/styles.scss#L4) has margin: 0 set, which removes spacing between cards inside a USWDS grid. We should consider removing this because it breaks the layout alignment when the card is used in a grid (e.g. no Y-space between the cards)
2. Once removed, the cards layout in the Data Catalog might need minor adjustment (which I think is why the override above was added)
3. [usa-accordion](https://github.com/NASA-IMPACT/veda-ui/blob/3dd419a0b1e62958b6dcd4279f249ff64a9e39b1/app/scripts/components/common/form/checkable-filter/checkable-filter.scss#L3) should be scoped more narrowly (e.g. to use a more specific selector). Right now it unintentionally affects elements like the dropdown in the .gov banner due to shared classnames

I’ve patched these manually in the EIC and GHGC instances but it would be better to fix them in veda-ui.

## Acceptance Criteria

- Card spacing works correctly in USWDS grid
- No visual regressions in Data Catalog layout
- Accordion styles are scoped to avoid affecting unrelated elements

## Definition of Done

- [ ] Unit tests
- [ ] (Where applicable) Automated playwright tests to cover primary use case(s)
- [ ] Merged to develop in relevant repo
- [ ] (?) Tested in instance context (GHG, earth.gov, template instance, etc)
- [ ] Relevant details of work are shared in a team forum (PR, ticket, slack, etc) to support knowledge sharing

PR #1679, opened 2025-05-15
fix: Namespace custom style

hanbyul-here:
**Related Ticket:** part of #1233 

Overwritten style of `.usa-accordion__button` is impacting other components of VEDA (ex. Banner). 
This should not have any visible impact on legacy instances. (Since legacy instances compile css per pages anyway.)

### DIFF
--- a/app/scripts/components/common/banner/banner.scss
+++ b/app/scripts/components/common/banner/banner.scss
@@ -1,10 +1,6 @@
 @use '$styles/veda-ui-theme-vars.scss' as themeVars;
 
 .veda-banner {
-  .usa-banner__button:after {
-    top: 3px;
-  }
-
   .banner--left-aligned {
     & .usa-banner__inner,
     & .usa-banner__content {
--- a/app/scripts/components/common/banner/banner.scss
+++ b/app/scripts/components/common/banner/banner.scss
@@ -1,12 +1,14 @@
 @use '$styles/veda-ui-theme-vars.scss' as themeVars;
 
-.usa-banner__button:after {
-  top: 3px;
-}
+.veda-banner {
+  .usa-banner__button:after {
+    top: 3px;
+  }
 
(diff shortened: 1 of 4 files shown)


vgeorge:
@hanbyul-here this looks like a very simple change, but I'm not sure if it might affect other unrelated parts of the code. Do you have any recommendations on how to best inspect or test it? Alternatively, do you think it would make sense to add a small unit test here?

dzole0311:
I think this also might be related to issue number 3) here: https://github.com/NASA-IMPACT/veda-ui/issues/1573

vgeorge:
@hanbyul-here I checked out https://github.com/NASA-IMPACT/next-earth-gov/pull/36 and the banner looked ok on that branch. I’m still a bit confused about how this PR connects to that one, and what exactly I should be looking at to review here. Could you walk me through the steps?

hanbyul-here:
@vgeorge 

- This PR is not connected to #1233  anymore. I started the PR when next-earth-gov was still using veda-ui, meaning it still imports css from veda-ui (or I based my branch on the old branch that still used veda-ui). next-earth-gov does not use veda-ui at all at this point so this PR is not connected to #1233 at all at this point. 

- But this style will still make trouble if any other instance is trying to use veda-ui with USWDS. (ex. Using E&A from veda-ui but using uswds banner straight out from USWDS - I set up an example in our next repo: https://github.com/NASA-IMPACT/next-veda-ui/pull/71) 

- This also resolves issue 3. listed in this ticket that @dzole0311 mentioned https://github.com/NASA-IMPACT/veda-ui/issues/1573

- This should not have any impact on legacy system. I opened a PR with the change in this PR on ghg to generate a preview https://github.com/US-GHG-Center/veda-config-ghg/pull/777 (It is expected for images in the banner to be broken in the preview.)

Issue #1674, opened 2025-05-13
Hide pageHero for custom routes 


dzole0311:
## Context

We're setting up a custom route in veda-config called `tools/wildfire-visualization`. The overrides and changes work fine, but the `PageHero` is shown at all times when creating a custom page in veda-config.

## Acceptance Criteria

- [ ] Pass a optional prop to hide the PageHero when creating custom pages

## Definition of Done

- [ ] Unit tests
- [ ] (Where applicable) Automated playwright tests to cover primary use case(s)
- [ ] Merged to develop in relevant repo
- [ ] (?) Tested in instance context (GHG, earth.gov, template instance, etc)
- [ ] Relevant details of work are shared in a team forum (PR, ticket, slack, etc) to support knowledge sharing

hanbyul-here:
I looked at it, and  I don't think our component for custom pages has an option to opt out of hero: https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/user-pages/index.tsx we probably need to modify the component.

PR #1685, opened 2025-05-20
feat: Add props to control the visibility of the PageHero, NavWrapper and footer in the veda UI

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1674

### Description of Changes
Add support for hiding `NavWrapper`, `PageHero` and `PageFooter` via frontmatter flags (hideNav, hideHero, hideFooter) in custom user MDX pages.

This enables instance-level customization via veda-config.

Example usage in a custom page from the instance side: https://github.com/NASA-IMPACT/veda-config/pull/851/files#diff-c560b51533d557b1479e089ad4132b940cf5febeebc3bc24e39103efb5a7dd93R2

### Validation / Testing

I've set up a custom page in veda-config to quickly test the new props:

1. Visit https://deploy-preview-851--visex.netlify.app/
2. Click on Exploration Tools in the nav bar -> Wildfire visualization
3. You should only see `Hello from the custom page!` text shown on the page, w/o any page hero, navbar or footer components being shown

### DIFF
--- a/app/scripts/components/common/layout-root/index.tsx
+++ b/app/scripts/components/common/layout-root/index.tsx
@@ -10,7 +10,6 @@ import { useDeepCompareEffect } from 'use-deep-compare';
 import styled from 'styled-components';
 import { Outlet } from 'react-router';
 import { reveal } from '@devseed-ui/animation';
-import { NavLink } from 'react-router-dom';
 import {
   getBannerFromVedaConfig,
   getCookieConsentFromVedaConfig,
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/layout-root/index.tsx
```diff
@@ -99,15 +98,18 @@ function LayoutRoot(props: { children?: ReactNode }) {
       {siteAlertContent && (
```
hanbyul-here, line 98: we still want siteAlert?

hanbyul-here (approved):
🤔 It might be better to just have a new layout for this type of page. But at the same time, it probably is better not to spend much time on legacy structure.

aboydnw (approved):
The earthdata preview looks good to me!

app/scripts/components/common/layout-root/index.tsx
dzole0311: Probably not, but maybe we should do a wider clean up of USWDS exposed via veda-ui?

Issue #1688, opened 2025-05-27
[Bug] useSlidingStickyHeader throws error when the header is disabled


dzole0311:
### Context

With the changes introduced in [PR #1685](https://github.com/NASA-IMPACT/veda-ui/pull/1685), we should add a guard inside useSlidingStickyHeader to ensure it only runs when the header is visible.

Spotted this issue (the whole page breaks) when resizing the screen with the FireTool enabled in veda-config.

URL: 

Blocking a release? (y/n): 

Regression issue? (y/n):
[If yes, add Regression label]

Priority: 

### Steps to reproduce the issue

1. [Step 1 to reproduce]
2. [Step 2 to reproduce]

PR #1689, opened 2025-05-27
fix: Pass optional disabled prop to useSlidingStickyHeader

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1688

### Description of Changes
Add a `disabled` boolean prop to the `useSlidingStickyHeader` hook to allow consumers to explicitly opt out of sticky header behavior (eg when `hideNav` is true). The `LayoutRootContextProvider` now passes `hideNav` to the hook to prevent unnecessary effects / measurements when the nav is intentionally hidden (otherwise the whole page would break when resizing the screen for example)

I opted for this guard behavior inside the hook via a disabled flag because it seemed a simpler solution than conditionally using the hook. In the case of the latter, I think we'll violate the rules of hooks (that they shouldn't be called conditionally).

### Validation / Testing

You can confirm that no errors are thrown already using the veda-config preview deployment for the WildfireExplorer:

1. Go to https://deploy-preview-851--visex.netlify.app/
2. Click on "Exploration Tools" -> "Wildfire Visualization"
3. The page should render properly when you try to resize the screen

### DIFF
--- a/app/scripts/components/common/layout-root/context.tsx
+++ b/app/scripts/components/common/layout-root/context.tsx
@@ -3,7 +3,7 @@ import React, {
   useState,
   Dispatch,
   SetStateAction,
-  ReactNode,
+  ReactNode
 } from 'react';
 import { useLocation } from 'react-router';
 import { useSlidingStickyHeader } from '$utils/use-sliding-sticky-header';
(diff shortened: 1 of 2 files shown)


hanbyul-here (approved):
hoefully we can deprecate the og nav soon so we don't have to do this anymore 🙏

PR #1746, opened 2025-06-30
Changes to sel parsing logic for sourceParams

sandrahoang686:
**Related Ticket:* https://github.com/NASA-IMPACT/veda-odd/issues/188

@abarciauskas-bgse I didn't want to hold this up any further so I just went in to make the changes myself. I updated the two methods you created and left comments on what i changed in the files here.

### DIFF
--- a/app/scripts/components/common/map/style-generators/cmr-timeseries.tsx
+++ b/app/scripts/components/common/map/style-generators/cmr-timeseries.tsx
@@ -21,31 +21,23 @@ export function CMRTimeseries(props: BaseTimeseriesProps) {
   const start_datetime = userTzDate2utcString(startOfDay(date));
   const end_datetime = userTzDate2utcString(endOfDay(date));
 
-  // Function to replace variables in curly braces with their values
-  const interpolateVariables = (
-    str: string,
-    variables: Record<string, string>
-  ) => {
-    return str.replace(/\{(\w+)\}/g, (match, variableName) => {
-      return variables[variableName] || match;
-    });
-  };
+  const processSourceParams = () => {
+    const parseSelAttribute = () =>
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/map/style-generators/cmr-timeseries.tsx
```diff
@@ -21,31 +21,23 @@ export function CMRTimeseries(props: BaseTimeseriesProps) {
...
-      return variables[variableName] || match;
-    });
-  };
+  const processSourceParams = () => {
+    const parseSelAttribute = () =>
+      sourceParams?.sel.map((sel) => {
+        if (sel.includes('{start_datetime}')) {
+          return sel.replace('{start_datetime}', start_datetime);
+        } else return sel;
+      }); // "start_datetime" is currently the only variable to be interpolated
```
sandrahoang686, line 30: I removed `interpolateVariables` because it looked like the function was created to scale but I dont think we need to scale in this case because we are looking for specific vars so I thought we could just be a little more straightforward here.

app/scripts/components/common/map/style-generators/utils.ts
```diff
@@ -2,15 +2,24 @@ import qs from 'qs';
...
+  } = params;
 
   const repeatParams: Record<string, string[] | undefined> = {
     bands,
     assets,
     bidx,
     asset_bidx,
-    sel
+    sel,
+    sel_method
```
sandrahoang686, line 22: I've added `sel_method` because it looks like it can still be passed in from an mdx file? I know you mentioned something about it defaulting to `nearest` but in case the user passes in something else, we'd probably need to pass it in?

PR #1757, opened 2025-07-03
Develop custom layer example

hanbyul-here:
develop custom layer example a bit more to be realistic.


app/scripts/components/common/map/hooks/use-layer-interaction.ts
```diff
@@ -5,14 +5,16 @@ import useMaps from './use-maps';
 interface LayerInteractionHookOptions {
   layerId: string;
   onClick?: (features: Feature<any>[]) => void;
+  enabled?: boolean;
```
hanbyul-here, line 8: I had to add this flag because there was no guarantee that child element will be rendered when the map instance is ready. (And the map instance should be ready before attaching the event.)

PR #1761, opened 2025-07-10
chore: update caniuse

vgeorge:
Updates the lockfile to use the latest versions of `caniuse`. This is the result of running `npx update-browserslist-db@latest`. No regressions expected and caniuse warnings should not be displayed. I updated the lockfile manually to use verdaccio.

Ready for review.


hanbyul-here (approved):
thanks! 

Not related to this PR - but this pr reminded me that we are resolving dependencies from Verdaccio 🤔 I don't see any reason to move away from it now. But I also don't see the reason why we would need it when we are not publishing anymore - if you have better grasps about Verdaccio, I would appreciate your insight!

vgeorge:
@hanbyul-here thanks for the review. It seems that we can remove Verdaccio references, but I think @dzole0311 knows more about it as he worked on it in the past. I opened  #1763 to track it.

PR #1772, opened 2025-07-21
Fix render extension logic

hanbyul-here:
### Description of Changes
I believe this fix was first introduced as a temporary solution for dashboard not being able to parse array value as query parameters for titiler. [We have solved that issue on the dashboard side then](https://github.com/NASA-IMPACT/veda-ui/blob/50328c357ad988699aa991b0f07866d2cc806818/app/scripts/components/common/map/style-generators/utils.ts#L2).  

I also deprecated the part of the code that hits /mosaic endpoint from raster-timeseries data since that endpoint is gone.

### Validation / Testing
Try the dataset without any sourceparams defined. I got rid of it from nightlights dataset for testing.


aboydnw (commented):
I did a smoke test and looks good to me. But I'll let an engineer give the green check mark

sandrahoang686 (approved):

Issue #1777, opened 2025-07-24
Add ability to support toggling between multiple layers and date selection in BlockMap component


sandrahoang686:
## Context

For the blockmap component we want to be able to add a configuration option to toggle between multiple layers and filter dates. Currently we cannot do this and the BlockMap only supports single layers (?). 

Might have to featureflag this. 

Look at designs [here](https://www.figma.com/design/PG7AhRloW6qJd9U00VdQ77/VEDA-Template-Instance?node-id=665-193&p=f&t=nO0s9cmHUi3DVuB7-0).

this is separate from the widget as the BlockMap component should already have this ability before being added inside the widget component.

_(from parent story) - The proposal here is to add this as a configuration option for our map widget, enabling instances to use the map widget as the sole visualization for a dataset and not having to rely on E&A, which is really more specialized for Time Series Analysis._

## Acceptance Criteria

* We can toggle between multiple layers in the BlockMap Component
* We can change date

## Definition of Done

- [ ] Unit tests
- [ ] (Where applicable) Automated playwright tests to cover primary use case(s)
- [ ] Merged to develop in relevant repo
- [ ] (?) Tested in instance context (GHG, earth.gov, template instance, etc)
- [ ] Relevant details of work are shared in a team forum (PR, ticket, slack, etc) to support knowledge sharing

PR #1781, opened 2025-07-28
feat: Draft support for multi layer/date selectors in the the MapBlock

dzole0311:
**Related Ticket:** https://github.com/NASA-IMPACT/veda-ui/issues/1777

### Description of Changes

This is a rough draft w/o styles to gather feedback on the approach of adding support for multi-layer selector and date selection for the MapBlock. The PR:

1. Adds layer picker for datasets with multiple layers that are fetched from STAC (disabled when compareDateTime is set)
2. Adds date selectior that dynamically adjusts to the `timeDensity` from STAC metadata
3. Disables new layer/date selectors when the MapBlock is in compare mode

See the first two map blocks here: https://deploy-preview-1781--veda-ui.netlify.app/stories/air-quality-and-covid-19

Notes / assumptions:

1. Only layers from the same dataset are shown in the layer selector
2. When in comparison mode, we don't need to show the new layer/date selectors
3. `timeDensity` of less than a day (e.g. '12h') is not currently supported by our calendar. To support half-day granularity, we’ll need to add an additional input below the calendar

Open questions:

1. Should we add an optional `excludeLayers` prop to allow content authors to hide specific layers?

### Validation / Testing

See the first two map blocks here: https://deploy-preview-1781--veda-ui.netlify.app/stories/air-quality-and-covid-19

### DIFF
--- a/app/scripts/components/common/blocks/block-map.tsx
+++ b/app/scripts/components/common/blocks/block-map.tsx
@@ -1,8 +1,12 @@
+/* eslint-disable */
+// @ts-nocheck
 import React, { useMemo, useState, useEffect } from 'react';
 import type { PropsWithChildren } from 'react';
 import styled from 'styled-components';
 import { MapboxOptions } from 'mapbox-gl';
 import * as dateFns from 'date-fns';
+import { DateInput, Icon } from '@trussworks/react-uswds';
+import Calendar from 'react-calendar';
 import {
   convertProjectionToMapbox,
   projectionDefault,
@@ -37,6 +41,8 @@ import { useReconcileWithStacMetadata } from '$components/exploration/hooks/use-
 import { ProjectionOptions, VedaData, DatasetData } from '$types/veda';
 import { useVedaUI } from '$context/veda-ui-provider';
 
+import 'react-calendar/dist/Calendar.css';
+
 export const mapHeight = '32rem';
 const Carto = styled.div`
   position: relative;
@@ -44,7 +50,6 @@ const Carto = styled.div`
   height: ${mapHeight};
 `;
 
-// This global variable is used to give unique ID to mapbox container
 let mapInstanceId = 0;
 
 const lngValidator = validateRangeNum(-180, 180);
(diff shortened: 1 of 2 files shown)


app/scripts/components/common/blocks/block-map.tsx
```diff
@@ -157,73 +161,70 @@ export default function MapBlock(props: PropsWithChildren<MapBlockProps>) {
...
+  }, [layerId]);
+
+  useEffect(() => {
+    setSelectedDate(dateTime);
+  }, [dateTime]);
+
+  const errors = validateBlockProps(props);
+  if (errors.length) throw new HintedError('Malformed Map Block', errors);
 
   const datasetLayers = getDatasetLayers(datasets);
```
dzole0311, line 188: Ideally, we won't need to pass all `datasets` to the MapBlock, but that's a bigger lift and will bloat the scope here.

hanbyul-here:
@aboydnw  - Do you expect these maps to be used inside of the story? Based on the design and the usage of mapbock so far, like it is mainly for dataset page.  If so, I wonder it would be better to have a separate component that uses MapBlock instead of adding more flags to MapBlock.

app/scripts/components/common/blocks/block-map.tsx
hanbyul-here: I agree that this can be a separate problem, but I wonder if it is a lighter lift than it seems. We already have a 'wrapper' element for Mapblock to pass down datasets (both for legacy: https://github.com/NASA-IMPACT/veda-ui/blob/859151403f689fb990155869fcdef84a2d388e11/app/scripts/components/common/blocks/lazy-components.js#L45 and the new structure: https://github.com/NASA-IMPACT/next-veda-ui/blob/5bda58c419d3be40c9ae8cd220d2a236eee0d9ff/app/components/mdx-components/block.tsx#L8

I can easily see delegating the work of selecting the layers we need out of the datasets and pass it to mapblock, so we can at least avoid passing whole datasets to mapblock. 🤔 how do you think?

hanbyul-here:
🤔 I am a bit concerned about how open-ended interaction can be included in a story without adding a complexity on user side. Aside from this concern, I still think it is better to make this component to consume BlockMap instead of making it as a part of BlockMap, so we don't bloat BlockMap.

dzole0311:
Closing in favor of the newer PR (all the discussion points are addressed there): https://github.com/NASA-IMPACT/veda-ui/pull/1808

Issue #1794, opened 2025-08-01
E&A Breakdown: Data Layer Information Card


hanbyul-here:
## Context

VEDA UI is going to export components necessary for Water Insight tool: https://github.com/NASA-IMPACT/veda-ui/issues/1792

As Water Insight instance requires a lot of features that DataLayerInformationCard offers on E&A (such as changing opacity, reordering of the layers), we will export DataLayerInformationCard from veda-ui. 

<img width="279" height="348" alt="Image" src="https://github.com/user-attachments/assets/e521b693-0e13-46dc-a277-c76a8436a008" />

## Acceptance Criteria

- [ ] Given the water insight instance's requirements, export DatasetLayerInformationCard that can edit the look of the data layers on the map.
- [ ] Given Map components https://github.com/NASA-IMPACT/veda-ui/issues/1793 is ready, offer a pattern, example through Sandbox how to put DatasetLayerInformationCard and Map together. 

## Definition of Done

- [ ] Outcome components are released through veda-ui
- [ ] Tested through Next instance

Issue #1796, opened 2025-08-01
E&A Breakdown: Timeseries Analysis


hanbyul-here:
## Context

VEDA UI is going to export components necessary for Water Insight tool: https://github.com/NASA-IMPACT/veda-ui/issues/1792

Water Insight instance is going to offer timeseries chart for selected layers (In the process of clarifying the requirement, whether to plot multiple charts at the same time or not). Based on the requirement and original sketch, Timeseries Chart from E&A might be doing way more than what Water Insight instance needs. This ticket requires doing a spike of evaluating if it is worth exporting Timeseries chart component from E&A page. If not, what we should offer from VEDA-UI ex. we can still offer hook to fetch timeseries analysis  (https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/hooks/use-analysis-data-request.ts) 

## Dependency 

We will use titiler-multidim for Water Insight tool. ODD team is converting the data into Virtual Zarr, and implementing timeseries endpoint for titiler-multidim. (expected to be done in ~2 weeks - 8/15) We should be able to get the mock response data from ODD team. The response data will be similiar to what we get from titiler-cmr.

## Acceptance Criteria

- [ ] Offer a solution for timeseries analysis, whether it is a visual component, hook or custom solution. 
- [ ] Offer an example of how to use the solution from the first step with other components such as aoi control and date inputs.

## Definition of Done

- [ ] Solution for timeseries available for Water Insight instance.

PR #1801, opened 2025-08-04
feat: Export LayerCard from E&A

hanbyul-here:
**Related Ticket:** #1794 

### Description of Changes
I made a LayerCardWithSync (If there is a better name for this component.. I am all ears!) component that hydrates LayerCard component with the values derived from Atom, so when it is exposed to users, users don't have to worry about all the functions to synchronize the state with the map. I made an example of using this component in sandbox.

I am kind of figuring out to-dos as I go, but opening this and tagging reviewers early in case this helps. I will mark the PR to be ready to be reviewed once it is in a shape that can be merged. 

1. [x]  Configurable Colormap seems to be not working on sandbox page 🤔 I am not sure why. (Blocker) : this was a mock data problem

2. To make DataLayerCard usable without atom value hydration, all these actions on tooltip also need to be reconsidered/exposed to the user. (Not a blocker since water insight tool won't need this part customized, but then we should only export LayerCardWithSync from veda-ui)

Screenshot 2025-08-04 at 4 10 50 PM

3. I noticed that URL gets stripped off when navigating back and forth from sandbox page 🤔 I would like to separate this bug but should handle it at some point.  (Probably a blocker but will handle separately - we had this problem before, it is because urlatom taking all the previous items : https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/utils/params-location-atom/url.ts#L85) 

### Validation / Testing

Check the components in action: https://deploy-preview-1801--veda-ui.netlify.app/sandbox/exploration-layer-card
Also check Exploration page is not impacted. https://deploy-preview-1801--veda-ui.netlify.app/exploration

### DIFF
--- a/app/scripts/components/sandbox/exploration/layer-card.js
+++ b/app/scripts/components/sandbox/exploration/layer-card.js
@@ -88,53 +88,96 @@ function DataLayerCardSet({ timelineDatasets, onChange }) {
   );
 }
 
-function SandboxExplorationMapWithLayerCard() {
-  const setExternalDatasets = useSetAtom(externalDatasetsAtom);
+class ReactPlugin extends React.Component {
+  // react specific things needed for plugin
+  getter(key) {
+    return this.state[key];
+  }
 
-  setExternalDatasets(mockRawData);
-  const [timelineDatasets, setTimelineDatasets] = useTimelineDatasetAtom();
-
-  useEffect(() => {
-    // IF no data came through URL
-    if (timelineDatasets.length == 0) {
-      setTimelineDatasets([...mockDatasets]);
-    }
-  }, [setTimelineDatasets, timelineDatasets.length]);
-
-  // eslint-disable-next-line no-console
-  const handleChange = useCallback((e) => console.log(e), []);
-  return (
-    <PageMainContent>
-      <HugResetter>
-        <GridContainer id='grid-contianer'>
-          <h2 className='margin-top-5 margin-bottom-2'>
-            Layer card in sync with Map
-          </h2>
-          <Grid row gap={3}>
-            <Grid style={{ height: '300px' }} col={6}>
-              <ExplorationMap
-                datasets={timelineDatasets}
-                setDatasets={setTimelineDatasets}
-                selectedDay={mockSelectedDay}
-              />
-            </Grid>
-            <Grid col={6}>
-              <DataLayerCardSet
-                timelineDatasets={timelineDatasets}
-                onChange={handleChange}
-              />
-            </Grid>
-          </Grid>
-          <h2 className='margin-top-5 margin-bottom-2'>
-            Custom UI using hooks
-          </h2>
-          {timelineDatasets.length > 0 && (
-            <ExampleComponent dataset={timelineDatasets[0]} />
-          )}
-        </GridContainer>
-      </HugResetter>
-    </PageMainContent>
-  );
+  setter(key, value) {
+    this.setState({
+      key: value
+    });
+  }
 }
+// Higher-Order Component that handles hooks
+function withHooks(WrappedComponent) {
+  return function HOCComponent(props) {
+    const setExternalDatasets = useSetAtom(externalDatasetsAtom);
+    const [timelineDatasets, setTimelineDatasets] = useTimelineDatasetAtom();
+
+    useEffect(() => {
+      // Initialize external datasets
+      setExternalDatasets(mockRawData);
+    }, [setExternalDatasets]);
+
+    useEffect(() => {
+      // IF no data came through URL
+      if (timelineDatasets.length === 0) {
+        setTimelineDatasets([...mockDatasets]);
+      }
+    }, [setTimelineDatasets, timelineDatasets.length]);
+
+    const handleChange = useCallback((e) => {
+      // eslint-disable-next-line no-console
+      console.log(e);
+    }, []);
+
+    // Pass hook values as props to the class component
+    return (
+      <WrappedComponent
+        {...props}
+        timelineDatasets={timelineDatasets}
+        setTimelineDatasets={setTimelineDatasets}
+        handleChange={handleChange}
+      />
+    );
+  };
+}
+
+// Class Component (pure presentation)
+class SandboxExplorationMapWithLayerCardClass extends ReactPlugin {
+  render() {
+    const { timelineDatasets, setTimelineDatasets, handleChange } = this.props;
+
+    return (
+      <PageMainContent>
+        <HugResetter>
+          <GridContainer id='grid-contianer'>
+            <h2 className='margin-top-5 margin-bottom-2'>
+              Layer card in sync with Map
+            </h2>
+            <Grid row gap={3}>
+              <Grid style={{ height: '300px' }} col={6}>
+                <ExplorationMap
+                  datasets={timelineDatasets}
+                  setDatasets={setTimelineDatasets}
+                  selectedDay={mockSelectedDay}
+                />
+              </Grid>
+              <Grid col={6}>
+                <DataLayerCardSet
+                  timelineDatasets={timelineDatasets}
+                  onChange={handleChange}
+                />
+              </Grid>
+            </Grid>
+            <h2 className='margin-top-5 margin-bottom-2'>
+              Custom UI using hooks
+            </h2>
+            {timelineDatasets.length > 0 && (
+              <ExampleComponent dataset={timelineDatasets[0]} />
+            )}
+          </GridContainer>
+        </HugResetter>
+      </PageMainContent>
+    );
+  }
+}
+
+// Final component with HOC applied
+const SandboxExplorationMapWithLayerCard = withHooks(
+  SandboxExplorationMapWithLayerCardClass
+);
 
 export default SandboxExplorationMapWithLayerCard;


app/scripts/components/common/map/layer-legend.tsx
```diff
@@ -140,65 +145,64 @@ const LayerLegendSelf = styled.div`
   }
 `;
 
-
-
 const LegendList = styled.dl<LegendListProps>`
+  width: 100%;
```
hanbyul-here, line 149: Only change I made is this line, all other changes on this file are lint.

hanbyul-here:
This PR is ready for review @vgeorge @AliceR  (also tagging @sandrahoang686   in case this pr helps the ticket https://github.com/NASA-IMPACT/veda-ui/issues/1796)

app/scripts/components/exploration/components/datasets/data-layer-card.tsx
```diff
@@ -123,16 +116,18 @@ const LegendColorMapTrigger = styled.div`
...
     setVisible,
     colorMap,
     setColorMap,
     colorMapScale,
     setColorMapScale,
-    datasetLegend,
     onClickLayerInfo
   } = props;
+
+  const datasetAtom = useTimelineDatasetAtom(dataset.data.id);
```
AliceR, line 128: We would want to get rid of the jotai atom in this component. Should we pass the `dataset.data.id` directly to `LayerMenuOptions` and delegate the fetching?

app/scripts/components/exploration/components/datasets/data-layer-card.tsx
```diff
@@ -317,3 +322,41 @@ export default function DataLayerCard(props: CardProps) {
     </>
   );
 }
+
+export function DataLayerCardWithSync(props: {
```
AliceR, line 326: Since this is a wrapper specific to the exploration page, should we move it there and keep the data-layer-card component clean?

AliceR (changes requested):
This is going in a good direction @hanbyul-here !
I pair-reviewed with @vgeorge and we were not able to run the changes, we suspect the missing `index.js` file is the reason?
If you could fix the issue, we can take it for another review. Left a few comments nonetheless.

app/scripts/components/exploration/components/datasets/data-layer-card.tsx
```diff
@@ -317,3 +322,41 @@ export default function DataLayerCard(props: CardProps) {
...
+}) {
+  const { dataset, setLayerInfo } = props;
+
+  const datasetAtom = useTimelineDatasetAtom(dataset.data.id);
+
+  const [isVisible, setVisible] = useTimelineDatasetVisibility(datasetAtom);
+  const [colorMap, setColorMap] = useTimelineDatasetColormap(datasetAtom);
+  const [colorMapScale, setColorMapScale] =
+    useTimelineDatasetColormapScale(datasetAtom);
+  // const [, setModalLayerInfo] = useState<LayerInfoModalData>();
```
sandrahoang686, line 338: commented out code

app/scripts/components/sandbox/exploration/layer-card.js
```diff
@@ -0,0 +1,123 @@
...
+import { ExplorationMap } from '$components/exploration/components/map';
+export const HugResetter = styled.div`
+  /* To escape from HUG grid */
+`;
+
+const mockSelectedDay = new Date('2017-12-01T00:00:00.000Z');
+
+/* @TODO: This component is not working, leaving to investigate more later */
+// eslint-disable-next-line no-unused-vars
+function ExampleComponent(props) {
```
sandrahoang686, line 22: TODO can probably be removed now? 

And the ExampleComponet is just the DataLayerCard that is meant to be offered by itself right? I dont think it can be offered singulary right now because it still depends on the url from the atom state. While debugging I noticed that `ExampleCard` only rendered correctly when the url was finished generating => `/?datasets=%5B%7B"id"%3A"casa-gfed-co2-flux-hr"%2C"settings"%3A%7B"isVisible"%3Atrue%2C"opacity"%3A100%2C"analysisMetrics"%3A%5B%7B"id"%3A"mean"%2C"label"%3A"Average"%2C"chartLabel"%3A"Average"%2C"themeColor"%3A"infographicB"%7D%2C%7B"id"%3A"std"%2C"label"%3A"St+Deviation"%2C"chartLabel"%3A"St+Deviation"%2C"themeColor"%3A"infographicD"%7D%5D%2C"colorMap"%3A"blues"%2C"scale"%3A%7B"min"%3A0%2C"max"%3A0.3%7D%7D%7D%2C%7B"id"%3A"casa-gfed-co2-flux"%2C"settings"%3A%7B"isVisible"%3Atrue%2C"opacity"%3A100%2C"analysisMetrics"%3A%5B%7B"id"%3A"mean"%2C"label"%3A"Average"%2C"chartLabel"%3A"Average"%2C"themeColor"%3A"infographicB"%7D%2C%7B"id"%3A"std"%2C"label"%3A"St+Deviation"%2C"chartLabel"%3A"St+Deviation"%2C"themeColor"%3A"infographicD"%7D%5D%7D%7D%5D`

app/scripts/components/exploration/components/datasets/data-layer-card.tsx
sandrahoang686, line 338: commented out code

app/scripts/components/sandbox/exploration/layer-card.js
sandrahoang686, line 22: TODO can probably be removed now? 

And the ExampleComponet is just the DataLayerCard that is meant to be offered by itself right? I dont think it can be offered singulary right now because it still depends on the url from the atom state. 

While debugging I noticed that `ExampleCard` only rendered correctly when the url was finished generating => 
```
/sandbox/exploration-layer-card?datasets=%5B%7B"id"%3A"casa-gfed-co2-flux-hr"%2C"settings"%3A%7B"isVisible"%3Atrue%2C"opacity"%3A100%2C"analysisMetrics"%3A%5B%7B"id"%3A"mean"%2C"label"%3A"Average"%2C"chartLabel"%3A"Average"%2C"themeColor"%3A"infographicB"%7D%2C%7B"id"%3A"std"%2C"label"%3A"St+Deviation"%2C"chartLabel"%3A"St+Deviation"%2C"themeColor"%3A"infographicD"%7D%5D%2C"colorMap"%3A"blues"%2C"scale"%3A%7B"min"%3A0%2C"max"%3A0.3%7D%7D%7D%2C%7B"id"%3A"casa-gfed-co2-flux"%2C"settings"%3A%7B"isVisible"%3Atrue%2C"opacity"%3A100%2C"analysisMetrics"%3A%5B%7B"id"%3A"mean"%2C"label"%3A"Average"%2C"chartLabel"%3A"Average"%2C"themeColor"%3A"infographicB"%7D%2C%7B"id"%3A"std"%2C"label"%3A"St+Deviation"%2C"chartLabel"%3A"St+Deviation"%2C"themeColor"%3A"infographicD"%7D%5D%7D%7D%5D
```

app/scripts/components/sandbox/exploration/layer-card.js
```diff
@@ -0,0 +1,123 @@
+import React, { useState, useEffect, useRef } from 'react';
```
sandrahoang686, line 1: Naming is a little odd. Since the file is `layer-card.js` but its exporting by default the Map too. Default export name is `SandboxExplorationMap` but file is just the card. So maybe we can just move `SandboxExplorationMap` to index file and keep just the card logic here.

sandrahoang686 (commented):
Looks good, I think I understand what was trying to be achieved here. The timeline dataset atoms is whats tying everything together state wise. So to use coupled, it would rely on the using the atoms but we want to be able to support each individually not tied together (?). This seems little complicated right now because the `DataLayerCard` still looks like it relies on the URL to resolve. But if a user wanted to say Bring their own card (BYOC) they just need to reference the atoms. Does this sound about right?

Once thing I did notice which is probably unwanted is that:
`/sandbox/exploration-layer-card` gets re-written over without the subpath so it looks like `/sandbox/?datasets=%5B%7B"id"%3A"casa-gfed-co2-flux-.....`

app/scripts/components/exploration/components/datasets/data-layer-card.tsx
hanbyul-here: I was thinking this can be used outside of E&A page, ex. water insight instance can use this component directly. 
Based on the convo we had today, I do think users (like dev users of veda-ui) can use the hooks (that basically this component is using: https://github.com/NASA-IMPACT/veda-ui/pull/1801/files/f6087fdbafdb62e326b10c1c44f4a609fbae1c99#diff-9ac16b2c33a88f83221cfa1b63bccc5eaa8ff468a605c0fe07d6f62f255c06a7R334)  and compose their own UIs too.  But I do think there is a value of offering this level of components to dev users, like offering already-composed components that use the hooks. I will talk more about this with water insight team today.

hanbyul-here:
I took a note on the url problem as a bullet point 3: https://github.com/NASA-IMPACT/veda-ui/pull/1801#issue-3290421768 will handle separately 🙇 @sandrahoang686

app/scripts/components/exploration/components/datasets/data-layer-card.tsx
vgeorge: @hanbyul-here PR https://github.com/NASA-IMPACT/veda-ui/pull/1809 should address this.

Issue #1793, opened 2025-08-01
E&A Breakdown: Map & Map Controls


hanbyul-here:
## Context

VEDA UI is going to export components necessary for Water Insight tool: https://github.com/NASA-IMPACT/veda-ui/issues/1792

ExplorationMap component is a composition of Map component along with other MapControl components. It is already usable as it is outside of E&A context. We have an example of this usecase in Sandbox: https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/sandbox/legacy/exploration-map/index.js#L113

But we have to reconsider what to expose from this map/what to include in this map based on Water Insight instance. (ex. Water Insight needs its own aoi selector, but ExplorationMap doesn't have a way to do opt out this control.)

To get more concrete idea, this experiment that I did might be helpful: https://github.com/NASA-IMPACT/veda-ui/pull/1786/files#diff-c3a80a3e585084dd4eb57e07fb259d5fda902ccb1e28b3b7f4da49834dbf33a3R1 

## Acceptance Criteria

- [ ] Given the water insight instance's requirements, expose a way of opting in/out from the map controls 
- [ ] Given the water insight instance's requirements, offer a way of communicating between the components (custom aoi selector -> Map AOI Layer)
- [ ] Given the water insight instance's requirements, we can offer a pattern that other devs can use EAMap through Sandbox

## Definition of Done

- [ ] Outcome components are released through veda-ui package
- [ ] tested through Next Instance

vgeorge:
Status update on the acceptance criteria:

- Since the Water instance can build its own map by replicating the Exploration page structure (as shown in [this PR](https://github.com/NASA-IMPACT/veda-ui/pull/1804)), it can use a custom AoiControl without needing changes to the existing one
- A custom AOI control can follow the same pattern as the current implementation, using `useAois` and `useMaps` hooks
- Given that the easiest way to build a custom EAMap is by reusing existing components, there doesn't seem to be a strong need to break things down further

@hanbyul-here please let me know if there is anything we need to address in this ticket or close it.

hanbyul-here:
I have a few follow up questions
- Can we offer an example of building custom AOI? It doesn't have to be well polished, or even hooked up with a typical aoi modifying tool. We can feed the hard-coded geojson as aoi but I do think it is valuable to offer a template.
- 🤔 If we are exporting DataLayerCardwithSync, should there be equivalent of components that react to this component? such as MapWithSync? 
[...]

PR #1804, opened 2025-08-06
feat: Set up sandbox page for custom map

vgeorge:
**Related Tickets:**

* #1793

This adds a sandbox page to showcase a custom map composed of individual map components, enabling customization of map controls and other behaviors.

@hanbyul-here It seems we can use this approach for the Water Insights tool. Is there a specific reason we need to use `ExplorationMap` directly? Our understanding (@AliceR and I) of #1793 was that we needed to decouple the AOI control from the map, but it looks like that's already the case with this approach.

### DIFF
--- a/app/scripts/components/sandbox/custom-map/index.tsx
+++ b/app/scripts/components/sandbox/custom-map/index.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import Map, { MapControls } from '$components/common/map';
+import { Basemap } from '$components/common/map/style-generators/basemap';
+
+import {
+  NavigationControl,
+  ScaleControl
+} from '$components/common/map/controls';
+import MapCoordsControl from '$components/common/map/controls/coords';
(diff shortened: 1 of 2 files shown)


AliceR:
Thanks for the context @hanbyul-here ! 
When we think about components, I believe they should be individually useful and composable, just like the map pattern demonstrated here.
If we want to support users who need predefined data layers (like in the E&A case), maybe we could offer those as separately exported components or **reusable hooks**. That way, users can opt into that behavior when needed, while still retaining full control over the map composition. It might be a nice balance between flexibility and convenience for different use cases.
What do you think? Maybe there are technical constraints or user experience goals I'm not aware of that could influence this direction?

vgeorge:
It seems this PR fulfilled the goal of demonstrating a custom map, so I don’t think we need to merge it. I posted an update in the upstream ticket [here](https://github.com/NASA-IMPACT/veda-ui/issues/1793#issuecomment-3175022908).

Issue #1814, opened 2025-08-14
Retire @devseed-ui/collecticons


vgeorge:
In https://github.com/NASA-IMPACT/veda-ui/pull/1844 we introduced guidance on how to migrate icons from Collecticons to USWDS and Custom Icons to drop this dependency.

This is a list of components that still need migration, organized by complexity. The idea is to start small and build up systematically.

When picking up one of the tasks from this list, click the three dots to the right of the task and create a sub-issue to indicate that you will be taking on this work. Then the sub-issue will be in progress, on the board, and linked to this issue.

**Components Using Collecticons:**

**Low Complexity** (1 icon)

* [x] scrollytelling/index.tsx – Scrollytelling
* [ ] card/index.tsx – Generic cards
* [x] catalog-content.tsx – Catalog content
* [x] catalog-legacy/catalog-content.tsx – Legacy catalog
* [x] filter-tag.tsx – Filter tags
* [ ] colormap-options.tsx – Colormap options
* [ ] colormap-section.tsx – Colormap section
* [x] empty-hub.tsx – Empty state
* [ ] figure.jsx – Figure component
* [ ] map-options/index.tsx – Map options
* [x] page-header-legacy/index.tsx – Main navigation
* [ ] nav-menu-item.tsx – Navigation items
* [ ] s-overview/index.tsx – Dataset overview
* [ ] layer-info-modal.tsx – Layer info
* [ ] analysis-message-control.tsx – Analysis messages
* [ ] tour-control.tsx – Map tours
* [ ] timeline-datepicker.tsx – Date picker
* [x] home/index.tsx – Home page
* [ ] hub-content.tsx – Hub content

**Medium Complexity** (2–3 icons)

* [ ] browse-controls/index.tsx – Browse controls
* [ ] catalog-card.tsx – Catalog cards
* [x] checkable-filter-legacy/index.tsx – Legacy filters
* [ ] draw-tools.tsx – Draw tools
* [ ] maps.tsx – Map components
* [ ] page-footer-legacy.tsx – Footer
* [x] search-field.tsx – Search field
* [ ] development/index.tsx – Development page
* [ ] tile-link-modal.tsx – Tile links
* [ ] timeline-controls.tsx – Timeline controls
* [x] timeline-zoom-controls.tsx – Timeline zoom
* [ ] timeline.tsx – Main timeline
* [ ] aoi-control.tsx – AOI management
* [ ] preset-selector.tsx – AOI preset controls
[...]

PR #1831, opened 2025-08-28
refactor: Test @devseed-ui/collecticons-react [Spike] 

AliceR:
**Related Ticket:** _https://github.com/NASA-IMPACT/veda-ui/issues/1814_

### Description of Changes
To test if we can migrate collecticons to import from the newer @devseed-ui/collecticons-react, I replaced the import for some legacy-icons and display them in storybook. This is just using the `createCollecticon` wrapper function to create react icons from svg.

POC: This shows that we can use the new collecticon library in a vite / storybook context. 

### Validation / Testing
- Icons are displayed without error in a storybook page


storybook/.storybook/preview.tsx
```diff
@@ -9,23 +9,28 @@ import '@uswds/uswds/css/uswds.css';
...
-      >
-        <DevseedUiThemeProvider>
-          <ReactQueryProvider>
-            <Story />
-          </ReactQueryProvider>
-        </DevseedUiThemeProvider>
-      </VedaUIProvider>
-    )
+    (Story, context) => {
+      if (context.parameters?.withProviders) {
```
vgeorge, line 13: @AliceR I think most of the stories would need the providers, can we make a reversed flag to avoid adding `withProvider=true` to all of them? Maybe `excludeProviders` or `noProviders`?

dzole0311 (commented):
@AliceR I tried to start the storybook, but when opening the Collecticons it throws: `Failed to fetch dynamically imported module: http://localhost:6006/src/stories/collecticons.stories.tsx`

Do you also get this? 

EDIT: Ah, I did not install the new package, I can see the icons now.

storybook/.storybook/preview.tsx
vgeorge: ok, sounds good 👍

Issue #1763, opened 2025-07-10
Migrate lockfile dependencies from Verdaccion to npm


vgeorge:
## Context

In https://github.com/NASA-IMPACT/veda-ui/issues/1369 we worked on using npm instead of Verdaccio registry. But the `yarn.lock` still references it.

## Acceptance Criteria

- [ ] No references to Verdaccio are present in the lockfile

## Definition of Done

- [ ] Unit tests
- [ ] (Where applicable) Automated playwright tests to cover primary use case(s)
- [ ] Merged to develop in relevant repo
- [ ] (?) Tested in instance context (GHG, earth.gov, template instance, etc)
- [ ] Relevant details of work are shared in a team forum (PR, ticket, slack, etc) to support knowledge sharing

cc @dzole0311 @hanbyul-here

vgeorge:
Migration complete. All repositories checked and confirmed clean of Verdaccio references. All yarn.lock files use official npm registry. No action needed.

PR #1855, opened 2025-09-09
fix: replace Verdaccio registry URLs with npm registry in yarn.lock

vgeorge:
## Problem

CI release workflow failing with:
```
Error: connect ECONNREFUSED 3.220.121.64:4873
```

The `yarn.lock` file contains 2,207 references to `verdaccio.ds.io:4873` which appears unavailable.

## Changes

- Replace all `http://verdaccio.ds.io:4873` with `https://registry.npmjs.org` in yarn.lock
- Addresses #1763

## Testing

- ✅ `yarn install` works locally
- ✅ `yarn buildlib` completes successfully

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -40,7 +40,7 @@ jobs:
         id: cache-node-modules
         with:
           path: node_modules
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package.json') }}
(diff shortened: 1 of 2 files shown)


vgeorge:
@AliceR it seems that the lockfile is now clear of Verdaccion references and installing packages is working well. If all checks passes we can merge.

AliceR (approved):

Issue #1849, opened 2025-09-03
Export pre-defined aoi selectors as presentational component


hanbyul-here:
## Context

Water insight tool has pre-defined aoi tools, which is very similar to what we have on E&A. 

## Acceptance Criteria

- [ ] predefined aoi selector should be flexible re.style, input&output

## Definition of Done
- [ ] Given when predefined aoi selector is exported from veda-ui

PR #1857, opened 2025-09-09
feat: export predefined aoi selectors

hanbyul-here:
**Related Ticket:** #1849 

### Description of Changes
Exporting preset selector as a component. PresetSelector component asks users to pass a list of `PresetOption`, which includes path for geojson file and label, returns the selected option's features from the geojsonfile through `onConfirm`.

- It is still a question how we are going to make this component customizable regarding style. 🤔 I will separate this issue, but I would l like to get some ideas from @dzole0311 if you have some from your experience with uswds!

### Validation / Testing
- Everything should work as expected on E&A page. 
- You can see how independently preset-selector can be used through Storybook

### DIFF
--- a/app/scripts/components/common/map/controls/aoi/preset-selector-container.tsx
+++ b/app/scripts/components/common/map/controls/aoi/preset-selector-container.tsx
@@ -100,8 +100,8 @@ export default function PresetSelectorContainer({
 
   return (
     <PresetSelector
-      selectedState={selectedState}
-      setSelectedState={setSelectedState}
+      selectedPreset={selectedState}
+      setSelectedPreset={setSelectedState}
       resetPreset={resetPreset}
       presets={presets}
       onConfirm={onConfirm}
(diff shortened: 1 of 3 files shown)


app/scripts/components/common/map/controls/aoi/preset-selector-container.tsx
```diff
@@ -0,0 +1,110 @@
...
+        path: `${baseUrl}/states/${name}.geojson`
+      }))
+      .sort((a, b) => a.label.localeCompare(b.label));
+
+    const countryPresets: PresetOption[] = [
+      {
+        group: 'country' as const,
+        label: 'Contiguous United States (CONUS)',
+        value: 'United States (Contiguous)',
+        path: `${baseUrl}/countries/United States (Contiguous).geojson`
```
vgeorge, line 94:
```suggestion
        path: `${baseUrl}/states/United States (Contiguous).geojson`
```

vgeorge (approved):
@hanbyul-here this is working well! I added a suggestion to fix the Contiguous US GeoJSON path. I'd also suggest adding JSDoc comments to the touched components for better maintainability.

PR #1893, opened 2025-10-10
refactor: retire Collecticons

vgeorge:
**Related Ticket:** #1814

We’ve been migrating Collecticons to USWDS using a progressive approach supported by Storybook. However, that approach has proven very time-consuming. We’ve learned that Collections generally have the same appearance when replaced with USWDS icons at size 3.

In this PR, I’m simply replacing the instances so we can perform visual inspections and get this across the finish line.

~~This is WIP, opening for visibility.~~

cc @faustoperez @snmln @aboydnw


snmln (approved):
Everything looks good! One house keeping item is that we might want to add all the icons to the storybook icon reference.

Issue #1900, opened 2025-10-30
Migrate library build tooling from Parcel to Vite


vgeorge:
I’d like to propose migrating the library build from Parcel to Vite for the next major release, v7 (see https://github.com/NASA-IMPACT/veda-ui/issues/1889).

`next-veda-ui` already handles MDX processing, so Parcel’s only remaining role is building the library. Since Vite is faster, better maintained, and already part of our stack in storybook, it makes sense to consolidate around it. 

With this change, we won’t maintain compatibility with instances using the old Parcel-based setup, those will remain supported under v6.

Thoughts or concerns before we move forward with this direction?

cc @dzole0311 @ifsimicoded @snmln @AliceR @aboydnw

Issue #1889, opened 2025-10-02
Version 7 Roadmap


vgeorge:
## Overview

This ticket is a starting point for discussion around the next major release. The idea is that v7 should be the place where we introduce the breaking changes that come with removing deprecated code and finalizing migrations (e.g. Collecticons, Devseed UI).

The goal is to have a clear spot to track items that must be addressed in a breaking version, while also leaving space for other tickets that make sense for a major release.

## Initial Scope

* Complete Collecticons retirement ([#1814](https://github.com/NASA-IMPACT/veda-ui/issues/1814))
* Reorganize folder structure ([#1871](https://github.com/NASA-IMPACT/veda-ui/issues/1871))
* Remove deprecated components and import paths (e.g. `card/ExternalLinkFlag`, `catalog-legacy/`)
* Complete documentation website #1370 
* Clean up Storybook by removing temporary comparison stories
* CMS? Plugins?

## Next Steps

This list is intentionally minimal. Let’s discuss what else should be in scope for v7 when possible.

cc @aboydnw @AliceR @dzole0311 @sandrahoang686 @faustoperez @snmln @hanbyul-here

vgeorge:
Following up on this: v7 work is not planned. The codebase will remain active for bug fixes and blockers, but refactors and migrations are out of scope for the foreseeable future.

PR #1871, opened 2025-09-15
docs: Add ADR 004 - Component Library Restructure

AliceR:
Proposes renaming `/storybook` to `/core` and establishing it as the primary location for library components with gradual migration from the current scattered structure.

## Context
- Current library components are scattered throughout /app directory
- Need clear separation between library and application-specific code
- Opportunity to modernize build system based on successful Vite adoption

## Proposed Solution
- Rename /storybook → /core to reflect library purpose
- Gradual migration strategy with backward compatibility
- Dual build system during transition (legacy Parcel + new Vite)
- Re-export strategy to avoid immediate component copying
- Future state: standalone `/core` package

## Benefits
- Modern development environment with Vite
- Clear component library boundaries
- Unified documentation and development
- Future-proof architecture for standalone library
- Low-risk incremental migration path

With this setup, it will be much easier to keep track of our refactor and migration work!

**Update: After this discussion, there was a slight shift in direction. Instead of renaming /storybook to /core, we’re moving toward a monorepo-based structure where the library lives under packages/veda-ui/. The goals and motivations below still apply.**

Action items:

* [X] Move the library to a dedicated folder (handled in [https://github.com/NASA-IMPACT/veda-ui/pull/1898](https://github.com/NASA-IMPACT/veda-ui/pull/1898))
* Update references in instances:
  * [ ] https://github.com/NASA-IMPACT/veda-config
  * [ ] https://github.com/US-GHG-Center/veda-config-ghg
  * [ ] https://github.com/NASA-IMPACT/veda-config-eic
* [x] Move the SPA to `apps/legacy-spa`
* [ ] Move Storybook to `apps/storybook`
* [ ] Update this ADR with the results of these changes

PR #1901, opened 2025-10-30
fix: Rename React component files from .js to .jsx

vgeorge:
Contributes to:

- #1900 
- #1889 
- #1898 
- #1871 

Notes:

- Renamed 33 files that import React to have .jsx extension
- This prepares the codebase for modern build tools like Vite
- Library build with Parcel tested and working correctly
- Vite build recognizes JSX syntax (SASS config needed separately)

This is ready for review.


ifsimicoded (approved):
I'm not sure the specifics of what should be checked or the review process, so you may want another's 👀 . But this seems like a simple change to me -- if the app transpiles/builds it should be good to go! That said, I saw [this thread](https://github.com/vitejs/vite/discussions/14652#discussioncomment-10093283) on adjusting the vite config to accept jsx with js files, but haven't tested.

Related, curious if we'll move to tsx and introduce typing? 🚀

Issue #1791, opened 2025-07-31
Share button to URL shortener


aboydnw:
## Context

URL shortening is ready!
https://openveda.cloud/service/link/docs

We should add a share button to our existing tools to promote sharing, focusing on those that have long URLs (thanks to state management)

- [Time Series Analysis](https://www.earthdata.nasa.gov/dashboard/exploration)
- [Fire Event Explorer](https://www.earthdata.nasa.gov/dashboard/tools/fire-event-explorer)

Implementation/design decision on the button, should be an icon? text? where to place?

## Acceptance Criteria

- [x] From the map page of Time Series Analysis and Fire Event Explorer, I can easily click a share button to copy a shortened URL to my clipboard
- [x] I see a success message indicating the shortened URL was copied to clipboard (failure state as well?)
- [x] Pasting that shortened URL into a browser window opens the same view that I had with the longer URL

## Definition of Done

- [ ] Unit tests
- [ ] (Where applicable) Automated playwright tests to cover primary use case(s)
- [ ] Merged to develop in relevant repo
- [ ] (?) Tested in instance context (GHG, earth.gov, template instance, etc)
- [ ] Relevant details of work are shared in a team forum (PR, ticket, slack, etc) to support knowledge sharing

snmln:
Initial Designs for fire event explorer: https://www.figma.com/design/hpmfg1jAM1poVo7lnNAOTx/Fire-Events-tool?node-id=177-1614&t=mqRLcgxD9lVpCtQN-1

Veda exploration tool: https://www.figma.com/design/yurEzrLMfTBrKugXD8w66Q/Share-button?node-id=0-1&t=IgYXozVTNqd7Rvww-1

aboydnw:
I think that it could be used as a plugin, or maybe a prop on a plugin? But it is a consistent request that I think other tools will want to have as well.

PR #1905, opened 2025-11-04
feat: Share button to URL shortener

snmln:
[1791](https://github.com/NASA-IMPACT/veda-ui/issues/1791)

### Description of Changes
Implementing share button functionality and successful copy link notification. 

### Validation / Testing
This end point only works on production links, the links returned will not be shortened version.

### DIFF
--- a/app/scripts/components/exploration/components/map/share-button.tsx
+++ b/app/scripts/components/exploration/components/map/share-button.tsx
@@ -19,7 +19,7 @@ function ShareButtonComponent({
   return (
     <SelectorButton
       tipContent={tipContent}
-      tipProps={{ placement: 'left', showOnCreate: true, hideOnClick: false }}
+      tipProps={{ placement: 'left', showOnCreate: false, hideOnClick: false }}
       disabled={disabled}
       onClick={onClick}
       onMouseLeave={onMouseLeave}


app/scripts/components/exploration/components/map/index.tsx
```diff
@@ -40,10 +43,21 @@ export function ExplorationMap(props: ExplorationMapProps) {
...
+  // Auto-reset linkCopied after 3 seconds
+  useEffect(() => {
+    if (linkCopied) {
+      const timer = setTimeout(() => {
+        setLinkCopied(false);
+      }, 3000);
+
+      return () => clearTimeout(timer);
+    }
+  }, [linkCopied]);
```
ifsimicoded: Curious if this could be managed by the share-button itself, so it's a self contained component we can plug in anywhere.

app/scripts/components/exploration/components/map/index.tsx
```diff
@@ -176,6 +190,17 @@ export function ExplorationMap(props: ExplorationMapProps) {
...
+        <ShareButton setLinkCopied={setLinkCopied} />
+        <div className='width-full display-flex flex-justify-center margin-top-3'>
+          <Alert
+            type='success'
+            headingLevel='h4'
+            slim
+            className={`${linkCopied ? '' : 'display-none'}  maxw-mobile`}
+          >
+            Link Copied to Clipboard.
+          </Alert>
```
ifsimicoded: Same here, curious if this could be contained within the ShareButton component with a bit of code that appends an alert to the bottom of body or something similar, so we could easily re-use this component. I'm unsure if the design component library being used already has a Snackbar component (here's [MUI's implementation](https://mui.com/material-ui/react-snackbar/)) that has this behavior of appending a DOM node and then removing it after a delay.

ifsimicoded (commented):
Here's the rest of my review! But please keep in mind I'm not aware of the specific patterns employed by veda-ui vs what I know of as typical UI patterns.

app/scripts/components/exploration/components/map/index.tsx
ifsimicoded: So I did some reading on accessibility and the recommendation is that toasts/snackbars come with a lot of accessibility concerns and that's why [USWDS didn't include it](https://github.com/uswds/uswds/discussions/5770) within their component library. 

NB: Instead of using a toast, wondering if the button tooltip text could be opened and updated to say `Link copied` instead of `Share Exploration`.

Screenshot 2025-11-09 at 9 56 52 AM

app/scripts/components/exploration/components/map/share-button.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useCallback } from 'react';
+import { useAtomValue } from 'jotai';
+import { Icon } from '@trussworks/react-uswds';
+
+import { SelectorButton } from '$components/common/map/style/button';
+import useThemedControl from '$components/common/map/controls/hooks/use-themed-control';
+import { timelineDatasetsAtom } from '$components/exploration/atoms/datasets';
+
+export function ShareButtonComponent({
```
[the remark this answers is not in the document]
snmln: Good catch, this was a copy paste hold over from the original component

app/scripts/components/exploration/components/map/index.tsx
snmln: Unfortunately USWDS does not have a snack bar component. This is a good call out, one concern is that area in the screenshot is currently being leveraged as a tooltip, parlaying that into a notification would be an UX anti pattern than what we have established across the platform as a whole.

app/scripts/components/exploration/components/map/index.tsx
snmln: That's a good option, unfortunately there are limitations to the react-map-box-gl control manipulations that would require us to change the designs for this. Which would introduce UX anti patterns as far as notifications go and what we have established so far in the platform.

app/scripts/components/exploration/components/map/index.tsx
ifsimicoded: Ah, where can I learn more about the ux patterns we've established? Would you be open to pairing to show me what you mean? Or maybe I'm not being clear! I was thinking you could just update the text from 'Share Exploration' to 'Link Copied' for a few seconds. This way everything could be managed in one component. I know there was appetite to make this url shortener a plugin, so a self-contained implementation would be really helpful for that.

app/scripts/components/exploration/components/map/index.tsx
ifsimicoded: Non blocking comments, but I would like to better understand what you mean by `limitations to the react-map-box-gl control`, `introduce UX anti patterns`. Maybe we could pair, and you can show me?

app/scripts/components/exploration/components/map/share-button.tsx
```diff
@@ -0,0 +1,85 @@
...
+      await navigator.clipboard.writeText(shortenedUrl);
+      setLinkCopied(true);
+      // TODO: Consider adding a toast notification for success feedback
+    } catch {
+      // Fallback: copy current URL if shortening fails
+      try {
+        await navigator.clipboard.writeText(window.location.href);
+        setLinkCopied(true);
+      } catch {
+        // Silent fail - clipboard API may not be available in all contexts
```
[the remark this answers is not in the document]
ifsimicoded: nice! 

NB: Maybe the component should return early instead of disabling the button? This way the user isn't confused why it's disabled, as the button would just not be present. Unless there's some extra messaging for it?

app/scripts/components/exploration/components/map/share-button.tsx
```diff
@@ -0,0 +1,77 @@
...
+  setLinkCopied: (copied: boolean) => void;
+}) {
+  const handleShare = useCallback(async () => {
+    try {
+      // Get current URL
+      const currentUrl = window.location.href;
+
+      // Call the URL shortening API
+      const response = await fetch(
+        `https://openveda.cloud/service/link/shorten?url=${encodeURIComponent(
```
slesaad: i think it's a good idea of setup this url as an env var instead of hardcoding it here

app/scripts/components/exploration/components/map/share-button.tsx
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      // get current URL
+      const currentUrl = window.location.href;
+      const urlShortenerEndpoint = envUrlShortenerEndpoint;
+
+      if (urlShortenerEndpoint) {
+        try {
+          const response = await fetch(
+            `${urlShortenerEndpoint}?url=${encodeURIComponent(currentUrl)}`
+          );
```
vgeorge:
```suggestion

      if (envUrlShortenerEndpoint) {
        try {
          const response = await fetch(
            `${envUrlShortenerEndpoint}?url=${encodeURIComponent(currentUrl)}`
          );
```

I don't think this alias is necessary here.

vgeorge (approved):
@snmln looks good to me! 👍

Issue #1864, opened 2025-09-12
[Bug] Request passed to `titiler-cmr` has the `legend` rather than `rescale`


paridhi-parajuli:
### Context
The dashboard is sending `legends`  to the rescale param of the api call rather than the `rescale` from  [mdx file](https://github.com/Air-Quality-Portal/veda-config-aq/pull/13/files#diff-25395bd3a034ff2c73328f6b4b3123d1473b7018a3342674b348e800a0be2189). 

URL: 

Blocking a release? : y

Regression issue? (y/n):
[If yes, add Regression label]

Priority: P2

### Steps to reproduce the issue

1. Goto https://deploy-preview-13--aqportal.netlify.app/exploration
2. Search for `TROPESS` layer and add to map
3. Look at the `titiler-cmr` api call in Networks call for whats being passed in the `rescale` parameter. 

<img width="3438" height="1802" alt="Image" src="https://github.com/user-attachments/assets/28c9513a-6c88-4a92-aef8-295a298decbc" />

ifsimicoded:
**Issue:**
@vgeorge and I were able to reproduce this behavior on the `update/veda-ui` branch. We determined this is caused by an override of the `reScale` prop in the[ Layer component](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/components/map/layer.tsx#L111) of veda-ui. The consuming component for the issue referenced here is `CMRTimeseries`, but this override also exists in other consuming components.

<img width="488" height="347" alt="Image" src="https://github.com/user-attachments/assets/8687c30a-449b-4d97-9959-03232953f7b6" />. 


**🔧 Potential fix:**
We could allow `rescale` defined within mdx sourceParams to take precedence over the override (shown below in [RasterPaintLayer](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/common/map/style-generators/raster-paint-layer.tsx#L55)), but we're unsure if this would be desired by other downstream consumers of `veda-ui`.
<img width="682" height="159" alt="Image" src="https://github.com/user-attachments/assets/3b603ec1-60fc-4e64-8028-b9730eb02d35" />. 


**❓ Clarification needed:**
@snmln would you know why this override was put in place and if it's safe to override? I see the related commits were made by you quite sometime ago. 

[...]

PR #1915, opened 2025-11-11
fix: [veda-ui-1864] Do not override rescale defined from source parameters in map layer

ifsimicoded:
**Related Ticket:**
https://github.com/NASA-IMPACT/veda-ui/issues/1864

### Description of Changes
The reScale override is removed as sourceParams.rescale is the target value that should be used to generate tileParams in RasterPaintLayer. Adjusting the gradient sliders should no longer update the TiTiler rescale request params. I think this what was said to be the desired behavior? (see screenshots)

mdx file: 
Image

local view:
Image

PR: https://github.com/NASA-IMPACT/veda-ui/pull/1915

### Validation / Testing
I need some assistance here as I'm unsure the full impact of this change.

cc @dzole0311 @vgeorge

### DIFF
--- a/app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
+++ b/app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
@@ -17,7 +17,6 @@ interface RasterPaintLayerProps extends BaseGeneratorParams {
   colorMap?: string | undefined;
   tileParams: Record<string, any>;
   generatorPrefix?: string;
-  reScale?: { min: number; max: number };
   metadataFormatter?: (
     tileJsonData: TileJSON | null,
     tileParamsAsString: string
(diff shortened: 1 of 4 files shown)


app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
```diff
@@ -51,10 +49,9 @@ export function RasterPaintLayer(props: RasterPaintLayerProps) {
   const updatedTileParams = useMemo(() => {
     return {
       ...tileParams,
```
ifsimicoded: the original sourceParams (including rescale) are passed through here in tileParams

app/scripts/components/common/map/style-generators/raster-paint-layer.tsx
```diff
@@ -51,10 +49,9 @@ export function RasterPaintLayer(props: RasterPaintLayerProps) {
...
-      ...(colorMap && { colormap_name: colorMap }),
-      ...(reScale && { rescale: Object.values(reScale) })
```
ifsimicoded: remove reScale (from settings) override.

PR #1916, opened 2025-11-12
fix: [veda-ui-1864] Update colormapscale to default to mdx rescale

ifsimicoded:
Closes: https://github.com/NASA-IMPACT/veda-ui/issues/1864

### Description of Changes
This PR updates the color map scale min and max within the legend to default to the rescale value set within the dataset's mdx file. Adjusting the gradient sliders after initial load SHOULD STILL update the TiTiler rescale request params. These changes are per @dzole0311 suggestions, Thank you for your guidance here!

mdx file: 
Screenshot 2025-11-12 at 9 14 21 AM

local view:
Screenshot 2025-11-12 at 9 13 38 AM

### Validation / Testing
I need some assistance here as I'm unsure the full impact of this change.

cc @dzole0311 @vgeorge

### DIFF
--- a/app/scripts/components/exploration/components/datasets/data-layer-card-container.tsx
+++ b/app/scripts/components/exploration/components/datasets/data-layer-card-container.tsx
@@ -24,7 +24,6 @@ export default function DataLayerCardContainer(props: CardProps) {
   // All atom-based state management
   const [datasets, setDatasets] = useAtom(timelineDatasetsAtom);
   const datasetAtom = useTimelineDatasetAtom(dataset.data.id);
-  const datasetLegend = dataset.data.legend;
   const [getSettings, setSetting] = useTimelineDatasetSettings(datasetAtom);
   const [isVisible, setVisible] = useTimelineDatasetVisibility(datasetAtom);
   const [colorMap, setColorMap] = useTimelineDatasetColormap(datasetAtom);


app/scripts/components/exploration/components/datasets/data-layer-card-container.tsx
```diff
@@ -31,6 +31,7 @@ export default function DataLayerCardContainer(props: CardProps) {
   const [colorMapScale, setColorMapScale] =
     useTimelineDatasetColormapScale(datasetAtom);
   const opacity = getSettings('opacity') ?? 100;
+  const effectiveReScale = colorMapScale ?? getSettings('reScale');
```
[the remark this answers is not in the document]
ifsimicoded: If not, I'll remove line 27

app/scripts/components/exploration/types.d.ts.ts
```diff
@@ -107,6 +107,7 @@ export interface DatasetSettings {
   colorMap?: string;
   // Active colormap scale.
   scale?: colorMapScale;
```
ifsimicoded: To be honest I'm very unclear on how mdx values for scale, legend min and max and rescale are intended to be used and are reconciled to `colorMapScale` in data-layer-card-container above and `reScale` in the layer.tsx component. 

I couldn't find documentation on the scale prop - https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md. This makes me worry that maybe scale and reScale are duplicative and I shouldn't be adding it?

🤷‍♀️ Maybe I shouldn't worry about it?

dzole0311: There are few things here:

- `scale` in the url should be renamed to `rescale`, so that there's no confusion. The prop [lives under the `sourceParams` of the layer](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md#properties). Again, this should be renamed to `rescale` rather than `scale`
- `legend` is only used in two cases 1) [to show visual guidance](https://github.com/NASA-IMPACT/veda-ui/blob/main/docs/content/frontmatter/layer.md#legend) for the user to know the min and max values the user can rescale the dataset from,to and 2) to show a warning if the user sets a rescale outside of those bounds in the UI (e.g. legends min and max are [0, 20], but the user sets the rescale to [0, 100), in which case the UI shows a warning)

> This makes me worry that maybe scale and reScale are duplicative and I shouldn't be adding it?

We're only adding a new `rescale` prop in the `DatasetSettings` not the general mdx configuration. The `DatasetSettings` is just for us to easily access props that we need in the UI side (e.g. opacity, colormap etc). The reason why the UI needs both `scale` and `rescale` is because it needs to know:

1. What are the min, max the user can rescale the colormap?
2. What is the actual set min, max that the user has set via the UI slider or the mdx configuration?

dzole0311 (approved):
Thanks for opening the PR! To summarize the issue and what the patch fixes:

1. The UI [first looks for a rescale value](https://github.com/NASA-IMPACT/veda-ui/blob/main/app/scripts/components/exploration/components/datasets/data-layer-card-container.tsx#L31) in the URL
2. Whatever is found there gets passed down to the [colormap range slider](https://github.com/NASA-IMPACT/veda-ui/blob/bbd7d50fa9da194ca636495585be2f4765e8a0ea/app/scripts/components/common/dataset-layer-card/color-range-slider/index.tsx#L45). There’s some prop-drilling happening here; long-term we should pull these values from a dedicated settings atom instead
3. If rescale is not present in the URL, the colormap range slider currently [falls back to using the legend min/max](https://github.com/NASA-IMPACT/veda-ui/blob/bbd7d50fa9da194ca636495585be2f4765e8a0ea/app/scripts/components/common/dataset-layer-card/color-range-slider/index.tsx#L142), which is incorrect (it should use the rescale value configured by the user in the mdx).
4. That incorrect fallback value is then used for rendering and gets persisted back into the URL, which is why the wrong legend values ended up being sent to TiTiler

So the correct precedence should be:

1. If the URL defines a `scale` value (this should be renamed to `rescale` for consistency but not in this PR due to scope creep), we use that
2. Otherwise, fall back to the rescale defined in the dataset’s MDX file (the value set by the content admin that created the MDX config for the dataset)

Legend min/max are for display only and should not override or substitute rescale.

Also I'd be happy to see suggestions on how we can make improvements in the workflow/code! (for example, there's some prop-drilling going on with the rescale value and the colormap that can be avoided).

ifsimicoded:
closes #1864

PR #1924, opened 2025-11-19
fix: properly load API_URL_SHORTENER_ENDPOINT through VedaUI context

vgeorge:
Fixes e2e test failures in PR #1905 by properly loading API_URL_SHORTENER_ENDPOINT through VedaUI context instead of directly accessing process.env which doesn't work in browser environments.


vgeorge:
@snmln @dzole0311 this now failed from a different error. It the shortener URL correct? When visit it returns a stac catalog.

Issue #1925, opened 2025-11-19
[Bug] No Closest Date Error When WMS Layer Is In Scrollytelling


acblackford:
### Context

When a WMS layer is used in a Scrollytelling block in a data story (e.g., NLCD data), the entire block will not populate and instead produce a 'no closest date' error. However, if the site refreshes, the Scrollytelling will occasionally work as intended and the layer appear correctly. 

URL: https://www.earthdata.nasa.gov/dashboard/stories/Tornadoes_2011_Vegetation

Blocking a release? (y/n): n

Regression issue? (y/n): n
[If yes, add Regression label]

Priority: P2

### Steps to reproduce the issue

1. Scroll through the data story until you get to Scrollytelling
2. The first attempt will likely cause the error to appear. Refresh, and it may work correctly.

PR #1926, opened 2025-11-20
fix: guard WMS date domain lookup in scrollytelling

vgeorge:
Contributes to #1925

Prevents 'No closest date found' error when WMS layer domain is not yet loaded in scrollytelling blocks. Returns requested date directly when domain is empty, allowing layers to render before STAC metadata loads.

**Changes:**
- Added guard in `layer.tsx` before calling `getRelevantDate()` for WMS layers
- Added guard in `getRelevantDate()` to handle empty domain arrays


vgeorge:
This is ready for review. Unfortunatelly we cannot to visual inspection on the deploy preview. I recommend the following steps:

- Run the instance https://github.com/NASA-IMPACT/veda-config locally with this branch
- Visit http://localhost:9000/stories/Tornadoes_2011_Vegetation
- Map loads without errors

PR #1933, opened 2025-12-02
fix: add missing export for embedded exploration

slesaad:

### Description of Changes
~~- Removes `LayoutProps` - it's not needed and is also breaking the library when importing in another app for some reason~~ (this is false apparently, it is needed so we need to figure out whats causing the issue)
- Adds export for EmbeddedExploration

### Validation / Testing
`yarn link` and import into https://github.com/US-GHG-Center/ghgc-ena.

### DIFF
--- a/app/scripts/libs/index.ts
+++ b/app/scripts/libs/index.ts
@@ -56,6 +56,7 @@ export {
 
 export {
   CatalogContent,
+  EmbeddedExploration,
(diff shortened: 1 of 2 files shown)


slesaad:
oof - just realized that it doesn't hide the header and footer without `LayoutProps`. need to figure out how to do that without breaking the library

PR #1922, opened 2025-11-17
feat: embed mode for e&a

ram-sharan25:
This PR introduces an `&embed=true` URL parameter that transforms the standard exploration view into a minimal, simplified dashboard optimized for embedding in an <iframe>. This view removes standard E&A (Explore and Analyze) components, focusing only on essential visualization elements: the map/visualization layer, the legend, and a date/time-change component. This significantly reduces visual clutter and load time for external embeds.

Uses: 
- MapBlock for single-layer viewing as well as comparison of two datasets.
- TimelineDatePicker for the calendar.

Other features added:
- zoom to a particular area on the map (using zoom and center url params)
- comparison slider with two different sliders
- calendar selection in embed mode

To test: 
Currently, there is no interface to create the embedding link. You can add a layer you want to embed (or also can compare two datasets)  edit the url by adding `&embed=true` at the end of the url in the search bar. At last,  reload to see the preview.

- View single Data:  
   -  Select a dataset in the exploration mode 
   - Edit the link by adding `&embed=true` at the end
   - Reload
 Example: 
   -  e-a link: https://deploy-preview-1922--veda-ui.netlify.app/exploration?datasets=%5B%7B%22id%22%3A%22VIIRS_SNPP_DayNightBand_At_Sensor_Radiance%22%2C%22settings%22%3A%7B%22isVisible%22%3Atrue%2C%22opacity%22%3A100%2C%22analysisMetrics%22%3A%5B%7B%22id%22%3A%22mean%22%2C%22label%22%3A%22Average%22%2C%22chartLabel%22%3A%22Average%22%2C%22themeColor%22%3A%22infographicB%22%7D%2C%7B%22id%22%3A%22std%22%2C%22label%22%3A%22St+Deviation%22%2C%22chartLabel%22%3A%22St+Deviation%22%2C%22themeColor%22%3A%22infographicD%22%7D%5D%7D%7D%5D&taxonomy=%7B%7D&search=&date=2024-05-05T05%3A00%3A00.000Z
[...]

PR #1950, opened 2025-12-08
fix: 1929 - follow on fixes to pr 1922

### LINKED ISSUE #1929: E&A embed mode refactor / questions
[https://github.com/NASA-IMPACT/veda-ui/pull/1922](https://github.com/NASA-IMPACT/veda-ui/pull/1922) introduced the embed mode for E&A.

Because of time constraints we merged it as-is, leaving open questions, documentation and file-organization improvements for follow-up work.

### Questions:
How are users vs developers using this view?
The expectation is that they will review ghgc exploration in full view to get the view they need with the necessary datasets and then take the url with embed=true appended to use that view in another website. 
This view should have a fully functional compare feature, but no timeline feature.
~Or is the expectation that they will only see the embed/simple view only and they are NOT using this exploration view to generate a view to embed elsewhere? Or some other use case?~
- ~how are datasets selected if a user is not allowed to be in full view mode~
- ~how can the embed/simple view be shared if it's within an iframe as the url won't change as the dates are modified (this is our assumption for how this is being used or is this being used as a typical e&a implementation)~ This is not an immediate concern

### TODO / Questions
- [x] fix such that embedExploration can be exported without issues with LayoutProps
- [x] fix such that embedExploration does not include SPA header and footer 

Image

- [x] fix tooltip on embed mode compare view 

Image

- ~update update ghgc-ena next js instance to allow hiding header and footer, with clear documentation (we're unsure we need this embed mode in next-veda-ui)~ Not necessary because we completed above
- [x] rename embed mode to simple mode as that's more reflective of the view
- ~remove extraneous time series components from embed/simple view?~ This is needed for compare
- ~fix compare components? (currently doesn't work. The date selector on the left isn't loading different tiles per date)~ This is needed and was caused by data being unavailable for the selected dataset
- do we want to add a control/button to allow a user to switch to simple view (similar to world view)?

### Previous todos below still apply:
In this ticket I propose we revisit the changes and confirm:

- [x] All files include JSDoc comments that explain their functionality -- some of this may be completed here: https://github.com/NASA-IMPACT/veda-ui/pull/1931
- [x] Filename and file structure updates related to the embed mode for better browsability (https://github.com/NASA-IMPACT/veda-ui/pull/1946)
- ~Adding LayoutProps component breaks the downstram application (see #1933)~

cc @dzole0311 @ifsimicoded @slesaad @vgeorge

ifsimicoded:
**Related Ticket:** 
Relates to the work completed in PR #1922 
Closes #1929 

### Description of Changes
This PR addresses some of the issues that came up when the "embed" mode was created for the GHGc portal. Details in the related ticket. Note, because the entire app is 'embeddable', I've changed the name of this view to be 'simple' mode. (Most of the changes are renames).

⚠️ @slesaad Note that I've changed the URL parameter to viewMode=simple, so this will need to be updated in downstream consumers when the package is updated.

### Validation / Testing
Ensure viewMode=simple renders the simple view, and any other viewMode renders the usual E&A view. embed should no longer alter the view.

### DIFF
--- a/app/scripts/components/exploration/index.tsx
+++ b/app/scripts/components/exploration/index.tsx
@@ -20,7 +20,7 @@ import { DATASETS_PATH, EXPLORATION_PATH } from '$utils/routes';
  * Container component that manages exploration view routing and data state.
  *
  * Routes between two view modes based on URL parameter:
- * - Simple view (`?viewMode=simple`): Minimal interface for embedding
+ * - Simple view (`?viewMode=simple`): Minimal interface used primarily for embedding in viewport constrained external apps
  * - Default view: Full exploration and analysis interface
  *
  * @LEGACY-SUPPORT
--- a/app/scripts/components/exploration/atoms/viewMode.ts
+++ b/app/scripts/components/exploration/atoms/viewMode.ts
@@ -15,7 +15,7 @@ const dehydrateViewMode = (value: ViewMode) => {
 /**
  * Atom that manages the exploration view mode via URL parameter.
  *
- * - 'simple': Minimal view for embedding (no navigation/header)
+ * - 'simple': Minimal view for used primarily for embedding (no navigation/header/footer/dataset selection UI or time series visualization)
  * - 'default': Full exploration and analysis interface
  *
  * URL parameter: `?viewMode=simple` (defaults to 'default')
--- a/app/scripts/components/exploration/views/simple/index.tsx
+++ b/app/scripts/components/exploration/views/simple/index.tsx
@@ -32,7 +32,7 @@ interface ExplorationAndAnalysisSimpleViewProps {
  * Simplified exploration view optimized for embedding.
  *
  * Renders only the map visualization and timeline controls,
- * without navigation, header, or dataset selection UI.
+ * without navigation, header, footer, dataset selection UI and time series visualization.
  *
  * @param props.datasets - Timeline datasets to display
  *


app/scripts/components/exploration/components/exploration-simple-view/exploration-simple-view.tsx
```diff
@@ -181,28 +182,34 @@ function EmbeddedLayersExploration(props: EmbeddedLayersExplorationProps) {
...
+          <TimelineSimpleView
             label=''
             date={selectedDay}
             setDate={setSelectedDay}
             datasets={layers as TimelineDataset[]}
             timeDensity={baseTimeDensity}
+            tipContent={
+              selectedCompareDay
+                ? 'Date shown on left map'
+                : 'Date shown on map'
```
ifsimicoded: fixes the tipContent issue

app/scripts/components/exploration/components/exploration-simple-view/exploration-simple-view.tsx
```diff
@@ -181,28 +182,34 @@ function EmbeddedLayersExploration(props: EmbeddedLayersExplorationProps) {
...
+      <StyledDivCompareTimelineContainer>
         {selectedCompareDay && (
-          <EmbedTimeline
+          <TimelineSimpleView
             label=''
             date={selectedCompareDay}
             setDate={setSelectedComparedDay}
             datasets={layers as TimelineDataset[]}
             timeDensity={compareTimeDensity}
+            tipContent='Date shown on right map'
```
ifsimicoded: fixes tip content issue.

ifsimicoded:
> 1\. Considering ExplorarionAndAnalysis and ExplorationSimpleView now have the same hierarchy level, it might make sense to enhance the folder structure further. Example:

👍 

> 2. I noticed that even when setting LayoutProps to hide layout components, there is still the banner and the site alert, which makes the simple view to slitghly offset and scroll. Not sure if we should address in this PR tough.

Will do in follow on PR. 

> 3.  This PR will supersede https://github.com/NASA-IMPACT/veda-ui/pull/1931, could we use this opportunity to add a bit more JSDoc on the exported helpers/types (view mode utils, props, etc.)? The code is already very readable, but this would improve IntelliSense in editors and generate autodocs in the future.

Will rely on you to complete this since you have a better idea of what you're looking for.

app/scripts/components/exploration/index.tsx
```diff
@@ -1,142 +1,89 @@
...
-      content: '';
-      display: block;
-      width: 2rem;
-      background: ${themeVal('color.base-200')};
-      height: 0.25rem;
-      border-radius: ${themeVal('shape.ellipsoid')};
-    }
-  }
-`;
+import { DatasetSelectorModal } from '$components/exploration/components/dataset-selector-modal';
```
ifsimicoded: This file is largely the same as the prior container  but I updated all the relative path imports.

app/scripts/components/exploration/index.tsx
```diff
@@ -17,12 +17,20 @@ import PageHero from '$components/common/page-hero';
...
+ * Routes between two view modes based on URL parameter:
+ * - Simple view (`?viewMode=simple`): Minimal interface for embedding
+ * - Default view: Full exploration and analysis interface
+ *
  * @LEGACY-SUPPORT
  *
- * @NOTE: This container component serves as a wrapper for the purpose of data management, this is ONLY to support current instances.
- * veda2 instances can just use the direct component, 'ExplorationAndAnalysis', and manage data directly in their page views
+ * @NOTE: This container component serves as a wrapper for the purpose of data management,
+ * this is ONLY to support current instances. veda2 instances can just use the direct
```
ifsimicoded: @vgeorge what is current instances vs veda2? is current instances, veda as a submodule? and veda2 is the the next-veda-ui implementation as a library (instead of submodule)?

when you say current instances, if I read this in one year, will that still apply?

PR #1957, opened 2025-12-17
refactor: move parcel dashboard to /apps

vgeorge:
Contributes to [https://github.com/NASA-IMPACT/veda-ui/pull/1871](https://github.com/NASA-IMPACT/veda-ui/pull/1871). Aims to be a step forward in separation of concerns in this repository.

Moves the parcel app to `/apps` as `dashboard-parcel`, allowing future variants (e.g. `dashboard-vite`, `dashboard-nextjs`).

Only dashboard-specific files were moved. MDX resolver code and content samples remain in root, to keep the scope of this small.

cc @dzole0311 @ifsimicoded

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@teamimpact/veda-ui",
   "description": "Dashboard",
-  "version": "6.20.0",
+  "version": "6.20.1",
   "author": {
     "name": "Development Seed",
     "url": "https://developmentseed.org/"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,13 +100,6 @@ jobs:
           repository: nasa-impact/veda-config
           event-type: update-version
           client-payload: '{"ref": "${{ github.ref }}", "VERSION_NUMBER": "${{ steps.git-release.outputs.VERSION_NUMBER }}"}'
-      - name: Trigger version update in Template Instance
-        uses: peter-evans/repository-dispatch@v3
-        with:
-          token: ${{steps.generate-token.outputs.token}}
-          repository: nasa-impact/next-veda-ui
-          event-type: update-version
-          client-payload: '{"ref": "${{ github.ref }}", "VERSION_NUMBER": ${{ env.VERSION }}}'
   notify:
     # If any of job fails
     if: failure()
--- a/docs/adr/004-component-library-restructure.md
+++ b/docs/adr/004-component-library-restructure.md
@@ -1,0 +1,131 @@
+# Component Library Restructure
+
+* Status: **Accepted**
+* Authors: @AliceR, @vgeorge, @ifsimicoded, @dzole0311
+* Deciders: VEDA UI Team
+* As of: December 2025
+
+## Context and Problem Statement
+
+This work builds on broader architectural changes captured in [ADR 006: VEDA2 Architecture Refactor](./006-veda2-architecture-refactor.md). ADR 004 focuses on repository structure changes to better separate library code from runnable applications.
+
+VEDA UI had a split architecture:
+
+1. **Storybook** in `/storybook` (Vite, separate deps)
+2. **Library exports** scattered through `/app` and bundled from `/app/scripts/libs/index.ts`
+3. **Two build systems**: Gulp/Parcel (app + library) and Vite (Storybook)
+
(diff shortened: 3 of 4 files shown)


tsconfig.json
```diff
@@ -17,7 +17,7 @@
       "$data-layer/*": ["packages/veda-ui/src/data-layer/*"],
       "$uswds": ["packages/veda-ui/src/uswds/"],
       "$graphics/*": ["packages/veda-ui/graphics/*"],
-      "$test/*": ["test/*"],
+      "$test/*": ["apps/dashboard-parcel/test/*"],
```
ifsimicoded: could tests be universal to all builds so we don't have to duplicate them and maintain several test copies.
Wondering if all tests files could live at the /apps/test/?
And this config would just become apps/test/*? Maybe there were other issues?

apps/dashboard-parcel/README.md
```diff
@@ -0,0 +1,30 @@
+# VEDA Dashboard
```
ifsimicoded: I'm wondering if only parcel specific documentation should live in this readme so we're not duplicating readmes for each build. 

Docs that reference Parcel which will need updates. 
- MDX_BLOCKS.md (references parcel specific url builder new URL)
- ARCHITECTURE.md (a lot of parcel specific build notes)
- DEPLOYMENT.md (you updated the path, but this is specific to the parcel build)
- PAGE_OVERRIDES_DEV.md (developer note on parcel type definitions)
- SETUP.md (remove link to parcel env var docs)

Also wondering if the parcel specific folders should be moved here from the top level too?

/parcel-resolver...
/parcel-transfomer...
and the automated parcel bundle reports

package.json
```diff
@@ -31,7 +31,7 @@
     "lint:css": "stylelint 'packages/veda-ui/src/styles/**/**' 'packages/veda-ui/src/**/*.(js|ts|tsx|jsx)'",
     "ts-check": "yarn tsc --noEmit --skipLibCheck",
     "test": "TZ=UTC jest",
-    "pretest:e2e": "node test/playwright/generateTestData.js",
+    "pretest:e2e": "node apps/dashboard-parcel/test/playwright/generateTestData.js",
```
ifsimicoded: This would be specific to a parcel only build. Wondering if playwright tests can be general to all builds (vite, next, etc) to prevent the need to duplicate tests that should be build neutral?

ifsimicoded (commented):
Couple questions, 

Are we expecting development on veda-ui to continue? My understanding, was we were "sunsetting" this repo and weren't making substantive changes, just bug fixes. But maybe I'm wrong! 

If we are continuing to develop, I'm curious why the dashboard is being updated to support multiple build tools? What is the value? How would an engineer choose which build to use? I'm open to understanding this better!

Added a few more comments below, if we're intending to support multiple builds for the dashboard, but my initial instinct is this adds confusion.

ifsimicoded (approved):
Thank you for clarifying!

**Requests / Questions:**
* Are there any updates that need to be made in consuming repos? Have you tested both submodule inclusion and library inclusion?
* can "Parcel folders at the root: agreed they should move" be documented in the ADR or ticketed in github as part of a monorepo or pre-v7 initiative? My worry is a lot of what is in your head about the plan for migration is only in your head, making it difficult for others to understand the plan, contribute, or pick up if your time is taken else where. If there's additional work that's required to take the pre-v7 to completion, can it be ticketed as [pre-v7] or similar and those could be referenced in TODO comments in the repo.

dzole0311 (approved):

vgeorge:
I forgot to say, there are no updates expected downstream.

ifsimicoded:
@vgeorge 

>  didn’t anticipate the parcel issues

Could you update the ADR with the parcel issues that came up, including 
* that the top-level parcel folders are part of the legacy parcel-dashboard output and that if the parcel build is deprecated and removed, those folders can be as well
* as part of the v7 body of work if/when a new build is introduced for the dashboard, additional updates will be made: 
  * documentation that references parcel will be updated
  * tests and test infrastructure will be updated
  * veda-app target will be updated to the new build
* anything else?

> I'd appreciate regrouping on this in our sync and getting your input on how we should move forward, then reflecting what we align for the future of veda-ui.

Can we do this after we figure out our PI objectives and staffing, and see what additional hours we'll have left?

Issue #1961, opened 2026-01-26
[Tech Debt] Use trusted publishing for release


slesaad:
## Context
Currently, we're using an NPM token to authenticate to NPM when releasing the @team-impact/veda-ui package. 

This requires creating a long-lived token, which is inherently unsafe. 

Also, we need to generate a new token every so often, which is an operational burden.

NPM Trusted Publishing is a new feature that replaces token-based authentication with a more secure, modern approach.

## Acceptance Criteria

- [x] Switch to NPM Trusted Publishing for package release

PR #1969, opened 2026-02-03
ci: use npx for npm version instead of global installation

slesaad:
### Description of Changes
Uses npx for npm version for publishing instead of installing globally which we apparently don't have permission for. See [this test workflow run](https://github.com/NASA-IMPACT/veda-ui/actions/runs/21605162352/job/62361132757)

I've already tested this via the [test-trusted-publishing branch](https://github.com/NASA-IMPACT/veda-ui/compare/test-trusted-publishing?expand=1) - [this](https://github.com/NASA-IMPACT/veda-ui/actions/runs/21636292958/job/62362594306) is the workflow run using `npm publish --dry-run`. It only failed because of the following error:
> npm error You cannot publish over the previously published versions: 6.20.3.

But the other steps in the workflow should create a new version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,8 +53,6 @@ jobs:
         with:
           node-version: ${{ env.NODE }}
           registry-url: https://registry.npmjs.org/
-      - name: Install npm 11.5.1+ for trusted publishing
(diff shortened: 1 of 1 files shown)


ifsimicoded:
relates to: #1961

PR #1982, opened 2026-04-15
feat: support `tilingMode: cog` for `raster` data type to allow visualization from external stac

slesaad:
# Support external STAC servers via `tilingMode: cog`

## Summary

Adds a new `tilingMode` option to `raster` datasets so they can be tiled
either via VEDA's mosaic register endpoint (the existing default) or via
titiler's per-item `/cog/tiles` endpoint. The latter unlocks consumption of
external STAC servers (Element84 Earth Search, GHG Center, NASA CMR
CloudSTAC, etc.) that don't expose a mosaic register endpoint.

## Dataset config

```yaml
type: raster
tilingMode: cog              # 'mosaic' (default) | 'cog'
stacApiEndpoint: https://earth-search.aws.element84.com/v1
tileApiEndpoint: https://openveda.cloud/api/raster
stacCol: sentinel-2-l2a
searchLimit: 100             # caps items per /search request
sourceParams:
  assets: red
  colormap_name: viridis
  rescale: [0, 10000]
```

- `tilingMode: 'mosaic'` (default): unchanged behavior.
  `/searches/register` produces one tilejson for the whole CQL filter.
- `tilingMode: 'cog'`: per-item COG tiling. For each STAC item, the
  selected asset's `href` (preferring `alternate.s3.href` when present)
  is passed to `/cog/tiles?url=...`. Multiple items render as separate
  layers stacked on the map.

## What's also new

- **`searchLimit`**: configurable per-dataset cap on items per `/search`.
  Some external servers (e.g. Element84) error out with `limit: 1000`.
  Works in both tiling modes.
- **Pagination**: STAC search responses with a `next` link now power a
  "Showing N of M items — Load More" overlay. Handles both GET and
  POST next-link methods.
- **`FootprintsLayer`**: replaces the marker-based `PointsLayer` for both
  raster modes. Renders STAC item geometry when the response includes it
  (cog mode), bbox polygons otherwise (mosaic mode). Respects parent
  layer's `hidden` / `opacity` / `generatorOrder`.

## File-level changes

| File | Change |
| --- | --- |
| `types/veda.ts`, `parcel-resolver-veda/index.d.ts` | Add `tilingMode`, `searchLimit` to `DatasetLayer`; remove unused `external-stac` from `DatasetLayerType`. |
| `style-generators/raster-cog-timeseries.tsx` *(new)* | Per-item COG tiling component. |
| `style-generators/raster-timeseries.tsx` | Adds `searchLimit` + pagination. Mosaic register guarded against re-registration on pagination. |
| `style-generators/footprints-layer.tsx` *(new)* | Polygon footprints layer used by both raster modes. |
| `style-generators/pagination-overlay.tsx` *(new)* | Shared "Load More" UI. |
| `exploration/components/map/layer.tsx` | `case 'raster'` branches on `tilingMode === 'cog'`. |
| `exploration/hooks/use-stac-metadata-datasets.ts` | New branch for `type === 'raster' && tilingMode === 'cog'` to handle external STAC metadata structures. |
| `mock/datasets/*-external.data.mdx`, `*-element84.data.mdx` | Test datasets exercising the new path against Element84, GHG Center, and CMR CloudSTAC. |

## Backwards compatibility

`tilingMode` defaults to `'mosaic'`. All existing `raster` datasets
continue to work without any config change.

## Test plan

- [ ] Existing internal raster dataset still renders identically (no new
  `searchLimit`/`tilingMode` in MDX).
- [ ] Sentinel-2 / Landsat (Element84) renders cog tiles; "Load More"
  reveals additional footprints without yanking the map.
- [ ] HLS L30 (NASA CMR CloudSTAC) renders cog tiles.
- [ ] Hide / show / change opacity / change order on a cog layer
  toggles its footprints alongside its tiles.
- [ ] Removing a cog dataset from the timeline cleans up its sources
  and layers from the map style.
- [ ] Date change while a `loadMore` is in flight aborts cleanly (no
  console warnings about state-after-unmount).

### DIFF
--- a/packages/veda-ui/src/components/common/map/style-generators/raster-cog-timeseries.tsx
+++ b/packages/veda-ui/src/components/common/map/style-generators/raster-cog-timeseries.tsx
@@ -1,17 +1,15 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
-import styled from 'styled-components';
 import startOfDay from 'date-fns/startOfDay';
 import endOfDay from 'date-fns/endOfDay';
-import { Button } from '@devseed-ui/button';
-import { glsp, themeVal } from '@devseed-ui/theme-provider';
 
-import { ExternalStacTimeseriesProps, ExternalStacItem } from '../types';
+import { RasterCogTimeseriesProps, StacItemWithAssets } from '../types';
 import { FIT_BOUNDS_PADDING, getMergedBBox, requestQuickCache } from '../utils';
 import useFitBbox from '../hooks/use-fit-bbox';
 import useMaps from '../hooks/use-maps';
 import FootprintsLayer from './footprints-layer';
 import { useRequestStatus, STATUS_KEY } from './hooks';
 import { RasterPaintLayer } from './raster-paint-layer';
+import PaginationOverlay from './pagination-overlay';
 import { userTzDate2utcString } from '$utils/date';
 import { S_FAILED, S_LOADING, S_SUCCEEDED } from '$utils/status';
 
@@ -32,7 +30,7 @@ interface StacLink {
 }
 
 interface StacSearchResponse {
-  features: ExternalStacItem[];
+  features: StacItemWithAssets[];
   context?: {
     matched?: number;
     returned?: number;
@@ -50,26 +48,13 @@ interface PaginationState {
   nextLink: StacLink | null;
 }
 
-const PaginationOverlay = styled.div`
-  position: absolute;
-  bottom: ${glsp(1)};
-  left: 50%;
-  transform: translateX(-50%);
-  z-index: 10;
-  background: ${themeVal('color.surface')};
-  padding: ${glsp(0.5, 1)};
-  border-radius: ${themeVal('shape.rounded')};
-  box-shadow: ${themeVal('boxShadow.elevationA')};
-  display: flex;
-  align-items: center;
-  gap: ${glsp(0.5)};
-  font-size: 0.875rem;
-`;
-
 /**
  * Extracts the asset href from a STAC item, preferring S3 alternate if available.
  */
-function getAssetHref(item: ExternalStacItem, assetKey: string): string | null {
+function getAssetHref(
+  item: StacItemWithAssets,
+  assetKey: string
+): string | null {
   const asset = item.assets[assetKey];
   if (!asset) {
     return null;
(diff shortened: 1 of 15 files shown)


packages/veda-ui/src/components/common/map/style-generators/raster-cog-timeseries.tsx
```diff
@@ -0,0 +1,500 @@
...
+import useMaps from '../hooks/use-maps';
+import FootprintsLayer from './footprints-layer';
+import { useRequestStatus, STATUS_KEY } from './hooks';
+import { RasterPaintLayer } from './raster-paint-layer';
+import PaginationOverlay from './pagination-overlay';
+import { userTzDate2utcString } from '$utils/date';
+import { S_FAILED, S_LOADING, S_SUCCEEDED } from '$utils/status';
+
+// Whether or not to print the request logs.
+const LOG = process.env.NODE_ENV !== 'production' ? true : false;
```
dzole0311:
```suggestion
const LOG = process.env.NODE_ENV !== 'production';
```

packages/veda-ui/src/components/common/map/style-generators/raster-cog-timeseries.tsx
```diff
@@ -0,0 +1,500 @@
...
+
+interface TileSource {
+  id: string;
+  tilesTemplate: string;
+  bbox: [number, number, number, number];
+  assetHref: string;
+}
+
+interface StacLink {
```
dzole0311: Looks like we could reuse StacLink, PaginationState and the loadMore callback between the two timeseries files

dzole0311 (approved):
@slesaad I did a smoke test using the mocked datasets from this PR and it works well. Approving this since it sounds like you’d like to get it merged for the GHG instance.

I'm adding a few notes below for documentation purposes especially if the same datasets are used downstream in other instances:

1. In the downstream app, we should disable colormap and analysis controls for the new vector layers (if the same layers are used there like the mocked ones here)
2. Clicking and zooming to the vector layer makes it invisible. This might be related to min/max zoom settings

https://github.com/user-attachments/assets/6587391c-0d7e-458b-b206-eb55acdad8a5

3. I don’t see the Mangrove AGB dataset mentioned in the test instructions available to test with. I only see/tested:
   - HLS Landsat via NASA CMR STAC
   - Landsat C2 L2 via Element84
   - Sentinel-2 via Element84

slesaad:
thanks @dzole0311 !

regarding your concerns:
1. yes, depending on the dataset, that makes sense!
2. the vector data is showing the footprint of the raster data that exist in these regions and when you click/zoom to the area, the vector layer disappears and the raster layer appears (they probably were slow to appear)
3. that's my bad, i didn't push mangrove dataset 😅 , i'll update the description

PR #1987, opened 2026-05-11
fix: make `variables` a repeat query params instead of comma separated

slesaad:
# Description

titiler-cmr expects variables to be a repeat param instead of comma separated


dzole0311 (approved):
Tested using the daily MiCASA dataset with the following `sourceParams`:

```
    sourceParams:
      concept_id: C3273639213-GES_DISC
      backend: xarray
      variables:
      - NPP
      - Rh
      expression: b1/b2
```

Tiles render correctly with variables=NPP&variables=Rh (repeat params) on the PR branch.

Image

Note: when passing multiple variables, an `expression` is required (titiler-cmr returns a 500 without one since it can't colormap a multi-band array directly).

